Issue #81, opened 2026-08-24
[Product] Support and verify Windows classroom deployment


djibian:
## Source and product requirement

The user currently validates WebeeBlocks locally on Linux, but the target classroom uses Windows PCs. Windows support is therefore a mandatory product property, not an assumed portability claim.

Webots R2025a officially supports Windows and ships its own MinGW C/C++ compiler, but WebeeBlocks must prove the complete product path on Windows.

## Required target

A student or teacher on a supported Windows machine must be able to install/open a prepared WebeeBlocks release, launch the Runtime v2 activity and use the real Blockly 13.2.1 + Zelos Robot Window without Git, Node.js, npm, Bash or a developer toolchain.

Webots R2025a and a supported browser may be explicit prerequisites. Any further prerequisite must be justified and documented.

## Required product behavior

- provide a Windows-compatible preparation/build/release path; do not require `tools/prepare_runtime_v2.sh` on student machines;
- ship prepared pinned Blockly 13.2.1 assets in the classroom release artifact;
- compile or ship the Runtime v2 controller correctly for Windows using the Webots-supported toolchain;
- handle Windows paths, spaces in `C:\Program Files\Webots`, path separators and executable suffixes correctly;
- support a documented browser on Windows (prefer the installed system browser or a clearly selected supported browser);
- keep project files from #71 portable between Linux and Windows;
- retain the same AST, preflight, interpreter, WWI protocol, PID/STOP, observability, reset/replay and French UI behavior;
- provide a one-action classroom launch path where feasible;
- define an offline-first deployment boundary: after the teacher prepares/installs the release, normal student use must not depend on npm/CDN/GitHub/network access. Audit the current remote Webots/PROTO/RobotWindow assets and package or cache what is causally required.

## Required gates

Automated, where reliable:

1. Windows runner prepares/checks the pinned UI assets without Bash-only assumptions;
2. Runtime v2 JavaScript/core/preflight/project-file tests pass on Windows;
3. the C controller builds with the official Webots R2025a Windows toolchain;
4. release artifact contains all required UI/controller assets and no development-only dependency;
5. Linux gates remain green;
6. project files round-trip Linux ↔ Windows to the same full AST.

Real Windows acceptance before claiming support:

1. clean Windows machine/profile with Webots R2025a;
2. install/unpack the classroom artifact;
3. launch the reactive world through the documented one-action path;
[...]

PR #144, opened 2026-09-04
Ignore purely visual top-level Blockly moves

djibian:
## Scope
Fix the W2 product finding recorded on #81 where dragging an unconnected top-level Blockly stack to a new screen position incorrectly displays `Programme modifié — réinitialisez la simulation avant de relancer`.

## Change
- treat a Blockly `BLOCK_MOVE` with no parent before or after as a purely visual workspace move;
- preserve the reset warning for structural moves that can change program semantics;
- add a discriminating runtime UI test covering both cases.

## Boundary
No AST, interpreter, backend, project-file or checkpoint/governance behavior is changed.

Refs #81

### DIFF
--- a/plugins/robot_windows/blockly_v2/main.js
+++ b/plugins/robot_windows/blockly_v2/main.js
@@ -265,8 +265,12 @@ async function runProgram() {
   }
 }
 
+function isPureVisualWorkspaceMove(event) {
+  return !!(event && event.type === Blockly.Events.BLOCK_MOVE &&
+    event.oldParentId == null && event.newParentId == null);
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO 09c65cbad9f8f1904810d896dd14503099242856

Bounded review: the change distinguishes a top-level visual BLOCK_MOVE (no parent before or after) from structural moves, preserving the reset warning for semantic topology changes. The discriminating test covers both the visual no-op and structural move path. No AST, backend, project-file or governance boundary is weakened.

Issue #147, opened 2026-09-04
[Product] Voluntary stop for an active simulation flight


djibian:
## Source

Human W2 Windows stability acceptance on #81 found that an active flight cannot currently be interrupted voluntarily from the student UI.

## Current implementation evidence

On current `main`:
- Runtime v2 exposes Run, reset/replay and step controls, but reset is deliberately unavailable while `runtimeRunning` is true;
- `WebeeBlocksWwiBackend` exposes TAKEOFF/MOVE/LAND/RANGE/RESET only; there is no active-flight stop/cancel request;
- `crazyflie_runtime_v2.c` likewise parses TAKEOFF/MOVE/LAND/RANGE/RESET only, although it already has a fail-safe motor-stop path and action timeout.

Therefore the W2 finding is a real missing product capability, not merely a hidden button.

## Goal

Provide one explicit student command to stop an active **simulation** execution without modifying the Blockly program.

The stop is an environment/runtime control, not a Blockly block and not part of the backend-neutral student AST.

## Required behavior

- visible French action only while a simulation program is actively executing;
- one click cancels the currently pending Runtime v2 execution promptly and fail-closed;
- controller reaches a neutral/safe simulation state and does not continue queued student actions;
- UI reports a neutral stopped outcome such as `ARRÊTÉ`, not a technical error;
- workspace/project-file locking remains coherent during cancellation;
- after stop, normal rerun follows the existing reset/replay boundary rather than silently resuming the old program;
- step/Continue and ordinary timeout/failsafe semantics remain non-regressed;
- no new AST statement, Blockly block or physical-flight debug capability.

## Smallest architecture to prove

Prefer a narrow Runtime-v2 control path, e.g. a dedicated cancel/stop request from UI/backend to the simulation controller, rather than abusing RESET or injecting LAND into the student AST. The implementation must define what happens to an in-flight pending request and ensure late responses cannot resume execution.

## Automated proof

At minimum prove:
1. STOP/cancel during an active MOVE prevents subsequent AST actions from being issued;
2. the pending request resolves/rejects deterministically without a generic technical error;
3. UI transitions to a neutral stopped state and requires the established reset/replay path before rerun;
4. malformed/late responses do not resurrect the cancelled execution;
5. existing fail-safe, reset, step and project-file locks remain green.

## Human boundary

No separate human test is requested merely to implement this slice. Real Windows acceptance is only needed later if machine evidence cannot establish the final interaction ergonomics required by #81.

Refs #81

[...]

PR #150, opened 2026-09-04
Add voluntary stop for active simulation flights

djibian:
## Scope
Implement #147: allow a student to stop an active **simulation** flight without modifying the Blockly program.

## Change
- add a simulation-only `Arrêter le vol` action visible only while Runtime v2 is executing;
- add a dedicated WWI `STOP` request outside the student AST;
- accept STOP before generic BUSY handling in the Webots controller;
- reject the active action as `USER_STOPPED`, hold the current measured pose when airborne, latch reset-required state, and prevent queued AST actions from continuing;
- classify `USER_STOPPED` as neutral `ARRÊTÉ` with reset/replay guidance;
- add regression evidence for UI wiring, protocol cancellation, pending-request cleanup and late-response no-op behavior.

## Safety boundary
STOP does not inject `land`, does not bypass the existing reset requirement, and does not add a Blockly statement or physical-flight control path. Existing timeout/failsafe semantics remain intact.

Refs #147
Refs #81

### DIFF
--- a/controllers/crazyflie_runtime_v2/crazyflie_runtime_v2.c
+++ b/controllers/crazyflie_runtime_v2/crazyflie_runtime_v2.c
@@ -339,11 +339,12 @@ int main(void) {
         continue;
       }
       if (request.kind == REQUEST_STOP) {
-        if (command == CMD_IDLE || command == CMD_RESET) {
+        if (command == CMD_RESET) {
           response_error(request.id, "NOT_RUNNING");
           continue;
         }
-        response_error(active_id, "USER_STOPPED");
+        if (active_id >= 1)
+          response_error(active_id, "USER_STOPPED");
         target_x = x;
         target_y = y;
         target_z = z;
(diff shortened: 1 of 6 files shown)


djibian (commented):
NO_GO ad44c2edcfe5cc7ac7c329847c2d4d4c395f8678

STOP is rejected when the C controller is `CMD_IDLE`, even though `runtimeRunning` can still be true during a sensor gap or step-debug pause. The JavaScript execution has no stop flag, so it may continue with a later AST action after the stop attempt. Repair by making STOP establish reset-required state across controller-idle/program-active windows and by making the JavaScript run observe the stop request. Add a regression proving that no later AST action is issued after STOP in an idle/pause window.

djibian (commented):
GO a198ec41e52fcc9b3925b808c317f46d081f64e8

Independent review of the repaired STOP candidate found the previous NO_GO closed.

- STOP is accepted while the controller is CMD_IDLE, except during RESET.
- STOP latches reset-required state and freezes the measured pose; subsequent controller requests fail closed until RESET.
- The WWI backend records simulationStopped before sending STOP and rejects later AST actions locally with USER_STOPPED until a successful RESET.
- The UI resumes a paused interpreter only to let cancellation propagate, records runtimeStopRequested, and cannot report normal completion after STOP.
- Regression evidence proves a later MOVE is not emitted after STOP, late responses are harmless, and reset re-enables actions.
- Exact-candidate CI Gate is green.

No new blocking regression found within #147's bounded simulation-only stop scope.

PR #145, opened 2026-09-04
Make missing landing a student-correctable error

djibian:
## Scope
Fix the W2 finding on #81 where a program that omits the final `atterrir` block surfaces as a generic technical error.

## Change
- classify invalid top-level flight boundaries as `PROGRAM_INVALID`;
- show a French student-facing correction message requiring `décoller` first and `atterrir` last;
- add a preflight regression test for a takeoff-only program.

## Boundary
No AST shape, interpreter semantics, backend protocol, safety rule, project-file behavior or governance mechanism is weakened.

Refs #81

PR #151, opened 2026-09-04
Require final reconstruction before Controller termination

djibian:
## Problem

V4 Controllers have twice terminated after completing their local action while useful eligible engaged work still remained in GitHub.

- In one run, the Controller stopped after restoring/progressing the current work while #145 still remained to be closed.
- In the clearest reproduction, the Controller published `GO 2250beb629aed3fe9ec3af6c477cf9b20b60e809` on #145 after exact-candidate CI succeeded, while #145 was Ready, clean, up to date with `main`, and immediately mergeable. It then terminated instead of merging. #150 also had a bounded applicable NO_GO that could be progressed.

The demonstrated gap is termination proof, not scheduling: the existing Controller loop already prefers reducing engaged work and uses pending CI latency productively.

## Change

- Require one final reconstruction of exact current `main` and relevant engaged GitHub work before termination, followed by another application of the Controller loop.
- State explicitly that completion of the execution's current local work is not by itself a termination condition.
- Add a minimal static oracle protecting exactly those semantics.

## Scope

Only:
- `AGENTS.md`
- `tools/ci/test_controller_contract.py`

No workflow, CI, notification, human-checkpoint, roadmap, external prompt, product-code, scheduler, or V4-principle changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -135,7 +135,11 @@ At launch and after every durable transition:
 5. otherwise start the highest-value product work not already sufficiently
    covered, in an isolated short branch; publish a PR only once real durable work
    exists, Draft if still mutable and Ready if complete;
-6. if no useful eligible action exists, terminate silently.
+6. before terminating, resolve exact current main and rebuild relevant engaged
+   GitHub work one final time, then apply this Controller loop again. Terminate
+   silently only if that reconstruction exposes no useful eligible action.
+   Completion of the execution's current local work is never by itself a
+   termination condition.
 
 Pending CI is not a reason to notify or globally idle. Use its latency for
 independent useful work when available.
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO 0421b54a2ccdeb0d02018fc606633cb1a686c846

Independent review found the bounded governance change coherent with the demonstrated defect and V4 contract.

- The Controller loop now requires a final exact-main / engaged-work reconstruction before termination.
- It explicitly rejects completion of the execution's current local work as a termination condition.
- The change does not alter scheduling priority, CI semantics, human checkpoints, notifications, roadmap, or product behavior.
- The static oracle protects the exact new termination semantics without weakening existing assertions.

No blocking regression found within the stated two-file governance scope.

Issue #66, opened 2026-08-23
[Product] Progressive pedagogical activity model


djibian:
## Validated product objective

WebeeBlocks activities must support a **compact, teacher-guided pedagogical progression**, not merely different worlds/toolboxes.

Canonical product vision: `docs/PRODUCT_VISION.md`.

## Goal
Target approximately **8 to 12 substantial activities** that progressively introduce and then combine:

1. sequence / first movements;
2. parameters / precise movement;
3. repetition;
4. sensing;
5. conditions;
6. sensor + condition in reactive/repeated behavior;
7. several perceptions / nested control flow;
8. variables / memory;
9. open autonomous strategy;
10. final challenge prepared in simulation and, when the physical backend is proven, executed on the real Crazyflie under teacher authorization.

The exact number may vary inside the 8–12 range. Prefer a few meaningful cumulative activities over many repetitive micro-exercises.

A single Webots world may expose multiple activity profiles/difficulty levels where useful.

## Progression model

Progression is **guided by the teacher**, but WebeeBlocks does **not** maintain per-student unlock/progress state.

The teacher controls which activities are available, for example through Moodle or school storage. No student account, completion database, badge, progress dashboard or success history is required.

## Product rules
- the activity defines the problem, instructions, world, toolbox, bounds, capabilities, success/failure rules, optional timer/score and compatible hardware;
- the student builds the solution from generic primitives;
- each later activity should reuse prior concepts where pedagogically useful;
- avoid magic blocks that encode the intended strategy;
- timing/scoring belongs to the activity/world unless explicitly part of the learning objective;
- no student-to-student leaderboard;
- new activities should normally be configurable without changing WebeeBlocks core code;
- declarative/versionable activity definitions are sufficient for current authoring; no graphical teacher activity studio is required;
- preserve `activity/profile -> Blockly -> backend-neutral AST -> preflight -> shared interpreter -> backend`.

## Acceptance direction
Future product work should demonstrate a coherent mini-progression, ideally reusing at least one world through multiple profiles, where the visible toolbox and constraints introduce concepts progressively while preserving the same generic block catalogue and backend-neutral semantics.

Acceptance must **not** require student accounts, persistence of success/failure, automatic unlocking, assessment data or a teacher dashboard.

## Dependency on the functional capability baseline

Do not commit the representative 8–12 activity progression against a knowingly temporary subset of the student-facing Runtime capabilities.

[...]

PR #152, opened 2026-09-04
Experiment with profile-limited Blockly dropdowns

djibian:
## Scope

Bounded #66 research only. No product profile/schema change.

This branch extends the existing real Runtime v2 student-UI probe to test the proposed generic profile field restriction on Blockly 13.2.1 + Zelos, and fixes CI selection so changes to that experiment actually execute the Runtime `student-ui` proof.

## Experiment

In the real Robot Window:
- install a temporary restriction for `webeeblocks_v2_move.DIRECTION` to existing internal values `forward/left` **before** opening the real `Vol` flyout;
- verify the first observable flyout block exposes only those two values;
- drag that real flyout block into the student workspace and verify the created block remains restricted;
- restore the generic options, reopen the flyout, then apply a different restriction containing `right` to prove profile switching does not progressively shrink options.

## CI causality

`experiments/runtime-v2-student-ui/**` is routed to the Runtime suite, where `student-ui` actually runs `experiments/runtime-v2-student-ui/run.sh` / `probe.py`. A selector regression test prevents this experiment from again receiving a green gate without executing its discriminating proof.

## Boundary

The experiment changes no product profile/schema, AST/backend identifiers, block definitions, runtime semantics or capabilities. The selector change only makes the existing experiment path execute its existing Runtime proof.

## Decision use

If this exact experiment passes, #66 may implement the same generic mechanism as profile presentation/constraint while retaining fail-closed preflight as authority. If it fails, record the causal Blockly/flyout limitation instead of broadening 66-A.

Refs #66


djibian (commented):
NO_GO 516db56aa8f68b5a9c3a1269782a2b8d8a7872d1

The experiment does not yet prove the discriminating #66 claim it names.

- It opens the real Vol flyout, waits 400 ms, then explicitly calls filterWorkspace(flyoutWorkspace). That proves FieldDropdown.setOptions can mutate an already-open flyout block, but not that an active profile restriction is applied before the pupil can see or choose a forbidden option.
- It does not exercise profile switching/reopening, so it does not prove options are recomputed from the generic block definition rather than progressively shrinking.

Smallest repair boundary: make the temporary experiment install the restriction through the same pre-display/category/flyout construction path a product profile mechanism could use, then assert the first observable real flyout already exposes only forward/left. Add one switch/reopen cycle that restores the generic options and reapplies a different restriction without cumulative shrinkage. Keep the experiment product-neutral and leave AST/backend semantics unchanged.

djibian (commented):
NO_GO d0a1f77e2341297a6180282b02deee840af6f856

The CI-selector repair is internally inconsistent on this exact HEAD.

`experiments/runtime-v2-student-ui/**` is now in `SHARED_RUNTIME_WEBOTS`, so `select(["experiments/runtime-v2-student-ui/probe.py"])` correctly returns Runtime=true and Webots=true, preserving the previous Webots coverage while adding the Runtime student-ui proof.

But `test_runtime_v2_student_ui_experiment_routes_to_runtime` still asserts Webots=false, so the selector regression test contradicts the intended selector behavior.

Smallest repair boundary: keep the shared Runtime+Webots routing, change that assertion to Webots=true, and obtain fresh exact-candidate CI/review.

djibian:
NO_GO be0d759ae2cd9af3215c396d5f423df46093c0e4

Exact-candidate Runtime `student-ui` proof fails in the profile-switch cycle. The first restricted flyout and dragged workspace block pass, but after `setAllowed(null)` the reopened real `Vol` flyout still exposes only `forward/left` instead of the generic options.

Cause: Blockly reuses the existing flyout workspace across category reopenings. `setAllowed` updates only the student workspace, while the previously constructed flyout blocks retain their old FieldDropdown options and are shown again unchanged.

Smallest repair boundary: when the profile changes, update any already-constructed flyout workspace while it is hidden, before reopening the category. Preserve the stronger oracle that the first *observable* flyout after each profile change is already correct; do not revert to post-display mutation. Then obtain fresh exact-candidate CI and independent review.

djibian (commented):
GO 54498704b6f6a96ded7f00d8222a943af9eaec1c

Independent falsification found no remaining blocker for the bounded #66 research claim on this exact Ready HEAD.

- The restriction is installed before the real Vol flyout becomes observable.
- The dragged real flyout block preserves the restricted FieldDropdown options.
- A profile-switch cycle restores the generic option set and then applies a second restriction containing a previously forbidden value, falsifying cumulative shrinkage.
- The experiment remains product-neutral: no profile schema, AST/backend identifiers, Runtime semantics, or capability contract changes.
- CI routing now guarantees that this experiment path executes the Runtime student-ui proof, and exact-candidate CI Gate is green.

The prior authoritative findings on 516db56..., d0a1f77..., and be0d759... are resolved by this candidate.

PR #172, opened 2026-09-05
Execute yaw turns in Runtime v2 Webots

djibian:
## Scope

Closes the next reuse-first #157 C1b simulation gap by executing the already-existing generic signed `turn(angle)` semantic through Runtime v2 Webots. No new Blockly block or AST family is introduced.

### Product path
- `webeeblocks_v2_turn` and the backend-neutral signed `turn(angle)` AST/interpreter path already exist;
- the WWI backend now advertises `turn` and emits a bounded `TURN` request;
- the Webots controller holds horizontal position and altitude while converging to the requested yaw target;
- turns are accepted only while airborne and only for signed angles with absolute value `1..179°`;
- existing STOP, reset, unsafe-attitude and action-timeout fail-closed behavior remains applicable.

### Automated proof
The real Runtime v2 Robot Window/Webots harness executes `+30°` then `-30°`, records start/target/end geometry, and the CI oracle requires measured yaw convergence with bounded position/altitude drift before accepting the trace.

The reference capability matrix is updated to record yaw turning as simulation-covered while keeping physical student execution explicitly unproven.

### Boundary
No physical backend claim, real flight, new student block family, wait/speed/range-vocabulary/LED behavior, governance, checkpoint or notification change.

Draft until the exact candidate is exercised by CI.

Refs #157.


djibian:
GO 2f86c1ba2fd14f235ebff1da062a03f16cffb38c

Independent falsification review found the bounded #157 yaw-turn slice coherent with the current product/contract: it reuses the existing signed turn semantic, preserves the backend-neutral path and fail-closed controller lifecycle, keeps the simulation/physical claim boundary explicit, and adds discriminating real-Webots geometry evidence for signed yaw convergence with bounded drift. No applicable unresolved refutation found.

PR #174, opened 2026-09-05
Add bounded wait execution to Runtime v2 Webots

djibian:
Closes a bounded #157 capability gap by wiring the existing generic `wait(seconds)` semantics through the Runtime v2 WWI backend into the Webots controller. The controller holds position/yaw/altitude during a bounded 0.1–5.0 s wait, rejects invalid/unavailable use fail-closed, and the real Webots CI harness exercises the timing path.

Preserves the backend-neutral pipeline and adds no new student-facing magic block.


djibian (commented):
NO_GO 0f7750f93f91e9f26873bf11a852cefa4b1aedfa

Two bounded repairs are required before this candidate can be integrated:

1. The real-Webots CI oracle is stale. `prepare_runtime_v2_webots.py` now executes and instruments `WAIT`, but `.github/workflows/ci-runtime.yml` still expects the trace `... TURN, TURN, LAND` and never parses/asserts `WAIT_START`/`WAIT_END` or `WAIT_OK`. The candidate therefore does not yet provide a discriminating automated proof for `wait(seconds)` timing/hold semantics and its existing trace oracle is inconsistent with the new execution path. Update the oracle to require the WAIT trace/event, assert elapsed simulated time is at least the requested duration within a bounded tolerance, and assert position/altitude/yaw remain within the intended hold tolerances.

2. #157 requires the capability matrix to remain current. `docs/REFERENCE_CAPABILITY_MATRIX.md` still says the Webots backend explicitly rejects wait execution and marks the capability partial/missing. Update that row only once the repaired exact-candidate Webots evidence supports the covered-in-simulation claim.

No broader redesign is required; the controller/backend implementation itself is otherwise coherent with the existing generic `wait(seconds)` AST/interpreter path.

djibian (commented):
GO 0b0c5b876d94f9d8f4648f554b0b4c4c12925332

Independent review of the exact Ready candidate found the bounded wait slice coherent with the existing backend-neutral wait semantics. The prior NO_GO on 0f7750f93f91e9f26873bf11a852cefa4b1aedfa is resolved on this head: the real-Webots oracle now requires WAIT trace/timing/hold evidence and the capability matrix is updated consistently. No still-applicable refutation remains from that review.

PR #178, opened 2026-09-05
Prepare props-off S3 human checkpoint profile

djibian:
## Scope

Advances #70 from deterministic S3 build evidence to an explicit, fail-closed human-checkpoint capability without requesting a test yet.

- enables a checkpoint-only `s3-props-off` profile in the trusted `human-checkpoint.yml` registry;
- binds that profile to the existing deterministic `experimental-s3-surface-offset-2026-08` evidence artifact;
- copies the exact built `cf2.bin` into that artifact alongside its SHA-256 evidence so the physical checkpoint has a flashable firmware input;
- rejects `release` purpose for this Lab-only profile;
- updates workflow-contract checks and notification documentation.

The profile still requires the normal exact target SHA, full Runtime/Webots evidence, artifact digest, single human-test slot and deduplication before any `TEST_REQUIRED` can be published.

No checkpoint request is emitted by this PR. No motorized flight, `rangeUp`, Runtime v2 behavior change, estimator expansion, or physical-support claim.

Refs #70.


djibian:
GO 5583f1dc108f8fdcfb839454247b00c6fd511045

Independent falsification review against current main@a66242aff3cbd18ac2bbacbf1274296b93fbccfd, the current V4 human-boundary contract, ROADMAP X3 and durable #70 evidence found this exact Ready candidate coherent. It adds only an explicit checkpoint-only `s3-props-off` registry entry, binds it to the existing deterministic 2026.08 S3 build artifact, ensures the exact built `cf2.bin` is carried inside that artifact with digest evidence, and rejects release-purpose use. The trusted workflow still requires exact target SHA resolution, full Runtime/Webots evidence, one matching unexpired artifact with a GitHub sha256 digest, one-open-test serialization and fingerprint deduplication before publishing TEST_REQUIRED. The reusable Webots path deterministically selects and rebuilds S3 for non-PR checkpoint calls. No checkpoint is requested by this PR, and no motorized-flight, Runtime v2, rangeUp, estimator-expansion or physical-support claim is introduced. No applicable unresolved authoritative refutation found.

djibian (commented):
GO 5583f1dc108f8fdcfb839454247b00c6fd511045

Independent falsification against main@a66242aff3cbd18ac2bbacbf1274296b93fbccfd, the current human-boundary contract, and #70 found no applicable defect in this bounded checkpoint-enablement slice. The new profile is explicitly checkpoint-only, binds to the deterministic S3 artifact, makes the exact built cf2.bin available with recorded SHA-256 evidence, preserves exact target/provenance/digest/single-slot/deduplication checks, and emits no checkpoint request by itself. It does not claim motorized flight, physical support, rangeUp, or Runtime v2 behavior.

PR #192, opened 2026-09-05
Add read-only physical capability preflight contract

djibian:
## Scope

Advances #157 toward the physical-backend boundary with a deliberately non-motorized P0a contract.

This candidate:
- adds a read-only physical capability descriptor/inspection contract for the reference Crazyflie path;
- requires an established Crazyradio/Crazyflie identity and explicit reference hardware declarations;
- checks activity hardware requirements plus the already-generic action/range/move/vertical capability facts before any future physical execution layer;
- fails closed on missing decks, missing generic capabilities, disconnected/unknown transport, duplicate/malformed descriptor values, or an adapter interface exposing flight/action authority;
- adds a deterministic machine oracle to the existing Runtime v2 core gate.

The slice exposes no physical execution backend, takeoff/thrust/setpoint/arming path, and does not claim that cflib/Crazyradio transport has been product-proven. The actual host-library adapter and any indispensable read-only real-hardware evidence remain subsequent bounded work. No #70 estimator behavior, student vocabulary, final-flight logic, or checkpoint mechanism changes.

Refs #157.

### DIFF
--- a/plugins/robot_windows/blockly/webeeblocks/physical_capability_contract.js
+++ b/plugins/robot_windows/blockly/webeeblocks/physical_capability_contract.js
@@ -1,0 +1,131 @@
+(function(root, factory) {
+  if (typeof module === 'object' && module.exports)
+    module.exports = factory();
+  else
+    root.WebeeBlocksPhysicalCapabilityContract = factory();
+})(typeof self !== 'undefined' ? self : this, function() {
+  'use strict';
+
+  var FORBIDDEN_AUTHORITY_METHODS = [
+    'takeoff', 'land', 'move', 'vertical', 'turn', 'wait', 'setSpeed',
+    'setLight', 'arm', 'disarm', 'startMotors', 'stopMotors', 'setpoint',
+    'sendSetpoint', 'thrust'
+  ];
+
+  function fail(message) {
+    throw new Error('physical capability contract: ' + message);
+  }
+
+  function isObject(value) {
+    return value !== null && typeof value === 'object' && !Array.isArray(value);
+  }
+
+  function requireString(value, path) {
+    if (typeof value !== 'string' || value.trim() === '')
+      fail(path + ' must be a non-empty string');
+  }
+
+  function normalizeStringArray(value, path) {
+    if (!Array.isArray(value))
+      fail(path + ' must be an array');
+    var seen = Object.create(null);
+    return value.map(function(item, index) {
+      requireString(item, path + '[' + index + ']');
+      if (seen[item])
+        fail(path + ' contains duplicate value: ' + item);
+      seen[item] = true;
+      return item;
+    });
+  }
+
+  function normalizeDescriptor(value) {
+    if (!isObject(value))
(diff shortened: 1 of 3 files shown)


djibian (commented):
NO_GO d7b710db218752e5fdc7490bd1c9e9ca9701a722

Independent falsification against current main@636e669e4c9268753e2dbb8657e3f138443ac8a3, PRODUCT_VISION.md, current ROADMAP.md and the durable #157 P0 evidence guard finds two bounded authority defects in this exact Ready candidate.

1. `normalizeDescriptor()` requires and then republishes `device: 'crazyflie-2.1'` as if it were a machine-established fact. The current supported cflib/platform/parameter evidence inspected for P0 establishes a connected Bitcraze Crazyflie plus firmware/deck facts, but does not establish exact Crazyflie 2.1 airframe identity. The test descriptor therefore proves only a self-asserted value, not the required fail-closed evidence boundary. Smallest repair: represent machine-observed Crazyflie-family connection separately from exact model identity; keep exact model explicitly unproven unless supported by distinct evidence, and make any profile requirement for `crazyflie-2.1` fail closed while that evidence is absent.

2. The contract checks for forbidden adapter methods, but the normalized descriptor has no explicit non-authority invariant. Capability compatibility can therefore be serialized without an unambiguous statement that it grants no execution authority. The durable P0 guard requires an explicit field such as `executionAuthority: false`, validated fail-closed and covered by the oracle, in addition to the absence of motor/setpoint/arming methods.

The separation of the physical capability contract from the executable Runtime backend, hardware/capability subset checks, and CI wiring are otherwise directionally coherent. The actual cflib/Crazyradio read-only adapter remains legitimately outside this candidate's stated P0a scope. A later exact HEAD needs fresh CI and an independent review establishing both findings resolved.

djibian (commented):
GO d2ab517bd83d5a39dc2a378cdf5429f402dda9ea

Independent falsification of this exact Ready candidate against main@636e669e4c9268753e2dbb8657e3f138443ac8a3, the #157 P0 evidence guard, PRODUCT_VISION.md and ROADMAP.md finds the two still-applicable findings from NO_GO d7b710db218752e5fdc7490bd1c9e9ca9701a722 resolved.

1. Exact airframe identity is now separated from the observed Crazyflie family: `identity.modelEvidence='unproven'` requires `model=null`, exact `crazyflie-2.1` profile requirements fail closed without verified model evidence, and exact airframe identity is rejected from generic hardware evidence. The oracle exercises the unproven and misencoded cases.
2. The descriptor now requires and republishes `executionAuthority:false`, rejects any other value, and the adapter boundary still rejects flight/action-authority methods. The oracle covers both invariants.

I also tried the main bounded failure modes in this slice: disconnected/unknown transport, missing required deck/range capability, malformed/duplicate descriptor values, and accidental action authority all fail closed. The candidate still introduces no cflib transport adapter or motorized execution path, so it does not overclaim real-hardware proof.

No unresolved applicable refutation remains for this HEAD. CI Gate is still required independently before integration.

PR #193, opened 2026-09-05
Add read-only Crazyradio capability probe

djibian:
## Scope

Advances #157 P0b after integrated P0a (#192) with the actual supported-host-library read-only evidence path.

This candidate:
- requires an explicit `radio://` Crazyradio URI and performs no automatic interface scan;
- connects through cflib, waits for the parameter TOC/values, then queries firmware revision and reference-deck presence/self-test parameters without exposing a WebeeBlocks write/command API;
- maps observed Flow Deck V2, Multi-ranger and bottom Color LED evidence into the integrated generic physical capability descriptor;
- preserves the exact-airframe boundary from P0a: Crazyflie-family connectivity is observed while exact `crazyflie-2.1` identity remains `unproven`;
- preserves `executionAuthority:false` and exposes no WebeeBlocks parameter-write, commander/supervisor, arming, setpoint, thrust or motor API;
- fails closed on missing/malformed required read-only evidence and does not advertise failed/missing deck capabilities;
- adds a deterministic CI oracle using synthetic read-only values and cross-checks the result against the integrated P0a JavaScript normalizer without requiring cflib or hardware in CI.

**cflib transport side-effect boundary:** the normal `SyncCrazyflie` close path emits cflib's safety-zero shutdown setpoint before closing the link. Therefore `read-only` here is deliberately scoped to capability-evidence access and the WebeeBlocks adapter API; it does **not** claim that cflib emits no command packet at transport level. That library shutdown effect is not exposed as, or treated as, WebeeBlocks execution authority.
This slice does not create a physical execution backend or authorize flight. Exact Crazyflie 2.1 model proof and any indispensable real-hardware acceptance remain later bounded evidence. No TEST_REQUIRED is introduced here because the machine-only read-only path can be established without a human result.

Refs #157.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -41,6 +41,10 @@ def main():
     if physical_capability_test.returncode:
         print('FAIL read-only physical capability contract',file=sys.stderr);print(physical_capability_test.stdout,file=sys.stderr);print(physical_capability_test.stderr,file=sys.stderr);return physical_capability_test.returncode
     print(physical_capability_test.stdout.strip())
+    physical_probe_test=subprocess.run([sys.executable,'tools/ci/test_physical_capability_probe.py'],text=True,capture_output=True)
+    if physical_probe_test.returncode:
+        print('FAIL read-only Crazyradio capability evidence',file=sys.stderr);print(physical_probe_test.stdout,file=sys.stderr);print(physical_probe_test.stderr,file=sys.stderr);return physical_probe_test.returncode
+    print(physical_probe_test.stdout.strip())
     browser=browser_binary()
     with socketserver.TCPServer(('127.0.0.1',0),QuietHandler) as server:
         port=server.server_address[1]; threading.Thread(target=server.serve_forever,daemon=True).start(); time.sleep(.05); url=f'http://127.0.0.1:{port}/{HARNESS.as_posix()}'
--- a/tools/ci/test_physical_capability_probe.py
+++ b/tools/ci/test_physical_capability_probe.py
@@ -1,0 +1,159 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+import importlib.util
+import json
+import pathlib
+import subprocess
+import sys
+
+ROOT = pathlib.Path(__file__).resolve().parents[2]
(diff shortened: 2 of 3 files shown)


djibian (commented):
NO_GO c11e66f2b4ba433426be132cbcb3e5d3a7647aa2

One blocking fail-closed defect remains. The candidate advertises Flow movement/vertical capability whenever `deck.bcFlow2 != 0` and Multi-ranger capability whenever `deck.bcMultiranger != 0`. In current Crazyflie firmware those are initialization/presence flags, not successful self-test results. `flowdeck2Test()` performs a later independent Z-ranger test, and `mrTest()` performs later independent initialization/tests of all five range sensors; either test can fail while the corresponding `deck.*` presence flag remains nonzero.

Therefore present-but-failed hardware can be serialized as usable capability evidence, contrary to the P0 requirement that failed/incompatible evidence fail closed. Current firmware exposes read-only `system.selftestPassed`, and its boot self-test includes `deckTest()`; Color LED already uses its own explicit read-only test result.

Smallest repair: require established read-only health evidence before advertising Flow/Multi-ranger capability facts. A conservative bounded repair is to require `system.selftestPassed == 1` together with the deck presence flags, or use a narrower proven health signal if available. Add oracle cases for present-but-self-test-failed hardware.

The explicit Crazyradio path, exact-model-unproven boundary, and `executionAuthority:false` remain coherent. CI remains an independent requirement for a later exact HEAD.

djibian:
GO 8f910d9987c48e4a4f5408c866c16b7a6d39fa85

djibian (commented):
GO 8f910d9987c48e4a4f5408c866c16b7a6d39fa85

Independent falsification of the current Ready HEAD found the prior fail-closed finding resolved. Flow Deck and Multi-ranger capability advertisement now requires both deck presence and `system.selftestPassed == 1`; present-but-self-test-failed cases are covered and fail closed. Color LED retains its narrower explicit deck-test evidence. The exact-airframe model remains deliberately unproven, the probe requires an explicit `radio://` URI, and the WebeeBlocks surface preserves `executionAuthority:false` with no parameter-write/commander/arming/setpoint API.

The current canonical pull-request run #34143829996 is associated with this exact HEAD and completed successfully, including the final `CI Gate` job. The historical NO_GO for c11e66f2b4ba433426be132cbcb3e5d3a7647aa2 is therefore resolved for this candidate.

PR #198, opened 2026-09-07
Diagnose Firefox Qt crash boundary on current main

djibian:
## Scope

Reactivates #87 only for the preserved next causal experiment, because the higher-priority #157/#70 machine paths are currently waiting on the single outstanding physical checkpoint.

This candidate does **not** implement Firefox file operations. It now conforms to the V4 CI topology and the preserved #87 experiment boundary:

- no additional workflow file: the bounded experiment runs as extra steps inside the already-authorized `crazyflie-runtime-wwi` job;
- no debugger or intuitive diagnostic dependency is installed;
- the diagnostic uses only the controller itself plus libc signal/backtrace facilities already present in the CI/runtime environment;
- the official pinned Webots R2025a desktop SDK/runtime closure, Xvfb, bundled Qt and controller-local `WEBOTS_LIBRARY_PATH` are preserved;
- the probe installs fail-closed fatal-signal capture for SIGSEGV/SIGABRT/SIGBUS/SIGILL/SIGFPE immediately before `QApplication`, records an in-process backtrace, and distinguishes that from successful traversal through `QFileDialog`;
- ordinary Runtime v2 behavior is unchanged unless `WEBEEBLOCKS_QT_INIT_DIAGNOSTIC=1` is explicitly enabled.

The earlier transient runs remain research evidence only: run 34146595043 established that the real controller can be observed blocked in `wb_robot_init()` before the former rendezvous, but the installed-gdb/extra-workflow harness was correctly rejected and has been removed.

The goal remains causal evidence, not a product fix. No QPA/DISPLAY change, new dependency, file-operation behavior or Firefox product semantic is introduced.

Resolves the applicable NO_GO findings on `6538e184e6b62b359c7c11792dd814e00d6a1a8a` and `63709c299ffb9f28adf41e3f976192c8d0375bad`.

Refs #87.


djibian (commented):
NO_GO 6538e184e6b62b359c7c11792dd814e00d6a1a8a

Two independent blockers refute this exact Draft candidate.

1. The new `.github/workflows/runtime-v2-qt-crash-diagnostic.yml` violates the current V4 repository workflow contract. Exact-head CI run 34146288995 fails `test_workflow_contract.py::test_only_v4_workflows_exist`; do not weaken that oracle to admit the research workflow.

2. The diagnostic workflow installs `gdb` with `apt-get`. Issue #87 explicitly preserves the next experiment boundary as a diagnostic mechanism already available in CI and says not to install intuitive dependencies. Installing the missing debugger changes that agreed evidence boundary instead of testing it.

Smallest repair boundary: remove the additional workflow and debugger installation. If this diagnostic remains useful, run it through an already-authorized V4 CI surface using only an actually available diagnostic mechanism, while preserving the R2025a/Xvfb/runtime.ini/Qt environment. Otherwise abandon this candidate; do not turn these blockers into product-semantic changes.

djibian (commented):
NO_GO 63709c299ffb9f28adf41e3f976192c8d0375bad

The new diagnostic result is useful, but this exact candidate still does not resolve the two blocking findings from the prior head.

- `.github/workflows/runtime-v2-qt-crash-diagnostic.yml` remains an extra pull-request workflow and canonical CI still fails the V4 workflow allowlist contract.
- The workflow still installs `gdb` with `apt-get`, contrary to #87's preserved experiment boundary requiring a diagnostic mechanism already available in CI and explicitly forbidding installation of intuitive dependencies.

Run 34146595043 is nevertheless valid research evidence for a narrower fact: the exact real controller process was found and an early stack was captured before the in-process Qt rendezvous; frame #0 is in libc `read()` and the stack includes `wb_robot_init()` in Webots `libController.so`, with final result `PRE_RENDEZVOUS_STACK_CAPTURED`. That evidence should be preserved on #87, but it does not make this branch integrable.

Smallest repair remains: remove the extra workflow and installed debugger. Any next experiment must use an authorized V4 CI surface and an actually available diagnostic mechanism without changing QPA/DISPLAY/product semantics.

djibian (commented):
NO_GO 21be76a58d27a70f7325f74c4e45f8df07a7a8d8

The two prior governance findings are resolved on this head, but the proposed experiment is now contradicted by the durable P12 evidence already recorded on #87.

Run 34146595043 established on this same current-main Runtime v2 path that the true controller sleeps inside `wb_robot_init()` and never reaches the former marker immediately after it. This head still calls `wb_qt_init_probe()` only **after** `wb_robot_init()`; the fatal handlers are therefore installed only after the already-proven blocking boundary. Nothing in this candidate changes that pre-probe path. A result of `PRE_QAPPLICATION_NOT_REACHED` can only reproduce P12 and cannot discriminate the preserved historical C8/C10 `QApplication` crash.

There is a second oracle weakness if the handler is ever reached: `FATAL_SIGNAL_BACKTRACE_CAPTURED` requires only signal + BEGIN/END markers, not a non-empty usable frame or a frame attributable to the Qt initialization boundary. #87's preserved requirement is an exact signal/exit plus a usable backtrace/first causal frame; marker presence alone is too weak for that claim.

Smallest repair: keep the experiment inside an authorized V4 CI surface, but replay the **frozen historical C10 source/environment** where the pre-`QApplication` rendezvous was already proven reachable, and make a positive fatal outcome require at least one captured frame sufficient to identify the causal boundary. Do not alter QPA/DISPLAY or install a debugger/dependency merely to obtain evidence.

djibian:
Abandoning this exact Draft candidate after authoritative `NO_GO 21be76a58d27a70f7325f74c4e45f8df07a7a8d8`.

The current-main probe path is now durably contradicted by P12: it does not reach the post-`wb_robot_init()` / pre-`QApplication` boundary and therefore cannot discriminate the preserved historical C8/C10 Qt crash. The earlier governance findings are already resolved, and their useful research evidence is preserved on #87.

A future #87 experiment, if still valuable, must be a new bounded candidate that replays the frozen historical C10 source/environment and requires a usable causal frame for a positive fatal outcome. This PR is not that candidate; keeping it open would preserve obsolete engaged work rather than useful progress.

PR #199, opened 2026-09-07
Prepare reference hardware capability checkpoint

djibian:
## Scope

Prepares a deterministic, checkpoint-only real reference-hardware capability observation path for #157 without requesting a second human test while #180 remains open.

- adds `physical-capabilities-readonly` to the trusted checkpoint profile allowlist;
- packages the exact WebeeBlocks non-authority probe, a pinned cflib source dependency and a one-command fail-closed reference-stack verifier;
- runs the existing physical capability probe/contract oracles before producing the checkpoint artifact;
- preserves the cflib safety-zero close-path qualification and grants no WebeeBlocks execution authority;
- extends the workflow contract and checkpoint documentation.

No CHECKPOINT_REQUEST is created by this PR. The existing single human-test slot remains authoritative.

Refs #157.

### DIFF
--- a/.github/workflows/human-checkpoint.yml
+++ b/.github/workflows/human-checkpoint.yml
@@ -62,10 +62,12 @@ jobs:
           PROFILES = {
               'windows-low-end': 'WebeeBlocks-Windows-R2025a',
               's3-props-off': 'experimental-s3-surface-offset-2026-08',
+              'physical-capabilities-readonly': 'WebeeBlocks-Physical-Capability-Probe',
           }
           PURPOSES = {
               'windows-low-end': {'checkpoint', 'release'},
               's3-props-off': {'checkpoint'},
+              'physical-capabilities-readonly': {'checkpoint'},
           }
 
           def get_commit(sha):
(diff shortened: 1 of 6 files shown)


djibian (commented):
NO_GO d1d216b3607688b9e44c99b1fe251ae7a264b650

The checkpoint artifact is not actually bound to the pinned cflib dependency claimed by this candidate.

`reference_probe_requirements.txt` names commit `45fdb784c9d13074c42835f3b5ac1d12133bf873`, but the preparation job only copies that requirements file into the artifact. `run_reference_probe.sh` never installs it, never creates an isolated environment, and never verifies the cflib used by `probe_reference_hardware.py`. On a test machine with any preinstalled cflib, the probe will silently use that version. The workflow-contract test proves only the requirement string, not the runtime dependency provenance.

That breaks the checkpoint requirement that the exact test artifact/support be deterministically prepared and provenance-bound.

Smallest repair: make the checkpoint bundle self-contained with the exact pinned cflib dependency (for example build/download a wheel/source archive for the pinned commit during preparation, include its digest, and have `run_reference_probe.sh` use an isolated temporary venv with only that local dependency and no network). Then test that the runner cannot fall back to an ambient cflib installation. Preserve `executionAuthority:false` and the documented safety-zero close-path qualification.

djibian (commented):
GO c47848c9d30ae8df3589b27888b6d17c9b76b390

Independent review of the exact Ready candidate.

The prior applicable NO_GO on `d1d216b3607688b9e44c99b1fe251ae7a264b650` is resolved on this HEAD: the checkpoint preparation now builds the exact cflib Git-commit dependency into the artifact wheelhouse, records commit provenance, includes all packaged files in `SHA256SUMS`, and the runner installs only from that local wheelhouse into an isolated target with `--no-index`, `--ignore-installed`, `PYTHONNOUSERSITE=1` and `python3 -S`. The preparation also exercises `--verify-environment` with a hostile ambient `cflib` on `PYTHONPATH`, establishing that ambient cflib cannot satisfy the packaged probe path.

The profile remains checkpoint-only, preserves `executionAuthority:false`, and explicitly retains the cflib close-path safety-zero qualification rather than claiming transport-level read-only behavior. Existing single-slot TEST_REQUIRED deduplication remains unchanged.

Canonical exact-head CI is successful: workflow `.github/workflows/ci.yml`, run 34147359293, effective attempt 2, final `CI Gate` successful. The active main ruleset still requires squash PR integration, strict `CI Gate`, blocks deletion/non-fast-forward, has zero native approving-review requirement, and no bypass actor.

No remaining counterexample found within this PR's bounded claim.

djibian:
LATE_NO_GO c47848c9d30ae8df3589b27888b6d17c9b76b390

This finding was discovered only after merge to `main@9b078ec5c2b902fa27dace59dc3f01d7d38cf6b6`, so it is trunk-health evidence and requires fix-forward rather than retroactive integration action.

The merged profile correctly isolates the cflib wheel from ambient packages, but its checkpoint preparation is not deterministic for a fixed WebeeBlocks SHA. The pinned cflib commit declares ranged runtime dependencies (`pyusb~=1.2`, `libusb-package~=1.0`, `scipy~=1.14`, `numpy~=2.2`, `packaging~=25.0`, `pyyaml>=6.0.3`) and ranged/unpinned build requirements (`setuptools>=61.0`, `wheel`, `setuptools_scm`). The checkpoint job runs `pip wheel` against the live package index, so two preparations of the same target SHA may resolve different wheelhouses/build tooling. `SHA256SUMS` makes the produced bundle tamper-evident, but does not make its preparation deterministic.

This conflicts with the current human-boundary invariant that an enabled TEST_REQUIRED profile has deterministic preparation before publication.

Smallest repair boundary: lock the complete offline cflib support closure, including build inputs, to exact artifacts/versions with integrity evidence (or an equivalently exact vendored/source closure), and retain the existing hostile-ambient isolation proof, `executionAuthority:false`, and documented cflib safety-zero close behavior. Until fixed, do not rely on `physical-capabilities-readonly` as a contract-conformant checkpoint preparation path.

djibian:
RESOLVED_LATE_NO_GO c47848c9d30ae8df3589b27888b6d17c9b76b390

The post-merge checkpoint-determinism finding is resolved on current trunk by merged PR #201.

Fix-forward source HEAD: `14083e008248c502d15aaf4b79286015dfc3a7e2`
Integrated main commit: `059c4b4fe4c0a813546c8623b7ad21cdf8629ec1`

The nondeterministic `physical-capabilities-readonly` profile and its live-resolution support preparation are no longer enabled. Unknown profile requests fail closed again. The earlier #193/#196 non-authority capability/preflight mechanism remains intact; only the invalid human-checkpoint enablement/support from #199 was removed.

The same fix-forward also closed the latent offline light-sensor CI resource leak exposed during restoration by disabling only the two non-semantic Pioneer wheel sounds in the job-local pinned Webots checkout. Exact-head CI Gate #604 and independent GO both passed before integration.

The late NO_GO therefore no longer makes current main unhealthy. Re-enabling a physical-capability checkpoint remains future #157 work and still requires a complete deterministic offline support closure.

PR #86, opened 2026-08-26
[Deferred research] Native Webots/Qt bridge for Firefox project files

djibian:
## Deferred — preserved for #87

This PR is intentionally **closed without merge** after the product sequencing decision of 26 August 2026.

#71 no longer requires Firefox parity for short-term classroom acceptance. The supported acceptance path is now Google Chrome on Linux; Firefox parity and the native Webots/Qt bridge are tracked in **#87**.

Nothing in this PR is declared production-ready and nothing from this branch should be merged into `webots-ci` merely to close #71.

## Preserved evidence

Final research head at deferral: `fa78eb7dc8fd423abfd7e656a3bbf8c3e7e26c1c`.
Base observed for C10: `webots-ci@5676e959d615999dc1d8bd775226fbb13742c3ee`.

Established across #86:

- official Webots R2025a desktop SDK contains Qt 6.5.3 headers/libs/moc/plugins;
- official Linux runtime dependency closure was pinned and exercised;
- controller-local `runtime.ini` with `WEBOTS_LIBRARY_PATH` gets the true Webots-launched controller to Webots-bundled Qt;
- the old `libQt6Core.so.6` loader failure was eliminated;
- `libqxcb.so` is loaded in the real Webots/Xvfb launch;
- the controller then crashes during `QApplication` initialization before `QT_APP_INITIALIZED`, before `QFileDialog`, and before the broker handshake;
- C8 run `32983674490` failed correctly on that real crash;
- C10 run `32989542721` proved the diagnostic mechanism currently stops fail-closed because `gdb` is absent from the runner; no signal/backtrace is fabricated.

The next bounded experiment, when #87 is explicitly reactivated, is to capture the true controller signal/exit and first usable causal backtrace in the unchanged Webots/Xvfb/runtime.ini environment.

## Status

`DEFERRED_RESEARCH / NOT_MERGED / FIREFOX_TRACK=#87`

Branch/history are retained as audit evidence. Do not automatically reopen this PR as the active #71 path.

`main` remains strictly out of scope.

PR #200, opened 2026-09-07
[Research] Capture Firefox native bridge Qt crash with LLDB

djibian:
## Scope

Reactivates the bounded diagnostic defined on #87 without reopening or rebasing historical PR #86.

This branch adds only an isolated research workflow. The workflow:
- checks out the exact frozen C10 research head `fa78eb7dc8fd423abfd7e656a3bbf8c3e7e26c1c`;
- preserves the established Webots R2025a + Xvfb + controller `runtime.ini` + Webots-bundled Qt environment;
- reuses the existing pre-`QApplication` SIGSTOP rendezvous from C10;
- uses an LLDB binary already available on the hosted runner if present, without installing a debugger or changing QPA/DISPLAY/runtime semantics;
- continues the true Webots-launched controller under LLDB and records the exact stopping signal plus all-thread backtrace;
- keeps the original product handshake intentionally unproven.

This is diagnostic evidence only. It does not merge the old native bridge into current Runtime v2 and does not change Firefox product semantics.

Refs #87.

### DIFF
--- a/.github/workflows/firefox-native-bridge-lldb-diagnostic.yml
+++ b/.github/workflows/firefox-native-bridge-lldb-diagnostic.yml
@@ -1,0 +1,405 @@
+name: Firefox native bridge LLDB causal diagnostic
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  HISTORICAL_C10_SHA: fa78eb7dc8fd423abfd7e656a3bbf8c3e7e26c1c
+  WEBOTS_ARCHIVE: webots-R2025a-x86-64.tar.bz2
+  WEBOTS_URL: https://github.com/cyberbotics/webots/releases/download/R2025a/webots-R2025a-x86-64.tar.bz2
+  WEBOTS_SHA256: c5127fb4206c57a5ae5523f1b7f3da8b670bc8926d9ae08595e139f226f38c38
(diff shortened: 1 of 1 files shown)


djibian (commented):
NO_GO 24bdb4d0334cde18245ba5cbcf6720858e608656

Two blockers refute this exact Draft research candidate.

1. The added `.github/workflows/firefox-native-bridge-lldb-diagnostic.yml` is outside the current V4 workflow inventory. Exact-head CI run 34146409457 fails in `Verify selector and repository contracts`; adding a one-off research workflow is not an eligible way around that contract.

2. The custom diagnostic run 34146409177 proves the hosted runner's visible `lldb` command is not a usable diagnostic mechanism for this experiment: invocation reports `ModuleNotFoundError: No module named 'lldb.embedded_interpreter'`, the expected C10 rendezvous is never captured, and the run times out without an exact signal/frame. `command -v lldb` is therefore insufficient evidence that LLDB is actually available for the preserved #87 boundary.

Smallest repair boundary: do not weaken the workflow inventory and do not treat a merely present debugger executable as usable. Reuse an already-authorized V4 CI surface for any further bounded experiment, and first prove the chosen in-run diagnostic mechanism can actually attach/capture evidence in that environment. Otherwise abandon this duplicate research branch; no product-semantic or Qt/QPA change is justified.

PR #202, opened 2026-09-07
Replay frozen C10 Qt crash boundary

djibian:
## Scope

Bounded #87 research successor to the abandoned #198 candidate.

This Draft changes only the already-authorized `.github/workflows/ci-webots.yml` surface and is intended to capture discriminating evidence, not to integrate product behavior.

- replays frozen C10 source at `fa78eb7dc8fd423abfd7e656a3bbf8c3e7e26c1c`;
- verifies exact historical `file_broker.cpp`, Makefile and `runtime.ini` blob identities before instrumentation;
- preserves the R2025a/Xvfb/Webots-bundled-Qt environment and does not alter QPA/DISPLAY;
- installs no debugger or new diagnostic dependency;
- replaces only C10's deliberate `SIGSTOP` rendezvous with in-process fatal-signal capture immediately before `QApplication`;
- accepts a fatal result only when a non-empty captured backtrace contains a Qt/QPA-attributable frame;
- also treats successful traversal through historical `QFileDialog` construction as discriminating evidence that the frozen crash no longer reproduces;
- fails closed for any other outcome.

The step is explicitly limited to this research branch and must not be interpreted as a production workflow addition. No Firefox file operation or product semantic is implemented.

This follows the authoritative repair boundary from `NO_GO 21be76a58d27a70f7325f74c4e45f8df07a7a8d8` and the durable P12 evidence on #87.

Refs #87.

### DIFF
--- a/.github/workflows/ci-webots.yml
+++ b/.github/workflows/ci-webots.yml
@@ -1194,7 +1194,22 @@ jobs:
 
           controller="$historical/controllers/crazyflie_runtime_v2/crazyflie_runtime_v2"
           test -x "$controller"
-          LD_LIBRARY_PATH="$diagnostic_webots/lib/webots" ldd "$controller" |
+          real_controller="$controller.real"
+          mv "$controller" "$real_controller"
+          cat > "$controller" <<'SH'
+          #!/usr/bin/env bash
+          set +e
+          real="${BASH_SOURCE[0]}.real"
+          "$real" "$@" &
+          pid=$!
+          printf 'WEBEEBLOCKS_C10_PROCESS pid=%s\n' "$pid" >&2
+          wait "$pid"
+          code=$?
+          printf 'WEBEEBLOCKS_C10_PROCESS_EXIT pid=%s code=%s\n' "$pid" "$code" >&2
+          exit "$code"
+          SH
+          chmod +x "$controller"
+          LD_LIBRARY_PATH="$diagnostic_webots/lib/webots" ldd "$real_controller" |
             tee "$artifact_dir/ldd.log"
           qt_lines="$(grep 'libQt6\(Core\|Gui\|Widgets\)' "$artifact_dir/ldd.log")"
           test "$(printf '%s\n' "$qt_lines" | wc -l)" -eq 3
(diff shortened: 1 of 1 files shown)


djibian (commented):
NO_GO 8862dd0915890c88a10f4a5cd72288eb7437ce7e

The historical-source replay is now at the right causal boundary and the Qt-frame oracle is materially stronger, but this exact Ready candidate still does not satisfy #87's preserved evidence requirement: **exact signal/exit plus a usable backtrace/first causal frame**.

The in-process handler records the signal and then calls `_exit(128 + signal_number)`, but the workflow never observes or verifies the historical controller process's resulting exit status. `webots_exit` records only the outer Webots/timeout command, and the positive fatal branch accepts `C10_FATAL_SIGNAL_WITH_USABLE_QT_FRAME` solely from signal/backtrace markers. A broken or intercepted controller termination could therefore still produce a positive result without the required exact exit evidence.

Smallest repair: keep the frozen C10 replay and current usable Qt-frame requirement, but durably capture the exact historical controller PID and independently observe/verify its terminal status (or another equivalently direct process-exit oracle) against the captured signal. Do not substitute the outer Webots exit code, alter QPA/DISPLAY, or install a debugger/dependency.

djibian (commented):
NO_GO 6089ac64ba14867e7332420677b133d64a6d3a35

Exact-head CI refutes this candidate before the C10 experiment runs. In run 34151630320, the instrumented frozen file_broker.cpp fails to compile because Qt defines the identifier `signals` as a macro expanding to `public`. The declaration `const int signals[]` and the range loop are therefore syntactically rewritten by Qt headers.

Smallest repair: rename that local array to a Qt-safe identifier such as `fatal_signals`, without changing the signal set or any other experimental boundary. Then require a fresh exact-head run; the previous process-exit repair remains otherwise applicable.

djibian:
C13 achieved the bounded research objective on exact predecessor `2e4d3d93eb0f918c9e0b4c058949999cc967fee7`, run `34151983271`: frozen C10 reproduced a directly observed `SIGSEGV` with controller exit `139` and a usable Qt/XCB initialization backtrace. The causal evidence is now durably recorded on #87.

Current Draft `dfb1a81da199f1d12220ef19fad1222449a490a2` only synchronizes the already-integrated light-sensor CI restoration from current main; it does not add a new causal hypothesis or product result. The branch-only diagnostic was explicitly research instrumentation and should not become production workflow code.

The useful unique result is therefore preserved on #87; continuing this PR would duplicate the already-completed experiment rather than advance Firefox file semantics. Closing this research candidate without merge.

djibian:
Research result complete; do not merge this diagnostic branch.

The bounded C10 experiment has produced and durably recorded its discriminating result on #87 as C13: exact SIGSEGV, exact controller exit 139, and a usable Qt/XCB stack through QGuiApplication/QXcbConnection initialization. The workflow addition is branch-gated research instrumentation, not product behavior and not a mainline capability.

The current Draft sync only carries the already-merged main light-sensor CI repair so the research branch remains coherent; it creates no unique production value. Preserve this PR/branch history as evidence and close the bounded research candidate rather than integrating diagnostic-only workflow code.

PR #203, opened 2026-09-07
Replay frozen C10 Qt crash boundary

djibian:
## Scope

Clean current-main successor for #87 after trunk-health restoration. This replaces the closed stale-base #202 line without changing the bounded research claim.

- exact base: current `main@059c4b4fe4c0a813546c8623b7ad21cdf8629ec1`;
- replays frozen C10 source `fa78eb7dc8fd423abfd7e656a3bbf8c3e7e26c1c` and verifies historical source/blob identities;
- preserves R2025a/Xvfb/Webots-bundled Qt and does not alter QPA/DISPLAY;
- installs no debugger/new diagnostic dependency;
- replaces only the frozen C10 SIGSTOP rendezvous with in-process fatal-signal/backtrace capture before `QApplication`;
- records the exact historical controller PID and independently verifies its terminal exit status against the captured fatal signal;
- requires a non-empty Qt/QPA-attributable causal backtrace for a fatal positive result, or successful traversal through historical `QFileDialog` construction;
- fails closed for every other outcome.

The current-main historical light-sensor offline fix is preserved. No product Firefox file operation, flight behavior, or human checkpoint is introduced.

Supersedes closed #202. Refs #87.

### DIFF
--- a/.github/workflows/ci-webots.yml
+++ b/.github/workflows/ci-webots.yml
@@ -1056,7 +1056,7 @@ jobs:
           PY
 
       - name: Run frozen C10 Firefox Qt signal diagnostic
-        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref == 'research/firefox-c10-signal-frame' }}
+        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref == 'research/firefox-c10-signal-frame-main' }}
         shell: bash
         run: |
           set -euo pipefail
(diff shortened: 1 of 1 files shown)


djibian:
This current-main replay is now redundant with the durable C13 result already recorded on #87 from exact candidate `2e4d3d93eb0f918c9e0b4c058949999cc967fee7`, run `34151983271`: frozen C10 reached the pre-`QApplication` boundary, received `SIGSEGV`, exited 139, and produced a usable Qt/XCB initialization stack. Repeating the same C10 hypothesis cannot change the next decision. The next discriminating #87 work is the standalone bundled-Qt control in #204, which separates the Qt/Xvfb substrate from the Webots-controller lifecycle. Close this duplicate research candidate without merge.

PR #205, opened 2026-09-07
Discriminate frozen C10 Qt crash with standalone control

djibian:
## Scope

Bounded #87 causal experiment after C13 established the exact historical `SIGSEGV`/exit-139 Qt/XCB stack.

This candidate tests the smallest next hypothesis without implementing Firefox file operations and without changing QPA/DISPLAY:

- pins the same frozen C10 source (`fa78eb7dc8fd423abfd7e656a3bbf8c3e7e26c1c`) and exact historical source blobs;
- uses the same pinned Webots R2025a desktop Qt/runtime closure and no debugger/new diagnostic dependency;
- constructs a minimal standalone `QApplication` + `QFileDialog` control against the same Webots-bundled Qt;
- runs that standalone control and the instrumented frozen C10 Webots controller sequentially under the **same `xvfb-run` session / DISPLAY**;
- both paths have fail-closed exact signal/backtrace/exit evidence;
- classifies only two discriminating outcomes: standalone Qt reaches `QFileDialog` while C10 reproduces the Qt/XCB crash, or both paths reproduce a Qt-attributable fatal crash; any other comparison is UNPROVEN.

The goal is to distinguish whether the R2025a Qt/Xvfb substrate alone is sufficient to reproduce the crash or whether some additional C10/Webots-controller process context is required. It does not alter product semantics, install a debugger, switch QPA/DISPLAY, or propose a causal fix before this control settles.

The CI addition is branch-gated research instrumentation and is not intended for integration into production workflow code. Preserve the result on #87, then close this research PR without merge once the bounded evidence is complete.

Refs #87.

### DIFF
--- a/.github/workflows/ci-webots.yml
+++ b/.github/workflows/ci-webots.yml
@@ -1055,6 +1055,19 @@ jobs:
           print('PASS: Blockly Submit -> one WWI mission -> validation -> shared STOP executor -> L course; persistent runtime rearmed afterwards.')
           PY
 
+      - name: Run C14 standalone Qt causal control
+        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref == 'research/firefox-c14-standalone-qt-control' }}
+        shell: bash
+        run: bash tools/ci/firefox_c14_standalone_qt_control.sh
+
+      - name: Upload C14 standalone Qt causal control
+        if: ${{ always() && github.event_name == 'pull_request' && github.event.pull_request.head.ref == 'research/firefox-c14-standalone-qt-control' }}
+        uses: actions/upload-artifact@v4
+        with:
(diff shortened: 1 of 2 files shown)


djibian:
This candidate duplicates the already-engaged smaller #204 experiment. Both test the C13 next causal distinction with pinned Webots R2025a bundled Qt/Xvfb and no QPA/DISPLAY change; #204 does so in one authorized-workflow-only change and directly answers whether standalone `QApplication` initialization reproduces the C13 crash boundary. The extra same-Xvfb/C10 replay machinery here is not required for that next decision because C13 already durably proves the frozen-controller side.

Preserve this branch history, but prefer #204 as the smallest useful complete experiment. Closing #205 without merge rather than spending CI/review on a dominated duplicate.

PR #212, opened 2026-09-07
Isolate libController load from C14 Qt control

djibian:
## Scope

Bounded #87 C17 research implementing the pre-registered discriminator on issue #87.

C14 proves the minimal standalone QApplication succeeds under the pinned Webots R2025a Qt/XCB + Xvfb closure. C16 proves the exact frozen C10/C15 controller image still SIGSEGVs in the same Qt/XCB initialization class when executed directly, without the Webots launcher and before wb_robot_init.

This candidate changes only the C14 standalone binary-load boundary:
- reuses the exact C14 minimal QApplication source and its normal argv behavior;
- preserves the pinned R2025a archive/runtime recipe, Webots-bundled Qt/plugins, Xvfb, software rendering and unchanged QPA/DISPLAY;
- forces `libController.so` into the standalone executable's direct `DT_NEEDED` set using `--no-as-needed`;
- verifies the direct dependency with `readelf -d` and its pinned Webots resolution with `ldd`;
- calls no Controller/Webots API and performs no wb_robot_init, QFileDialog, debugger or product file operation;
- retains exact child PID/exit plus fail-closed Qt fatal-frame evidence.

Discriminator:
- Qt/XCB fatal crash -> loading libController is sufficient relative to C14;
- QApplication success/exit 0 -> libController loading alone is insufficient;
- anything else -> UNPROVEN.

Research-only. Preserve the exact result on #87 and close this PR without merge.

Refs #87.

### DIFF
--- a/.github/workflows/ci-webots.yml
+++ b/.github/workflows/ci-webots.yml
@@ -1055,6 +1055,187 @@ jobs:
           print('PASS: Blockly Submit -> one WWI mission -> validation -> shared STOP executor -> L course; persistent runtime rearmed afterwards.')
           PY
 
+      - name: Run C17 C14 standalone with forced libController load
+        if: \${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref == 'research/firefox-c17-libcontroller-load' }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          artifact_dir=ci-artifacts/firefox-c17-libcontroller-load
+          mkdir -p "$artifact_dir"
+
+          recipe="$RUNNER_TEMP/linux_runtime_dependencies-R2025a.sh"
+          recipe_url="https://raw.githubusercontent.com/cyberbotics/webots/R2025a/scripts/install/linux_runtime_dependencies.sh"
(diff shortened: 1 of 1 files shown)


djibian (commented):
NO_GO 4ae7e5e60eb6597023316e33bf9e268fece96079

The bounded C17 discriminator is otherwise aligned with the durable #87 boundary, but both new branch-gating expressions are escaped in the committed workflow: `if: \${{ ... }}` appears on the diagnostic step and on its artifact upload. That is not the repository's normal GitHub Actions expression form and no longer proves that this expensive research path is confined to `research/firefox-c17-libcontroller-load`. On a reusable Webots workflow this is a trust/scope defect: the experiment must fail closed outside its exact research branch.

Smallest repair: return the PR to Draft and remove only the literal backslash before `${{` in both C17 `if:` lines, leaving the exact C14 source, forced `libController.so` DT_NEEDED/readelf+ldd oracle, runtime inputs and result oracle unchanged. Then mark the repaired exact HEAD Ready and require fresh exact-head CI/review.

PR #214, opened 2026-09-07
Isolate frozen QApplication argv0 from C14 control

djibian:
## Scope

Bounded #87 C18 research following the settled C17 result.

C17 proved that forcing pinned Webots R2025a `libController.so` into the successful C14 standalone process is insufficient to reproduce the historical Qt/XCB crash. The next smaller difference is the argv supplied to `QApplication`: frozen C10 uses `argc=1` with `argv[0]="webeeblocks-file-broker"`, while C14 uses the process-provided argv.

This candidate keeps the exact C14 standalone QApplication source and one exact compiled binary/path. Inside one Xvfb session it runs:
- baseline: the same binary with its normal path as argv0 and no extra arguments;
- treatment: the same binary/path with only argv0 replaced by `webeeblocks-file-broker` via `exec -a`, still with no extra arguments.

The pinned R2025a runtime, Webots-bundled Qt/XCB, software rendering and QPA/DISPLAY boundary remain unchanged. No libController, Controller API, wb_robot_init, QFileDialog, debugger or product file operation is introduced. The oracle binds exact child PID/exit and fails closed outside the pre-registered success/fatal classes.

Discriminator:
- treatment reproduces the same Qt/XCB fatal frame/exit -> frozen argv0 is sufficient relative to C14;
- both baseline and treatment reach QAPPLICATION_OK and exit 0 -> frozen argv0 alone is insufficient;
- otherwise -> UNPROVEN.

Research-only. Preserve the exact settled result on #87 and close without merge.

Refs #87.

### DIFF
--- a/.github/workflows/ci-webots.yml
+++ b/.github/workflows/ci-webots.yml
@@ -1055,6 +1055,229 @@ jobs:
           print('PASS: Blockly Submit -> one WWI mission -> validation -> shared STOP executor -> L course; persistent runtime rearmed afterwards.')
           PY
 
+      - name: Run C18 exact-binary QApplication argv0 discriminator
+        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref == 'research/firefox-c18-qapplication-argv0' }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          artifact_dir=ci-artifacts/firefox-c18-qapplication-argv0
+          mkdir -p "$artifact_dir"
+
+          recipe="$RUNNER_TEMP/linux_runtime_dependencies-R2025a.sh"
+          recipe_url="https://raw.githubusercontent.com/cyberbotics/webots/R2025a/scripts/install/linux_runtime_dependencies.sh"
+          recipe_blob=4593476be2c985580a0e1738671f4b5bae81f669
+          curl --fail --location --retry 3 --output "$recipe" "$recipe_url"
+          test "$(git hash-object "$recipe")" = "$recipe_blob"
+          sudo env CI=true bash "$recipe"
+
+          archive="$RUNNER_TEMP/webots-R2025a-x86-64.tar.bz2"
+          archive_url="https://github.com/cyberbotics/webots/releases/download/R2025a/webots-R2025a-x86-64.tar.bz2"
+          archive_sha=c5127fb4206c57a5ae5523f1b7f3da8b670bc8926d9ae08595e139f226f38c38
+          curl --fail --location --retry 3 --output "$archive" "$archive_url"
+          echo "$archive_sha  $archive" | sha256sum --check
+          tar -xjf "$archive" -C "$RUNNER_TEMP"
+          diagnostic_webots="$RUNNER_TEMP/webots"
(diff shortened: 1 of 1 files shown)


djibian (commented):
NO_GO bb01b79572ed92d8897c106ce612afaba2a69478

The one-variable argv0 setup is now structurally sound: the same compiled C14 control binary is run twice in one Xvfb session, both arms keep QT_DEBUG_PLUGINS unset, and the treatment changes only argv[0] to `webeeblocks-file-broker`. One fail-closed classification mismatch remains.

The authoritative C18 discriminator on #87 requires the treatment to reproduce the historical **SIGSEGV/139** Qt/XCB failure before declaring the frozen argv0 sufficient. This HEAD still accepts SIGSEGV, SIGABRT, SIGBUS, SIGILL or SIGFPE, computes the corresponding exit code, and emits `DIAGNOSTIC_RESULT=C18_FROZEN_ARGV0_SUFFICIENT` for any of those fatal classes with a Qt/QPA frame.

Smallest repair: preserve the exact binary/path/session/environment and success oracle, but make the sufficient branch require exactly `WEBEEBLOCKS_QT_CONTROL_FATAL signal=SIGSEGV`, exact same-child exit `139`, and the existing non-empty Qt/XCB-attributable backtrace. Any other fatal signal must remain `C18_ARGV0_OUTCOME_UNPROVEN` and fail closed.

djibian (commented):
GO 4edeb311cfa16d3098030ee8fe58fdb3f3a0694c

Independent falsification review found the bounded argv0 discriminator sound.

- The prior NO_GO on bb01b79572ed92d8897c106ce612afaba2a69478 is resolved: the sufficient branch now requires the exact `SIGSEGV` marker, exact child exit `139`, and a non-empty Qt/XCB-attributable backtrace. Other fatal classes fall through to UNPROVEN.
- Baseline and treatment execute the same compiled binary/path in one Xvfb session; both keep `QT_DEBUG_PLUGINS` unset, while treatment changes only effective `argv[0]` to `webeeblocks-file-broker`.
- The quoted child harness and post-wait metadata recording preserve the intended runtime variables and bind each observed exit to its exact child PID without the earlier log-truncation race.
- Pinned R2025a archive/runtime identities, bundled Qt resolution, unchanged QPA/DISPLAY/software-rendering boundary, binary hash equality, and absence of `libController.so` are fail-closed.
- Canonical Ready run 34173223871 and final CI Gate job 101898123846 completed successfully for this exact HEAD. Diagnostic job 101897514178 established baseline PID 11012 exit 0 and treatment PID 11017 exit 0 on display :99, with `DIAGNOSTIC_RESULT=C18_FROZEN_ARGV0_INSUFFICIENT`.
- Artifact `firefox-c18-qapplication-argv0-r2025a` id 10036430110 has digest `sha256:db5bf2ec2031cff3f15efa822819bc616b1182518c58d7dad0ab5b3069d9b425`.

The result supports only the pre-registered research conclusion: the frozen QApplication argv0 value alone is insufficient to reproduce the historical Qt/XCB crash. It does not establish Firefox same-file semantics or a product fix. Preserve the exact result on #87 and close this research PR without merge.

PR #216, opened 2026-09-08
Reproduce file-broker object crash across fresh Xvfb order

djibian:
## Scope

Bounded #87 C20 research after the settled C19 result.

C19 associated the exact frozen C10 file-broker translation unit with the historical Qt/XCB SIGSEGV/139 in one baseline -> treatment Xvfb session. Before decomposing that surprising object/link effect, C20 checks that it is reproducible and independent of run order.

The exact C19-style baseline and treatment binaries are built once under the pinned Webots R2025a bundled Qt/XCB closure with exact libController and exact frozen broker source blobs. They are then run from the same staged path/argv in two fresh Xvfb sessions:
- AB: baseline -> treatment;
- BA: treatment -> baseline.

Each baseline must reach QAPPLICATION_OK and exit 0. Each treatment must reproduce SIGSEGV/139 with a non-empty Qt/XCB-attributable backtrace before any broker/provider construction marker. Exact binary hashes, dynamic sections, DT_NEEDED lists and ldd resolution are preserved as evidence.

Final discriminator:
- both baselines succeed and both treatments reproduce the exact crash class across AB and BA => C20_BROKER_OBJECT_REPRODUCED;
- any order dependence, treatment success, baseline failure, different fatal class, missing identity or other inconsistent outcome => UNPROVEN.

The previously recorded section-retention/--gc-sections discriminator is deliberately deferred to contingent C21 per the newer #87 execution alignment; it is not executed by this candidate.

No Controller API, wb_robot_init, broker construction, QFileDialog invocation, debugger, QPA/DISPLAY change or product file operation. Research-only; preserve the exact result on #87 and close without merge.

Refs #87.

PR #217, opened 2026-09-08
Isolate retained broker sections from C19

djibian:
## Scope

Bounded #87 C20 research after the settled C19 result.

C19 proved that adding the exact frozen C10 `file_broker.cpp` translation unit to the successful C17-style standalone `QApplication` + exact `libController.so` control is sufficient to reproduce the historical Qt/XCB SIGSEGV/139 even though no broker/provider code runs.

C20 isolates whether the **retained broker code/data sections and their load-time relocations** are necessary for that effect. It preserves the pinned Webots R2025a/Xvfb/Qt/Controller boundary and the exact frozen broker source blobs, then runs three same-path arms in one Xvfb session:

- baseline: C17-style standalone control;
- full: exact C19-style broker translation unit retained;
- gc: the same full build with only linker dead-section collection (`--gc-sections`) added.

The oracle requires the full arm to reproduce the exact C19 SIGSEGV/139 Qt/XCB witness before interpreting the gc arm, and requires the gc image to contain neither `webeeblocks::createQtFileDialogProvider()` nor the broker runtime marker.

Pre-registered discriminator:
1. baseline succeeds, full reproduces exact SIGSEGV/139, gc reaches `QAPPLICATION_OK`/0 -> retained broker sections/relocations are necessary; next isolate the smallest retained section/symbol group without invoking broker behavior;
2. baseline succeeds, full reproduces exact SIGSEGV/139, gc still reproduces the same Qt/XCB SIGSEGV despite broker symbol/marker elimination -> dead-section elision is insufficient; inspect residual ELF/layout/load metadata before source-level behavior;
3. C19 witness not reproduced, gc residue remains, or any other outcome -> UNPROVEN.

Research-only. Do not change QPA/DISPLAY or invoke `QFileDialog`. Preserve the exact settled result on #87 and close without merge.

Refs #87.

### DIFF
--- a/.github/workflows/ci-webots.yml
+++ b/.github/workflows/ci-webots.yml
@@ -1055,6 +1055,189 @@ jobs:
           print('PASS: Blockly Submit -> one WWI mission -> validation -> shared STOP executor -> L course; persistent runtime rearmed afterwards.')
           PY
 
+      - name: Run C20 broker dead-section discriminator
+        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref == 'research/firefox-c20-broker-gc' }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          artifact_dir=ci-artifacts/firefox-c20-broker-gc
+          mkdir -p "$artifact_dir"
+          recipe="$RUNNER_TEMP/linux_runtime_dependencies-R2025a.sh"
+          curl --fail --location --retry 3 --output "$recipe" \
+            https://raw.githubusercontent.com/cyberbotics/webots/R2025a/scripts/install/linux_runtime_dependencies.sh
+          test "$(git hash-object "$recipe")" = 4593476be2c985580a0e1738671f4b5bae81f669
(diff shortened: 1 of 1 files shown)


djibian (commented):
NO_GO 59535f05f8b52ada539ee4505cf3d902da14ca21

This candidate does not execute the durable #87 C20 discriminator. The pre-registration requires the standalone source and frozen broker object to be compiled with `-ffunction-sections -fdata-sections`, then B and C to link the exact same sectioned objects/library order with C changing only `--gc-sections`.

Here, baseline/full/gc are compiled directly from source without either section compiler flag, and the gc arm merely adds `--gc-sections` to a fresh source-to-executable build. Therefore provider-symbol/marker disappearance, if observed, would not establish the pre-registered dead-section experiment on identical sectioned objects.

Smallest repair boundary would be to compile the control and broker once into sectioned objects and link A/B/C from those exact objects. That repair is already represented by the current C20 path on PR #216, so duplicating it here is not useful unless contradictory evidence emerges.

PR #218, opened 2026-09-08
Isolate retained broker sections from C20 crash

djibian:
## Scope

Bounded #87 C21 research after the settled C20 reproducibility result.

C20 established on exact candidate `f6a7bb7fa7db44c081bbd007b1dc0cd0af328776` that the C19 broker-object association reproduces in both fresh Xvfb session orders while the standalone baseline remains successful. C21 now executes the pre-registered next discriminator from #87: whether otherwise-unreferenced code/data retained from the exact frozen C10 broker translation unit is necessary for the Qt/XCB crash class.

The candidate preserves the exact C20/C19 standalone `QApplication` source and pinned Webots R2025a runtime/Qt/XCB closure, exact forced `libController.so`, same staged executable path/argv, software rendering and unchanged QPA/DISPLAY. It fetches the exact frozen broker sources from `fa78eb7dc8fd423abfd7e656a3bbf8c3e7e26c1c` and verifies the three proven blob identities.

The control source and broker translation unit are each compiled once with `-ffunction-sections -fdata-sections`, then linked as:
- A: sectioned standalone control only, normal link;
- B: exact control + exact broker object, normal link;
- C: exact same objects and library order as B, changing only `--gc-sections` plus output-only link-map evidence.

Before execution classification, B must retain the broker provider symbol class and frozen `QT_APP_INITIALIZED` marker, while C must prove those provider symbols/marker absent after dead-section collection. A/C/B then run from the same staged path in one fresh Xvfb session, with B last so its intended crash cannot perturb C.

Pre-registered discriminator:
- A succeeds, B reproduces exact Qt/XCB `SIGSEGV`/139, C succeeds after proven dead-section elimination -> `C21_DEAD_SECTIONS_NECESSARY`;
- A succeeds, B reproduces exact crash, C still reproduces it after proven provider-symbol/marker elimination -> `C21_SURVIVES_GC`;
- failed controls/removal proof, different fatal class or any other outcome -> UNPROVEN.

No Controller API, `wb_robot_init()`, broker/provider construction, `QFileDialog`, debugger, QPA/DISPLAY change or product file operation. Research-only. Preserve the exact settled result on #87 and close without merge.

Refs #87.

PR #219, opened 2026-09-08
Stabilize student UI tooltip hover oracle

djibian:
## Problem

Canonical C21 run `34186940578` exposed a recurrent false-negative in the unchanged `Runtime suite / student-ui` oracle. Attempt 1 (`101937095661`) and failed-jobs attempt 2 (`101937941390`) both reached the real Blockly UI but failed at the same boundary: the repeat-block tooltip never became observable after `Escape` closed the preceding real direction dropdown. The exact same unchanged student-UI suite passed moments earlier on canonical C20 run `34185240716` (`101932174289`), and the C21 change itself is confined to a branch-gated Webots workflow step.

The failure mechanism is a UI-event race in the probe: it waited a fixed 200 ms after closing the dropdown, then hovered a hit-tested repeat-block coordinate captured *before* the dropdown had opened. If the transient Blockly widget overlay still owns that coordinate when the synthetic mouse move occurs, the overlay can consume the entry event; once it disappears, the pointer is already stationary over the block and no real block hover is generated.

## Fix

Preserve the existing tooltip product oracle and its 5 s observation bound. Change only the interaction preparation:
- after `Escape`, poll the existing real-overlay observation until the non-tooltip dropdown/widget overlay is actually closed, failing explicitly if it does not close within 2 s;
- re-evaluate the existing real rendered-block hit test after closure;
- hover that fresh repeat-block point;
- keep the same requirement that a non-empty, non-English real tooltip becomes visible within the original 5 s bound.

This does not weaken the product property, extend its tooltip timeout, fabricate a tooltip, or bypass real Blockly/Webots interaction. It removes a stale-coordinate/transient-overlay race in the evidence harness.

Refs #218.

### DIFF
--- a/experiments/runtime-v2-student-ui/probe.py
+++ b/experiments/runtime-v2-student-ui/probe.py
@@ -344,8 +344,23 @@ def main():
         if expected not in menu_text:
             raise RuntimeError('real direction dropdown lacks '+expected+': '+menu_text)
     c.screenshot(screenshot.with_name('direction-menu-1366x768.png'))
-    c.key('Escape'); time.sleep(.2)
-    c.hover(rendered['repeatRect'])
+    c.key('Escape')
+    blocking_overlay=[]
+    overlay_end=time.time()+2.0
+    while time.time()<overlay_end:
+        blocking_overlay=[
+            entry for entry in c.eval(VISIBLE_OVERLAY)
+            if 'tooltip' not in entry['className'].lower() and entry['text'].strip()
+        ]
+        if not blocking_overlay: break
+        time.sleep(.05)
+    if blocking_overlay:
+        raise RuntimeError('direction dropdown overlay did not close before tooltip hover: '+json.dumps(blocking_overlay,ensure_ascii=False))
+    # Re-hit-test after the dropdown has actually closed. A stale pre-menu
+    # coordinate can remain under the transient widget overlay, causing the
+    # synthetic mouse move to be consumed without ever entering the real block.
+    repeat_after_close=c.eval(RENDER_LOCALE)
+    c.hover(repeat_after_close['repeatRect'])
     tooltip=[]
     end=time.time()+5.0
     while time.time()<end:
--- a/experiments/runtime-v2-student-ui/probe.py
+++ b/experiments/runtime-v2-student-ui/probe.py
@@ -150,6 +150,23 @@ RENDER_LOCALE=r'''(() => {
  };
 })()'''
 
+REPEAT_HOVER_RECT=r'''(() => {
+ const repeat=workspace.getBlocksByType('controls_repeat_ext',false)[0];
+ if(!repeat)throw new Error('rendered repeat block missing for hover');
+ const repeatRoot=repeat.getSvgRoot();
+ if(!repeatRoot)throw new Error('rendered repeat SVG root missing for hover');
+ const rb=repeatRoot.getBoundingClientRect();
+ for(let y=Math.ceil(rb.top)+2;y<Math.floor(rb.bottom);y+=4){
+   for(let x=Math.ceil(rb.left)+2;x<Math.floor(rb.right);x+=4){
+     const hit=document.elementFromPoint(x,y);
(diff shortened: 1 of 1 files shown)


djibian (commented):
NO_GO b19c7a2c8c934b9135a3e2ce6b81f2228dd1cd68

The proposed race repair preserves the 5 s tooltip oracle and correctly waits for the transient Blockly overlay to close, but its fresh hit test is not read-only. It calls `c.eval(RENDER_LOCALE)` a second time after closing the dropdown. `RENDER_LOCALE` begins by calling `Blockly.Xml.domToWorkspace(..., workspace)`, so this evidence-harness step inserts another set of locale-evidence blocks into the real workspace merely to obtain a new repeat-block coordinate.

That creates an avoidable workspace mutation in the middle of the oracle and can perturb the subsequent real profile/flyout checks. It also means the fix is not bounded to interaction preparation as claimed.

Smallest repair boundary: keep the overlay-close polling and the original non-empty/non-English 5 s tooltip requirement, but replace the second `RENDER_LOCALE` invocation with a dedicated read-only repeat-block hit-test expression that only locates the existing `controls_repeat_ext` SVG root and returns a current `elementFromPoint`-validated hover rectangle. Do not call `domToWorkspace` or otherwise mutate Blockly state during that re-hit-test.

djibian (commented):
GO bda4a2aaf2011c986b5e25346eb648c78192a35c

Independent review accepts the exact Ready repair.

The prior `NO_GO b19c7a2c8c934b9135a3e2ce6b81f2228dd1cd68` is resolved on this HEAD: the post-dropdown re-hit-test is now a dedicated read-only `REPEAT_HOVER_RECT` expression. It locates the already-rendered `controls_repeat_ext` block and performs only `getSvgRoot()`, geometry and `elementFromPoint` inspection; it does not call `domToWorkspace` or otherwise mutate Blockly state.

The repair remains bounded to evidence interaction preparation:
- the direction overlay is observed closed before hover;
- a fresh real hit-tested coordinate is acquired only after closure;
- the existing real hover interaction is preserved;
- the tooltip must still become visible and non-empty within the unchanged 5 s bound and must remain non-English;
- no product/runtime source or semantic oracle is weakened.

Canonical Ready `CI Gate` run `34187634969` (#671) completed success on this exact HEAD. In particular, `Runtime suite / student-ui` job `101939105748` completed success through the real styled Zelos Robot Window and its follow-on visual/AST/keyboard validation, while the full selected matrix completed without a failing job.

This is a CI-evidence stabilization only; it establishes no new product behavior.

djibian (commented):
GO bda4a2aaf2011c986b5e25346eb648c78192a35c

Independent review accepts this exact Ready candidate.

The prior NO_GO on b19c7a2c8c934b9135a3e2ce6b81f2228dd1cd68 is resolved. That finding was specifically that the post-dropdown fresh hit test called RENDER_LOCALE again and therefore mutated the real Blockly workspace through domToWorkspace. This HEAD replaces that second mutation with REPEAT_HOVER_RECT, a read-only query that obtains the existing controls_repeat_ext SVG root, re-hit-tests the current rendered geometry through elementFromPoint, and returns only a fresh hover point.

The repair remains bounded to evidence interaction preparation:
- Escape still closes the real direction dropdown;
- the probe now polls the existing real-overlay observation for at most 2 s and fails explicitly if a non-tooltip widget/dropdown remains;
- no tooltip timeout is extended: the original 5 s non-empty/non-English real-tooltip oracle is unchanged;
- no fabricated tooltip, bypass, product UI change, or Blockly workspace mutation is introduced by the fresh hit test;
- the later profile/flyout and AST assertions therefore still execute against the intended workspace state.

Canonical Ready CI Gate run 34187634969 completed success on this exact HEAD; Runtime suite / student-ui job 101939105748 passed the real Webots/Blockly interaction, and final CI Gate job 101939756438 succeeded.

This fixes the demonstrated stale-coordinate/transient-overlay race in the test harness without weakening the product property.

PR #224, opened 2026-09-08
C26 isolate QCoreApplication::self COPY versus GOT binding

djibian:
Research-only C26 discriminator for #87. Do not merge; preserve the exact settled result durably on #87, then close without merge.

Pre-registration: #87 issue comment 5583763904.

Bounded claim: distinguish whether the `QCoreApplication::self` COPY relocation/data-interposition class is required relative to a GOT-bound reference to the same Qt data symbol, while preserving the settled C25 controls and pinned R2025a closure.

This PR is opened Draft for review of the exact current branch state before it is offered as a stable research candidate.

### DIFF
--- a/tools/ci/prepare_crazyflie_runtime_wwi.py
+++ b/tools/ci/prepare_crazyflie_runtime_wwi.py
@@ -143,9 +143,9 @@ script = r'''
 html_path.write_text(html + '\n' + script, encoding='utf-8')
 print('Prepared Blockly Robot Window runtime WWI harness.')
 
-if os.environ.get('GITHUB_HEAD_REF') == 'research/firefox-c25-direct-self':
+if os.environ.get('GITHUB_HEAD_REF') == 'research/firefox-c26-copy-vs-got':
     subprocess.run(
-        ['bash', str(ROOT / 'tools/ci/run_firefox_c25_qcore_self_copy.sh')],
+        ['bash', str(ROOT / 'tools/ci/run_firefox_c26_copy_vs_got.sh')],
         cwd=ROOT,
         check=True,
     )
--- a/tools/ci/prepare_crazyflie_runtime_wwi.py
+++ b/tools/ci/prepare_crazyflie_runtime_wwi.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 import json
+import os
+import subprocess
 from pathlib import Path
 
 ROOT = Path(__file__).resolve().parents[2]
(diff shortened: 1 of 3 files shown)


djibian (commented):
NO_GO 52e15ca97048ccee7b15e885952292c494526edc

The exact C26 pre-registration on #87 requires A/I/K/F to be compared under one `--no-relax` link closure so K's GOTPCREL reference cannot be silently relaxed while the COPY-vs-GOT mechanism is being isolated. This candidate does not pass `-Wl,--no-relax` to K, and A/I/F are reused from the settled C25 build without that flag. The static `! R_X86_64_COPY` check on K can fail closed if relaxation produces COPY, but it does not implement the pre-registered controlled comparison; a successful K classification would therefore not establish the stated C26 discriminator under the registered boundary.

Smallest repair: rebuild all C26 comparison arms A, K, I and F from the same settled C25 objects with identical `--gc-sections --no-relax` linkage (changing only the intended root/object per arm), preserve the existing exact static relocation/root/DT_NEEDED checks, and classify only after A/I/F reproduce their settled controls under that common closure. Do not alter QPA/DISPLAY or product behavior.

djibian (commented):
NO_GO 6e7d186c3410c775cbf8c73a0a61bf2b56c3087f

The common `--no-relax` linkage now repairs the prior finding, but the exact pre-registration on #87 also requires the structural controls to fail closed on the *absence* of the broader metaobject path and on A having no `QCoreApplication::self` dynamic relocation at all. This HEAD only checks `qobject_cast<QApplication*>` and `R_X86_64_COPY.*QCoreApplication::self`; it never rejects `QApplication::staticMetaObject` retention/COPY in A/I/K/F, and A would still pass if `QCoreApplication::self` appeared through a non-COPY dynamic relocation.

That leaves the discriminator under-characterized: K could be classified as a clean GOT-only self binding even if an unintended `QApplication::staticMetaObject` relocation survived, and A could cease to be the registered no-self-binding control without the oracle noticing.

Smallest repair: preserve the current link/runtime setup, add an exact `QApplication::staticMetaObject` relocation/root negative oracle for A/I/K/F (especially K), and require A's relocation table to contain no `QCoreApplication::self` reference of any relocation type. Then re-offer a new exact HEAD; do not alter QPA/DISPLAY or runtime semantics.

djibian (commented):
GO 56176deae3abd6c8b8f4301aa6aa3f695d420660

Independent review against #87 pre-registration 5583763904 and the current-main V4 contract. The two prior refutations are resolved on this exact HEAD: all A/K/I/F comparison arms are rebuilt under the common `--gc-sections --no-relax` closure; A now rejects any `QCoreApplication::self` relocation; A/K/I/F reject the broader `QApplication::staticMetaObject` path; K proves the same-symbol GOT/dynamic binding with no COPY while preserving the direct DT_NEEDED closure and excluding provider/QCore/metaobject roots.

Exact-head canonical CI run 34216842324 completed successfully. Its `crazyflie-runtime-wwi` job reproduces the C25 controls, then records A=QAPPLICATION_OK/0, K=QAPPLICATION_OK/0, I=Qt/XCB SIGSEGV/139, F=Qt/XCB SIGSEGV/139 and `DIAGNOSTIC_RESULT=C26_COPY_RELOCATION_REQUIRED_RELATIVE_TO_GOT`. This matches pre-registered outcome 1. No product fix or Firefox same-file semantics are inferred.

PR #229, opened 2026-09-08
Split S3 vertical-veto diagnostics without retuning

djibian:
Lab-only #70 evidence preparation after physical checkpoint #180 refuted the scalar S3 classifier.

The retained attachment is not consumable through the available machine path, so this implements the already documented smallest fallback discriminator: preserve the exact existing vertical-veto boolean and every threshold/persistence constant, but split `surfReason` into VZ-only, barometer-only, and both-veto codes. A pre-registered props-off decision table is included.

No Runtime v2 change, no estimator-structure change, no `rangeUp`, no threshold/barometer/ToF retuning, no motor authority, and no checkpoint request while #226 remains open.

Research/Lab candidate only; it does not authorize merge into product behavior or motorized testing.

### DIFF
--- a/experiments/crazyflie-ukf-surface-range/S3_VETO_DISCRIMINATOR.md
+++ b/experiments/crazyflie-ukf-surface-range/S3_VETO_DISCRIMINATOR.md
@@ -1,0 +1,48 @@
+# S3 veto discriminator — props-off, reason split only
+
+## Purpose
+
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO 8229971bc44c5eb2ed19aa293a0b81de0df36479

Independent review finds the Lab-only discriminator preserves the existing S3 veto boolean and frozen thresholds/persistence while splitting only the emitted reason code into VZ/BARO/BOTH. The exact upstream/applicator context is fail-closed and idempotent, the pre-registered decision table matches the durable #70 causal boundary, and no Runtime, estimator-structure, rangeUp, tuning, motor authority, or checkpoint expansion is introduced.

PR #233, opened 2026-09-08
Bind physical preflight to exact AST capabilities

djibian:
Implements the live physical capability-preflight decision recorded on #157 (human product decision comment 5588080013).

- keeps verified exact Crazyflie 2.1 identity mandatory for physical execution;
- continues to observe the connected hardware/deck descriptor without turning every profile-listed optional deck into a global prerequisite;
- gates physical compatibility on the capabilities actually required by the exact submitted AST: action kinds, range directions, move directions and vertical directions;
- keeps fail-closed behavior when a required capability is unavailable, including a required `set_light` capability;
- preserves the read-only capability-inspection boundary and `executionAuthority:false` safeguards;
- adds regression coverage proving that an unused optional deck does not block a compatible AST while a truly required capability still fails closed.

Refs #157

### DIFF
--- a/plugins/robot_windows/blockly/webeeblocks/physical_capability_contract.js
+++ b/plugins/robot_windows/blockly/webeeblocks/physical_capability_contract.js
@@ -114,17 +114,12 @@
     });
   }
 
-  function requireHardware(profile, descriptor) {
-    profile.hardware.forEach(function(requirement) {
-      if (requirement === EXACT_AIRFRAME) {
-        if (descriptor.identity.modelEvidence !== 'verified' ||
-            descriptor.identity.model !== EXACT_AIRFRAME)
-          fail('exact physical model evidence unavailable: ' + EXACT_AIRFRAME);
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO 00182fd7d11497a43b3c0d2def9aa621e9e823a7

Independent bounded review against current main contract and the owner decision on #157. The candidate preserves exact Crazyflie 2.1 fail-closed identity, stops treating profile-listed optional decks as unconditional prerequisites, and still rejects every capability actually required by the exact AST through action/range/move/vertical checks. The live probe remains conservative: movement/action capability is advertised only with a healthy Flow Deck, Multi-ranger directions only with a healthy Multi-ranger, and `set_light` only with a healthy Color LED Deck, while `executionAuthority:false` remains mandatory. Canonical exact-head CI Gate run 34249585621 is successful with its final `CI Gate` job green. No applicable unresolved refutation exists for this HEAD.

PR #234, opened 2026-09-08
Build the S3 VZ/BARO veto discriminator artifact

djibian:
Advances #70 at the already pre-registered X3 causal boundary without changing estimator thresholds or flight behavior.

- applies the existing `apply_surface_offset_s3_veto_discriminator.py` overlay after the base S3 applicator in the deterministic S3 build oracle;
- asserts the VZ-only, BARO-only and BOTH veto reason codes plus the unchanged aggregate veto predicate;
- preserves the pinned Crazyflie 2026.08 source, UKF/Flow build configuration, existing S3 thresholds/persistence and props-off-only research boundary;
- makes CI produce an exact firmware artifact containing the already-authorized observability split, suitable for later independent validation and, only if warranted, a serialized human checkpoint.

No product Runtime, estimator tuning, rangeUp fusion, controller behavior or motorized test authority is changed.

Refs #70

### DIFF
--- a/experiments/crazyflie-ukf-surface-range/run_s3_build_oracle.sh
+++ b/experiments/crazyflie-ukf-surface-range/run_s3_build_oracle.sh
@@ -6,6 +6,7 @@ UPSTREAM="${1:-$ROOT/.ci-crazyflie-firmware}"
 EXPECTED_COMMIT=54f31e243a0b28b67efef5ba20dbb6d9890a5478
 EXPECTED_BLOB=57c0e8405c07b63a29538019895ed17d0a379440
 APPLICATOR="$ROOT/experiments/crazyflie-ukf-surface-range/apply_surface_offset_s3.py"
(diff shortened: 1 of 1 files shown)


djibian (commented):
GO 55858b5442b16244659e7bd1607592c8f99476e1

Independent bounded review against current main contract and the durable #70 X3 causal boundary. The candidate only wires the already-integrated VZ/BARO/BOTH discriminator overlay into the pinned 2026.08 deterministic S3 build oracle and asserts the exact split reason codes plus the unchanged aggregate `verticalVeto = vzVerticalVeto || baroVerticalVeto` predicate. It does not change the pre-registered thresholds, persistence, UKF state, Flow/ToF weighting, Runtime, controller behavior, rangeUp fusion, or motorized-test authority. The discriminator remains fail-closed on exact upstream/S3 context and preserves the original veto semantics while making the cause observable. No applicable unresolved refutation was found for this exact HEAD.

PR #235, opened 2026-09-08
Integrate the bottom Color LED deck visually with the Crazyflie

djibian:
Implements the visual half of the 2026-09-08 human product decision on #157.

- replaces the detached legacy square with the reference bottom Color LED deck envelope (34.7 × 34.7 × 11.6 mm including diffuser);
- makes the deck visibly attached to the Crazyflie through four mounting posts;
- uses the existing Webots `LED` device with a full-height circular diffuser plus a larger translucent emissive halo, so the controlled color has visible side faces rather than only a bottom face;
- adds one bounded non-shadow-casting `PointLight` under that same LED Group, so the existing RGB LED command also creates a nearby color halo without changing Blockly/AST/interpreter/backend semantics;
- keeps the backend-neutral `set_light` semantics and controller device name unchanged;
- adds a deterministic structural contract shared by both Runtime worlds;
- adds an R2025a render probe generated from the actual Runtime v2 world: fixed top, three-quarter and side views are captured with the real `color_led` device OFF and BLUE, central rendered-pixel change must clear a deterministic threshold, and the six frames are assembled into an uploaded contact sheet for perceptual inspection.

The exact candidate now contains the previously missing fixed-angle rendered-evidence path; it still requires exact-head CI plus independent review. Blockly textual LED-state observability remains a separate slice.

Refs #157


djibian (commented):
GO de16e3e41ac425f0ca621a9890103f5e657a19a7

Independent review of the bounded visual implementation found no contract or backend-neutrality defect in this candidate. The existing `color_led` device and `set_light` semantics are preserved; the LED Group is compatible with Webots RGB LED propagation to both PBR emissive surfaces, the bottom-deck envelope is attached to the aircraft rather than left as the legacy detached square, and both Runtime worlds are kept in lockstep by the added oracle. This GO is limited to the exact candidate code. It does not claim the separate human perceptual acceptance (fixed rendered top/three-quarter/side evidence) or the Blockly LED-state observability follow-up described on #157.

djibian (commented):
NO_GO de16e3e41ac425f0ca621a9890103f5e657a19a7

Independent review against current main and the authoritative 2026-09-08 #157 visual decision finds the structural change promising but the bounded visual claim is not yet proven. The candidate removes the legacy detached square, uses the real bottom-deck 34.7 × 34.7 × 11.6 mm envelope, attaches the model with four posts, and preserves the existing generic `color_led` / `set_light` path. Webots' LED semantics will recursively drive the emissive colors of both shapes under the Group.

The applicable human acceptance boundary is perceptual, though: the active color must be clearly distinguishable from ordinary side and three-quarter classroom views, with a diffuse nearby halo, and fixed top/three-quarter/side visual evidence is explicitly required. This HEAD's new oracle proves only WBT source tokens and geometry arithmetic. It cannot falsify the failure mode that motivated the decision: a syntactically valid emissive object that still looks detached, unlike the reference diffuser, or effectively visible only from underneath. The PR body itself records that the required fixed-angle rendered evidence is still absent.

Smallest repair boundary: on this same visual slice, produce deterministic R2025a rendered evidence from fixed top/three-quarter/side cameras with the deck off and an active supported color, while exercising the existing LED device. Make the evidence fail if the deck is not visibly attached or the active color cannot be distinguished from those normal angles. If the current cylinder/transparent-emissive geometry does not yield the required diffuser/nearby-halo effect in those renders, adjust only the Webots visual geometry/lighting (for example a more faithful diffuser surface and/or a bounded non-shadow-casting light under the existing LED Group). Do not change Blockly/AST/interpreter semantics here. The separate textual Blockly LED-state slice remains separate and is not the reason for this NO_GO.

djibian (commented):
NO_GO b0e2a50cca88f3c5e2b883ad28bd9a59837b2e01

The fixed-angle evidence path now exists, but the three camera orientations in `color_led_visual_probe.py` do not actually aim the Webots Viewpoint at the fixed Crazyflie pose. In this Runtime world the camera forward axis is local +X (the existing product Viewpoint rotation maps local +X approximately toward its followed Crazyflie). At the probe pose `[0,0,0.18]`: `top` is above the drone at `[0,0,0.55]` but orientation `[0,0,1,0]` leaves +X horizontal instead of pointing down; `side` is at `[0,-0.45,0.18]` but rotation about X leaves the +X forward axis unchanged instead of pointing +Y toward the drone; and the three-quarter rotation maps local +X approximately to `[0.707,-0.707,0]`, whereas the direction from `[-0.32,-0.32,0.34]` to the drone is approximately `[0.667,0.667,-0.333]`. Thus a render failure/success would currently be dominated by wrong framing, not the intended LED perceptibility property.

Smallest repair boundary: correct only the fixed Viewpoint orientations so local +X points to `[0,0,0.18]`, then rerun the exact R2025a render oracle. Deterministic examples for the current positions are top `[0,1,0,1.5707963267948966]`, side `[0,0,1,1.5707963267948966]`, and three-quarter approximately `[-0.1565580108,0.3779644730,0.9124869568,0.8522498022]`. Keep the product geometry/LED semantics unchanged unless the correctly framed renders themselves falsify visibility.

djibian (commented):
NO_GO d83d6423316342d0a7a3c488812b8adfc6f4ca86

The fixed-camera framing defect from the prior NO_GO is repaired, but the exact Ready candidate still fails its own bounded visual acceptance. Canonical CI run 34260181174 reaches the real R2025a render probe and captures OFF/BLUE top, three-quarter and side frames, then `check_color_led_visual_render.py` aborts on the top-view measurement `5.73916e-05`. Two defects are exposed on the current HEAD:

1. the checker accepts only plain decimal syntax, so a valid ImageMagick floating-point result in scientific notation is rejected before the intended threshold comparison;
2. independently of that parser defect, `5.73916e-05` is far below the declared `MIN_CHANGED_FRACTION = 0.002`, so the current top-view render does not establish the claimed visible LED effect.

Smallest repair boundary: keep the existing fixed R2025a camera evidence path, make the numeric parser accept finite floating-point/scientific notation without weakening the threshold, then adjust only the visual geometry/lighting or the evidence framing if warranted so the correctly rendered normal views genuinely satisfy the predeclared visibility criterion. Do not lower the threshold merely to obtain green CI and do not change Blockly/AST/interpreter semantics. Re-run the exact Ready-head CI afterward.

djibian (commented):
GO 5e54b40ca05e342b664419f90524da4e0b824a2e

Independent review against current `main@1f2d628b14452c9fcbea63447f2a4b02aeefdbc2`, the authoritative #157 visual decision, and all still-applicable prior NO_GO findings finds the bounded visual slice coherent.

The prior framing defect is resolved by fixed top/three-quarter/side Viewpoint orientations aimed at the Crazyflie pose. The prior numeric-parser defect is resolved by finite float parsing that accepts ImageMagick scientific notation. The earlier grayscale oracle was also under-sensitive to pure blue because grayscale conversion attenuated the blue channel before the fixed 10% threshold; the current checker instead takes the maximum RGB channel difference, preserves the same 10% per-pixel threshold and the same `MIN_CHANGED_FRACTION = 0.002`, and therefore tests color visibility without weakening acceptance. On the exact Ready-head CI, the real Webots R2025a OFF/BLUE render step and `Prove Color LED is perceptible from fixed normal views` step both complete successfully.

The product geometry remains bounded to the bottom Color LED representation and lighting: attached deck envelope/posts, LED-driven diffuser/halo and one non-shadow-casting PointLight. The generic `color_led` device name and backend-neutral `set_light` path remain unchanged. The workflow/evidence changes preserve the existing CI obligations rather than changing governance or human boundaries.

This GO is for the exact HEAD only and does not claim the separate Blockly textual LED-state observability slice.

PR #237, opened 2026-09-08
Expose applied Color LED observer state

djibian:
Adds the backend-confirmed Color LED execution-observer transport needed by the 2026-09-08 #157 product decision.

- emits `onLight` only after the backend `setLight` promise has resolved successfully;
- maps the applied light event back to the exact Blockly action through the existing execution observer;
- proves blue → off ordering and proves a rejected backend action never becomes observed LED state;
- keeps the AST, backend-neutral light intent and WWI command semantics unchanged.

This is the transport/core slice only. Student-visible French LED state/reset rendering remains a separate follow-up surface slice. Ready for exact-head full CI and independent review.

Refs #157

### DIFF
--- a/plugins/robot_windows/blockly/webeeblocks/execution_observer.js
+++ b/plugins/robot_windows/blockly/webeeblocks/execution_observer.js
@@ -69,11 +69,13 @@
     this.waiter = null;
     this.credit = 0;
     this.sensorValues = Object.create(null);
+    this.lightColor = 'off';
     var self = this;
     this.hooks = {
       onNode: function(context) { return self._onNode(context); },
       beforeStep: function(context) { return self._beforeStep(context); },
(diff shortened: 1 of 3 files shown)


djibian (commented):
GO 653a29d59bcb70d69545209e9ffcebebf5b871a4

Independent bounded review finds this core observability slice consistent with the current #157 product decision and the backend-neutral Runtime contract. The `set_light` AST/backend command is unchanged; `onLight` is emitted only after `backend.setLight(color)` resolves, so a rejected backend action cannot be presented as applied state. The event remains an execution-observer hook, maps to the exact Blockly action through the existing source map, and its tests cover applied blue→off ordering plus backend failure. This GO is intentionally limited to the transport/core claim: it does not claim the separate student-visible French LED state/reset UI or the separate Webots perceptual deck-render acceptance.

PR #240, opened 2026-09-08
Expose applied Color LED state in execution observability

djibian:
Completes the remaining Blockly execution-observability part of the 2026-09-08 #157 Color LED product decision.

- adds a textual `LED : …` state alongside sensors and variables, with an optional swatch that never carries meaning alone;
- covers all existing backend-neutral Color LED values (`off`, `red`, `green`, `blue`, `yellow`, `white`) with French student labels;
- updates the UI from the existing interpreter `onLight` hook only after a `set_light` backend action has actually succeeded;
- keeps the applied state visible after execution and resets it to `éteinte` only after a successful simulation reset;
- preserves Blockly/AST/interpreter/backend semantics and physical execution authority boundaries.

The existing Runtime core already proves `onLight` ordering after successful backend application and no light observation on failed application; this slice only connects that proven hook to the student observability surface.

Refs #157


djibian (commented):
GO b343513476cd7d0500a788f96b040df76132a8f9

Independent review against current main@9e86d90a6dc9a7a4050ee6198d21085021113351 and the 2026-09-08 #157 Color LED observability decision finds the bounded UI slice coherent.

The new student-facing state is textual first (`LED : éteinte/rouge/verte/bleue/jaune/blanche`) with the swatch explicitly decorative (`aria-hidden`), so color alone never carries meaning. The wrapper preserves the existing execution observer and only renders after awaiting its `onLight` hook; the interpreter invokes that hook only after `backend.setLight(color)` succeeds, so a failed backend application cannot be presented as applied. The wrapper intentionally observes normal execution as well as step mode, while the original observer keeps its own debug semantics unchanged. Reset-to-off is bound to the existing reset event, which is dispatched only after a successful simulation reset.

Exact-head canonical CI run 34271988717 completed successfully with final `CI Gate` success and exercised the real Robot Window/runtime paths without browser errors. No Blockly, AST, interpreter, backend or physical-execution authority semantics are changed.

This GO is limited to exact HEAD b343513476cd7d0500a788f96b040df76132a8f9.

PR #242, opened 2026-09-08
Record the post-#236 S3 redesign boundary

djibian:
Updates the versioned #70 roadmap projection after the authoritative props-off discriminator checkpoint #236.

- records that VZ is not reproducibly dominant and the barometer veto participates across valid terrain repetitions;
- records that S3-B/S3-C true-vertical controls remain below the ToF rejection gate and never enter SUSPECT, so they do not prove the late commit veto can simply be removed;
- narrows the next Lab step to a bounded classifier-evidence/temporal redesign that preserves the proven local-range Flow split and requires a true-vertical control that actually reaches the revised late-decision path;
- explicitly forbids retuning, `rangeUp`, z/f/r expansion, Runtime/controller changes or motorized testing.

Refs #70 and authoritative human result #236.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -117,20 +117,31 @@ A minimal local-surface-range Flow split then kept fresh local range available
 through rejection, but stock barometer authority did not hold world Z and a
 stronger barometer weighting diverged.
 
-The reconstructed scalar `surfaceOffset` S3 classifier has now also been
-physically refuted under its pre-registered props-off checkpoint. S3-A exposed a
-clear ~0.23 m floor/platform range step and formed a coherent ~+0.212 m terrain
-candidate, but the candidate never committed because the detector predominantly
-raised its vertical-motion veto; `surfaceOffset` stayed zero and estimated Z
-remained surface-relative. The S3-B/S3-C true-vertical controls did not create a
-false terrain commit.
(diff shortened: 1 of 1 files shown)


djibian (commented):
GO 7307918df1aa216561e384598cf5ad29f5b4a42d

Independent review of the exact Ready head finds this roadmap-only update faithful to the authoritative #236 PASS and the existing #70 safety boundary. It correctly records that VZ is not reproducibly dominant, the barometer veto participates across the valid terrain repetitions, and the S3-B/S3-C vertical controls stayed below the ToF rejection gate and never entered SUSPECT. The key inference is appropriately bounded: current late VZ/barometer evidence is not a reliable terrain/vertical discriminator, but the data do not justify simply deleting the late veto because the available true-vertical controls never exercise that path.

The proposed next step is therefore correctly narrowed to a pre-registered classifier-evidence/temporal redesign with a true-vertical control that actually reaches the revised late decision path. Existing prohibitions against retuning, rangeUp, z/f/r expansion, Runtime/controller changes and motorized testing are preserved.

Canonical Ready CI run 34278839135 completed successfully on this exact head, with the final CI Gate success and intentionally skipped unaffected Runtime/Webots suites. No applicable NO_GO is present. This GO is limited to exact HEAD 7307918df1aa216561e384598cf5ad29f5b4a42d.

PR #244, opened 2026-09-08
Invalidate physical preflight after reconnect

djibian:
Implements the next bounded non-authority #157 physical-preflight freshness slice from the 2026-09-08 human decision.

- requires connected physical preflight adapters to expose an opaque connection epoch that is stable for one live connection and changes on reconnect;
- brackets capability acquisition with the epoch and fails closed if the connection changes while the descriptor is being read;
- binds a successful exact-AST preflight to that epoch and adds a later connected assertion that rejects a reconnect after preflight;
- preserves the existing fresh descriptor re-read on every connected preflight invocation, exact AST binding, and `executionAuthority:false`;
- introduces no motor, arming, teacher-authorization or physical execution surface;
- records the remaining boundary accurately: a real physical adapter must provide a trustworthy reconnect-sensitive epoch, a future authorization/submission consumer must verify immediately before its effect, and real physical execution remains unproven.

Refs #157.

### DIFF
--- a/docs/REFERENCE_CAPABILITY_MATRIX.md
+++ b/docs/REFERENCE_CAPABILITY_MATRIX.md
@@ -59,14 +59,17 @@ sensors and movement/vertical capabilities from the exact submitted
 backend-neutral AST, keeps optional deck requirements intent-dependent, and
 produces a canonical non-authority AST binding that later code can verify before
 authorization/submission. A connected-preflight entry point now acquires and
-normalizes the live descriptor from the read-only adapter on every invocation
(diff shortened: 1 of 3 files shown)


djibian (commented):
GO 4d06e2bf49cc0028b50932b95da5946256cab8fa

Independent bounded falsification found no blocking defect in the reconnect/session-freshness slice. It preserves executionAuthority:false and the read-only adapter boundary, re-reads live capabilities, fails closed on epoch change during preflight or before a later consumer, and leaves real-adapter epoch trust plus the future immediate pre-effect consumer explicitly unproven.

PR #246, opened 2026-09-09
Add reconnect-aware live capability session

djibian:
Advances the remaining non-authority #157 live-adapter boundary without introducing physical execution.

- adds a persistent host-side `ReadOnlyCapabilitySession` around one explicit Crazyradio/Crazyflie link;
- creates an opaque connection epoch only after a successful live connection and invalidates it from cflib's disconnect callback; reopening rotates the epoch;
- rebuilds each capability descriptor from the current connected evidence instead of reusing an earlier descriptor object, and brackets the read with the same session epoch;
- keeps the existing exact Crazyflie 2.1 device-type query, deck/self-test capability mapping, `executionAuthority:false`, and explicit Crazyradio URI boundary;
- exposes no WebeeBlocks movement, light, arming, setpoint or parameter-write method; cflib's documented safety-zero close-path qualification remains explicit;
- adds deterministic fake-link coverage for descriptor refresh, disconnect fail-closed behavior, reconnect epoch rotation, cleanup on invalid epoch, and the absence of authority methods;
- updates the capability matrix narrowly: the host live session exists, but Blockly/runtime submission wiring, immediate pre-effect AST/session verification, arming/abort/failsafe and real student execution remain unproven.

This PR does not authorize flight, does not wire a physical execution backend, and does not change #70 or Runtime v2 semantics.

Refs #157.

### DIFF
--- a/docs/REFERENCE_CAPABILITY_MATRIX.md
+++ b/docs/REFERENCE_CAPABILITY_MATRIX.md
@@ -54,20 +54,25 @@ real `color_led` device and gate that normal-view color visibility without
 modeling deck electronics or photometric fidelity.
 
 The substantive remaining #157 boundary is physical-backend continuity/proof.
-The integrated physical preflight core now derives required actions, range
-sensors and movement/vertical capabilities from the exact submitted
-backend-neutral AST, keeps optional deck requirements intent-dependent, and
-produces a canonical non-authority AST binding that later code can verify before
-authorization/submission. A connected-preflight entry point now acquires and
-normalizes the live descriptor from the read-only adapter on every invocation,
-brackets that acquisition with an opaque adapter-provided connection epoch, and
-binds the successful result to that epoch. The connected assertion fails closed
-if the epoch changes before a later consumer checks the preflight, and a
-connection change during descriptor acquisition invalidates the attempt. This
-still is not execution authority: a real physical adapter must provide a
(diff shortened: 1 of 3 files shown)


djibian (commented):
GO 056fb5f4c698960f500fc11b4e300d427bba0453

Independent review finds the bounded non-authority claim coherent. `ReadOnlyCapabilitySession` keeps one explicit Crazyradio link, creates an opaque epoch only after successful connection/parameter readiness, invalidates it from the live disconnect callback, rotates it on reopen, and rebuilds descriptors from current connected evidence rather than caching a prior descriptor. Capability reads are bracketed by the session epoch and fail closed after disconnect. The public adapter surface adds no movement, light, arming, setpoint, supervisor, or parameter-write operation; `executionAuthority:false` and the existing cflib safety-zero close-path qualification remain intact.

The fake-link oracle covers refresh, disconnect invalidation, reconnect rotation, cleanup on invalid epoch, and absence of authority methods. The capability-matrix update correctly leaves Blockly/runtime wiring, immediate pre-effect AST/session consumption, arming/abort/failsafe, and real student execution unproven. Exact pinned cflib `Caller.call()` copies the callback list before invocation, so SyncCrazyflie removing its own disconnect callbacks does not skip the session invalidation callback.

No #70, Runtime v2, governance, or flight-authority boundary is changed.

PR #249, opened 2026-09-09
Connect physical preflight to live read-only session

djibian:
Advances #157 with the smallest non-authority bridge between the backend-neutral physical preflight contract and the integrated live Crazyradio capability session.

- adds a loopback-only authenticated HTTP surface over `ReadOnlyCapabilitySession` exposing only fresh capabilities and the reconnect-sensitive epoch;
- adds a browser-side adapter implementing only `readCapabilities()` / `readConnectionEpoch()` so the existing `preflightConnected(...)` contract can consume the live session without gaining flight authority;
- binds that bridge into the production Blockly v2 runtime so the exact current activity profile/workspace AST can be preflighted and later re-asserted against the same live connection epoch;
- preserves an internal immutable AST/session ticket even if caller-owned preflight data is mutated;
- keeps `executionAuthority:false` throughout and exposes no arming, setpoint, motor, movement or light effect API;
- wires the HTTP/session/product binding tests into the existing Runtime v2 core harness, including reconnect, workspace-mutation, unauthorized-read and no-authority fail-closed checks.

This deliberately stops before the separately authorized physical effect consumer: arming/abort/failsafe and actual real-flight execution remain unproven and out of this slice.

Refs #157.

PR #250, opened 2026-09-09
Update physical preflight roadmap after #249

djibian:
Updates the smallest roadmap fragments whose prerequisite changed when #249 integrated.

- records #249 as the production submission/live-session binding and immediate pre-effect re-assertion boundary;
- removes the now-superseded wording that still listed that wiring as remaining work;
- leaves the remaining physical boundary explicit: separately authorized effect consumption, arming/abort/failsafe, and real execution continuity remain unproven;
- preserves `executionAuthority:false` and does not authorize flight or physical testing.

Refs #157.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -24,18 +24,21 @@ justified student-facing generic capability. Integrated #193/#196 establish the
 machine-readable read-only capability evidence path and exact Crazyflie 2.1
 identity; #238/#241/#243/#244 establish intent-dependent exact-AST preflight,
 canonical AST binding, fresh connected-descriptor acquisition on each preflight
-and reconnect invalidation through an opaque connection epoch; #246 now provides
-the concrete host-side non-authority live Crazyradio session that owns that epoch,
+and reconnect invalidation through an opaque connection epoch; #246 provides the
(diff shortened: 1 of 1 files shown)


djibian (commented):
GO 819e650fdbaaa2c8f1451855637920aba878ea79

Independent bounded review against current `main@2769b3ede3b56d6127ab781dd57f0710134602d0`, the integrated #249 source, and the authoritative #157 live-preflight decision confirms this roadmap-only candidate accurately updates the durable dependency boundary.

#249 is integrated on current main and the production physical-preflight runtime binds the current activity profile, semantic workspace AST, and reconnect-sensitive live connection epoch, then rechecks profile/AST after asynchronous re-assertion before returning success. The submission bridge remains `executionAuthority:false`, and the host session exposes capability/epoch reads without introducing WebeeBlocks arming, motor, movement, setpoint, or light-effect authority. The proposed roadmap text therefore correctly removes the superseded “wire submission to #246 / immediate pre-effect verification” dependency and moves the remaining boundary to separately authorized effect consumption, explicit arming/abort/failsafe semantics, and proof of real execution continuity.

The change does not broaden product authority, authorize flight or physical testing, or alter #70. Exact-candidate CI Gate is successful on this Ready HEAD.

Issue #72, opened 2026-08-23
[Product] Teacher-authorized final real-flight activity


djibian:
## Validated product objective

Real Crazyflie flight is the **finality of the module**, not a routine student action throughout the sequence.

Canonical product vision: `docs/PRODUCT_VISION.md`.

## Goal
Define, only after the physical backend is sufficiently demonstrated, a final activity where the student:
1. builds and validates the program in Webots simulation;
2. saves the final project/program;
3. presents it for the final physical run;
4. the teacher explicitly authorizes real execution;
5. the same backend-neutral student program executes on the physical Crazyflie under independent preflight/safety checks.

Assume **one physical Crazyflie**. Classroom sequencing is handled by the teacher, not by a software request/queue system.

## Hard constraints
- no student-facing “request a flight” workflow;
- no software queue for the drone;
- no routine physical flight in earlier activities unless a future explicit pedagogical decision changes the module design;
- teacher authorization is mandatory for the real run;
- no step-by-step debug on real hardware;
- no live pedagogical modification during flight;
- physical preflight and abort/failsafe behavior are infrastructure/safety concerns, independent of the student's algorithm;
- preserve `activity/profile -> Blockly -> backend-neutral AST -> preflight -> shared interpreter -> physical backend`.

## Dependencies / gates
Do **not** design the detailed final 3D course yet.

This issue remains queued behind demonstrated physical capabilities, notably:
- real read-only Crazyflie/Crazyradio/deck capability validation;
- safe physical execution architecture and explicit future arming/failsafe authorization;
- Lab #70 proving a credible solution for stable world altitude over `floor -> table -> floor` if the final activity requires furniture/terrain discontinuities.

## Acceptance direction
When the physical backend is mature enough, a bounded acceptance should prove:
- the exact saved program validated in simulation is the one authorized for physical execution;
- capability/preflight checks reject incompatible/unsafe programs before any motor action;
- teacher authorization is an explicit final gate;
- no debug/step controls are available in physical mode;
- abort/failsafe behavior does not depend on student program correctness;
- one representative final mission can run reliably on the real Crazyflie.

## Non-goals
- no multi-drone management;
- no booking/queue system;
- no student accounts;
- no physical leaderboard;
- no detailed final-course/beacon design before physical capability gates converge.

## Scheduling
Queued. Do not displace current Runtime/UI work or #70. Lead should activate this only when the physical backend evidence is strong enough to make the final mission design non-speculative.

PR #252, opened 2026-09-09
Clarify P1 teacher authorization on stock auto-arming Crazyflie

djibian:
Updates the roadmap to preserve the exact P1 source conclusion already recorded on #157.

- records that stock brushed Crazyflie 2.1 auto-arms when pre-flight checks pass, so the teacher gate cannot be modeled as merely approving a host arming packet;
- requires explicit teacher authorization before any flight-capable command/effect;
- keeps the firmware emergency-stop watchdog as an independent liveness/failsafe guard;
- distinguishes exceptional immediate motor-cut paths from normal controlled land/disarm completion or voluntary abort;
- adds no physical command path, execution authority, motorized test or student semantic.

Docs-only alignment with the authoritative #157 P1 findings. Refs #157 and #72.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -35,8 +35,13 @@ separately authorized effect. All of this remains non-authority with
 emits its documented safety-zero commander setpoint, so this is not a claim that
 the transport emits no command packet. The substantive remaining product boundary
 now starts after that validated pre-effect assertion: a separately authorized
-physical effect consumer plus explicit arming/abort/failsafe semantics and proof
-of safe real execution continuity. Do not invent additional simulation vocabulary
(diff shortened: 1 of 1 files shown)


djibian (commented):
GO 77a46588f772fd70d9498d935daae62f32d9ac5f

Independent review: the docs-only roadmap change matches the authoritative #157 source findings, including the stock CF2.1 auto-arming correction, the stronger teacher-authorization invariant, the independent emergency-stop watchdog/liveness boundary, and the distinction between exceptional motor-cut paths and controlled normal land/disarm. No execution authority, command path, motorized test, student semantic, governance obligation, or safety oracle is weakened.

PR #253, opened 2026-09-09
Refresh physical capability matrix after live preflight integration

djibian:
Aligns `docs/REFERENCE_CAPABILITY_MATRIX.md` with integrated #249/#252 and the authoritative #157 P1 source findings.

- records that the production physical submission path now consumes the live Crazyradio capability session and re-asserts the exact profile/AST/connection binding before any later separately authorized effect;
- preserves the non-authority boundary;
- records the remaining teacher-authorization, auto-arming-independent, watchdog/liveness, controlled land/disarm and execution-continuity boundary;
- records the source-selected direct `HighLevelCommander` semantic direction without claiming physical proof or authorizing flight.

Docs-only stale-inventory correction. Refs #157 and #72.

### DIFF
--- a/docs/REFERENCE_CAPABILITY_MATRIX.md
+++ b/docs/REFERENCE_CAPABILITY_MATRIX.md
@@ -68,13 +68,31 @@ a disconnect makes both epoch and capability reads fail closed. The session
 exposes no WebeeBlocks movement, light, arming or setpoint method and retains the
 documented cflib safety-zero close-path qualification.
 
-This still is not execution authority or a complete physical backend. The
-Blockly/runtime submission path does not yet consume this host session, and no
-authorization/execution consumer performs the bound AST/session assertion
-immediately before a physical effect. Arming, abort/failsafe behavior and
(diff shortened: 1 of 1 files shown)


djibian (commented):
GO 405b74023ddf93403da72ed73e96418b4a844d4a

Independent bounded review against current `main@32e870cd8a115ec1b12afff3ace7d89d64bc085f`, integrated #249/#252/#254, and the authoritative #157 P1 findings confirms this inventory-only update removes a real stale claim without expanding authority.

The new matrix text accurately records the live-session/profile/AST/connection-epoch preflight and re-assertion already integrated by #249, keeps that bridge non-authority, and places the remaining boundary after preflight: explicit teacher authorization before flight-capable effects, stock-CF2.1-auto-arming-independent gating, independent watchdog/liveness, controlled land/disarm, and real execution continuity. Its direct `HighLevelCommander` / body-to-world / supervisor-observed completion direction matches the source finding now projected in current ROADMAP by #254.

No physical execution support, motorized-test claim, student vocabulary, or safety oracle is added or weakened.

PR #254, opened 2026-09-09
Record direct HighLevelCommander physical semantics

djibian:
Updates the smallest roadmap proof-direction fragment after the authoritative #157 P1 command-path finding.

- selects direct cflib `HighLevelCommander` as the primitive physical effect substrate instead of `MotionCommander` / `PositionHlCommander` helpers;
- preserves existing WebeeBlocks body-relative horizontal semantics by requiring a body-to-world transform from accepted yaw before relative high-level moves;
- keeps vertical/yaw intent in the existing world/relative semantics;
- requires normal command completion to be bounded by observed supervisor/high-level trajectory state with timeout/locked/crashed checks;
- keeps emergency stop / high-level stop exceptional and normal completion/abort on a controlled high-level land then disarm path;
- adds no physical command implementation, execution authority, motorized test, or student-facing semantic.

Roadmap-only alignment with owner-authored #157 evidence. Refs #157, #70 and #72.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -297,12 +297,17 @@ Runtime/controller change is justified by #236 alone.
   land/disarm behavior and proof of physical execution continuity. Do not assume
   a host arming packet is the teacher gate: stock brushed Crazyflie 2.1
   auto-arms when pre-flight checks pass
-- proof direction: preserve backend-neutral AST continuity while adding only the
-  minimum teacher-authorized physical execution authority after the live
-  capability and safety gates are independently established. Keep immediate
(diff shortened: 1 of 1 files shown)


djibian (commented):
GO 801f1532f4a76fa5f18e8d40e5f26ed9f53dff83

Independent review: this roadmap-only change faithfully projects the authoritative #157 HighLevelCommander source finding. It preserves backend-neutral/body-relative student semantics, requires the explicit body-to-world transform and supervisor-bounded completion, keeps teacher authorization and watchdog/controlled-land safety gates intact, and adds no execution authority, command implementation, motorized test, or new student vocabulary.

PR #257, opened 2026-09-09
Add fail-closed fresh supervisor state reader

djibian:
Adds the bounded non-authority supervisor observation prerequisite identified on #157.

- reads one fresh CRTP supervisor state reply directly instead of using cflib `Supervisor`'s cached-timeout fallback;
- fails closed on timeout, disconnect, malformed matching response, or unsupported protocol;
- exposes decoded supervisor/high-level trajectory state only, with no arming, watchdog, setpoint, high-level commander, or other effect API;
- proves that a second timed-out read never reuses an earlier `hl_traj_finished` value;
- wires the test into the Runtime v2 core harness.

This does not add execution authority, does not authorize flight, and does not replace the independent emergency-stop watchdog or teacher-authorization boundaries. Refs #157 and #72.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -49,6 +49,10 @@ def main():
     if physical_probe_test.returncode:
         print('FAIL read-only Crazyradio capability evidence',file=sys.stderr);print(physical_probe_test.stdout,file=sys.stderr);print(physical_probe_test.stderr,file=sys.stderr);return physical_probe_test.returncode
     print(physical_probe_test.stdout.strip())
+    physical_supervisor_test=subprocess.run([sys.executable,'tools/ci/test_physical_supervisor_state.py'],text=True,capture_output=True)
+    if physical_supervisor_test.returncode:
+        print('FAIL connection-epoch physical supervisor freshness',file=sys.stderr);print(physical_supervisor_test.stdout,file=sys.stderr);print(physical_supervisor_test.stderr,file=sys.stderr);return physical_supervisor_test.returncode
+    print(physical_supervisor_test.stdout.strip())
     physical_http_test=subprocess.run([sys.executable,'tools/ci/test_physical_capability_http_bridge.py'],text=True,capture_output=True)
     if physical_http_test.returncode:
         print('FAIL read-only physical capability HTTP bridge',file=sys.stderr);print(physical_http_test.stdout,file=sys.stderr);print(physical_http_test.stderr,file=sys.stderr);return physical_http_test.returncode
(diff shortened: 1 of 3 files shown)


djibian:
Draft review against the newer #157 findings: two safety-oracle gaps remain on `5ee14e362e5bca65740c79f59b26b783e2a429d5`.

1. The raw firmware supervisor bitfield includes **bit 11 = deck fault**. The current `SupervisorState` stops at bit 10, so the future execution gate could ignore a freshly observed runtime hardware-fault signal. Preserve/expose the raw bitfield and decode/evaluate deck fault fail-closed.

2. A timeout leaves an untagged GET_STATE reply potentially outstanding. Because this CRTP command has no transaction ID, a later invocation on the same connection can accept that delayed old reply as its own “fresh” result. The current stateless function plus serialization does not exclude that race. Timeout/malformed transport must poison supervisor freshness for that connection until reconnect/session replacement rotates the connection epoch and preflight is re-established; add a deterministic late-reply regression.

Also, for the pinned protocol the GET_STATE payload is exactly command + uint16. Accepting arbitrary extra response bytes as part of the bitfield is not obviously conservative; exact-length validation is the safer contract unless a pinned protocol extension is separately proven.

These are non-authority corrections only; they do not require any motorized test.

djibian:
NO_GO ba51d15dab3736ad4d1a6b838125699e21a89e7c

Smallest remaining repair boundary: supervisor reads are serialized only by each `FreshSupervisorStateReader` instance (`self._read_lock`). Two reader instances can be constructed on the same connection epoch before either is poisoned and can issue the same untagged `CMD_GET_STATE_BITFIELD` concurrently. Because the protocol has no transaction ID, one matching reply can then be observed by both callbacks and treated as fresh for both reads, defeating the exact ambiguity this slice is intended to remove.

Make serialization/ambiguity state connection-epoch-wide, not instance-local: all readers bound to the same epoch must share one read lock (or duplicate readers must otherwise be rejected). After acquiring that shared guard, re-check epoch poison before emitting the request so a waiter cannot send after a concurrent timeout poisons the epoch. Add a deterministic two-reader regression proving no second request is emitted concurrently and that a waiting reader fails closed without sending if the first read poisons the epoch.

The earlier deck-fault, exact-frame and timeout-poison findings appear addressed on this HEAD. No motorized evidence is required.

djibian (commented):
GO 02d69e9f4618231594ec158564a175eb003a49f4

Independent review against current main and all still-applicable #157 findings: the reader now binds freshness to the reconnect-sensitive connection epoch, uses an epoch-wide shared read lock across reader instances, re-checks poison/epoch after acquiring that guard, and permanently fails closed for the epoch after ambiguous timeout, disconnect, malformed framing, transport failure or unknown supervisor state semantics. The deterministic two-reader regression proves a waiting reader emits no second untagged GET_STATE request and fails closed after the first reader poisons the epoch. The raw uint16 state preserves and blocks on deck fault, exact response framing is enforced, and no arming/watchdog/setpoint/HighLevelCommander authority is exposed. The earlier NO_GO on ba51d15 is therefore resolved on this exact Ready HEAD.

PR #256, opened 2026-09-09
Add pure physical HighLevelCommander semantic adapter

djibian:
Implements the smallest non-authority #157 command-semantics slice selected by the authoritative P1 source finding.

- adds a pure host-side semantic adapter for the existing Runtime v2 `move`, `vertical` and `turn` meanings;
- rotates body-relative horizontal movement into world-frame deltas from an explicitly supplied accepted yaw, exactly matching the current Webots controller geometry;
- maps vertical intent to relative world-Z and signed turns to relative yaw;
- fail-closes outside the already-integrated Runtime v2 bounds and vocabulary;
- adds deterministic parity tests to the existing Runtime v2 core harness.

The new module deliberately imports no cflib API and emits no command. It does not create execution authority, teacher authorization, a duration policy, watchdog/liveness, supervisor completion, arming, landing, or motorized-test capability. Those remain separate safety/effect boundaries. Refs #157 and #72.

PR #258, opened 2026-09-09
Refresh physical roadmap after HighLevel semantics integration

djibian:
Updates durable product documentation after integrated #256 and the verified emergency-watchdog lifecycle finding on #157.

- records #256 as the established pure non-authority body/world, world-Z and relative-yaw semantic adapter instead of leaving that work as an unimplemented proof direction;
- preserves the separate physical execution-authority boundary;
- records that the pinned emergency-stop watchdog has no disable/reset command after first activation, so trusted keepalives must span the reusable physical session and stopping them implies the latching emergency-stop/reboot path;
- keeps fresh supervisor completion, teacher authorization, controlled landing/disarm and real execution proof as remaining gates.

Docs only. Refs #157 and #72.

### DIFF
--- a/docs/REFERENCE_CAPABILITY_MATRIX.md
+++ b/docs/REFERENCE_CAPABILITY_MATRIX.md
@@ -82,8 +82,9 @@ assertion: a separately authorized physical effect consumer with explicit teache
 authorization before any flight-capable command/effect, firmware-independent
 arming semantics (stock brushed Crazyflie 2.1 auto-arms when pre-flight checks
 pass), an independent emergency-stop watchdog/liveness guard, controlled normal
-land/disarm behavior and proof of real execution continuity. Integrated #256 now codifies the non-authority semantic transform for the
-future direct cflib HighLevelCommander path: body-relative horizontal movement is
+land/disarm behavior and proof of real execution continuity. Integrated #256
+now codifies the non-authority semantic transform for the future direct cflib
+HighLevelCommander path: body-relative horizontal movement is
 rotated into world-frame displacement from an accepted yaw, vertical intent stays
 relative world-Z and turns stay relative yaw. That adapter imports no cflib API
 and emits no command, so physical execution authority remains unproven. A later
--- a/docs/REFERENCE_CAPABILITY_MATRIX.md
+++ b/docs/REFERENCE_CAPABILITY_MATRIX.md
@@ -95,7 +95,12 @@ reconnect. Neither slice emits flight authority.
 
 A later effect consumer still must supply the accepted live yaw, consume #257 to
 bound command completion, and establish the independent watchdog before any
-flight-capable effect. The pinned watchdog is one-way after first activation:
+flight-capable effect. For safety decisions, #257 must be the exclusive
+authoritative issuer of supervisor GET_STATE requests on the connection epoch:
+the passive cflib supervisor callback may remain registered, but later
+authority/completion/watchdog code must not invoke cflib `Supervisor` getters in
+parallel or use them as a second oracle because they bypass #257's epoch-wide
+serialization and ambiguity poisoning. The pinned watchdog is one-way after first activation:
 there is no normal disable/reset command and abandoning keepalives eventually
 enters the latching locked/reboot path. Its keepalive also has no application
 reply. The current bounded activation-proof direction is therefore a same-epoch,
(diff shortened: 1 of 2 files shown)


djibian:
NO_GO 816aca2448ccdce0bc56f4c2aecd68ef3f32d1b7

This documentation candidate is already stale against current `main@7a15102b27b135855b53fe531a894269dc053fd5` and the latest durable #157 evidence.

1. **#257 is integrated on current main.** Fresh supervisor observation is no longer merely a future proof direction. The established non-authority reader is connection-epoch bound, epoch-wide serialized, exact-framed, preserves the raw supervisor bitfield including deck fault, and poisons ambiguous freshness/transport until reconnect. The remaining boundary is later authority consuming this observer, not implementing it.

2. The watchdog lifecycle paragraph is correct but incomplete after the newer #157 causal-fence finding. On one unchanged epoch, enqueueing the watchdog keepalive and then successfully completing the fresh GET_STATE read on the same ordered supervisor port is the current bounded activation proof direction. This does not change the one-way-until-reboot lifecycle.

Please project both facts in the smallest affected C1b/P fragments (and the capability matrix if it continues to summarize the remaining boundary) before this docs-only candidate becomes authoritative. No motorized evidence or execution authority is implied.

djibian (commented):
GO 71f57b2083f5fd7e2fd7b6e1e7765a66cac6e384

Independent bounded review against current `main@7a15102b27b135855b53fe531a894269dc053fd5`, integrated #256/#257, and the authoritative #157 watchdog/supervisor findings resolves the earlier NO_GO on `816aca2448ccdce0bc56f4c2aecd68ef3f32d1b7`.

The candidate now projects the full established non-authority boundary rather than stale future work: #257 is recorded as the connection-epoch-bound, epoch-wide-serialized, exact-framed fresh raw supervisor observer including deck fault and ambiguity poisoning; the capability matrix carries the same established fact. It also preserves both watchdog conclusions together: first activation has no direct application reply but can be causally fenced by a successful same-epoch supervisor-port fresh GET_STATE after the keepalive under the documented strict in-port ordering guarantee, while the stock watchdog remains one-way for the powered session and cannot simply be abandoned after normal landing without the latching lock/reboot consequence.

The remaining teacher authorization, effect authority, accepted live yaw, trusted watchdog orchestration, controlled land/disarm and real execution-continuity claims stay explicitly unproven. No simulation vocabulary or motorized support claim is widened.

djibian (commented):
NO_GO 71f57b2083f5fd7e2fd7b6e1e7765a66cac6e384

The previous #257/watchdog corrections are now projected, but one newer #157 safety prerequisite is still absent: the integrated #257 observer must be the exclusive issuer of authoritative supervisor GET_STATE requests on the connection epoch. Pinned cflib always constructs `cf.supervisor`; its getters can independently issue the same untagged GET_STATE and therefore bypass #257's epoch-wide lock. The passive registered callback is fine, but the later authority/completion/watchdog layer must consume #257 results and must not invoke cflib Supervisor getters in parallel or as a second safety oracle.

A compact clause in the P proof direction / capability-matrix remaining boundary is sufficient. This is a docs-only dependency correction, not new execution authority or a motorized-test requirement.

djibian (commented):
GO df138f9fc871d3a489979bbfc26ab3fc8392da64

Independent bounded review against current `main@7a15102b27b135855b53fe531a894269dc053fd5` and all still-applicable #157 refutations confirms the docs-only candidate now closes the remaining projection gap from `71f57b2083f5fd7e2fd7b6e1e7765a66cac6e384`.

The roadmap and capability matrix explicitly require #257 to be the exclusive decision-authoritative supervisor GET_STATE issuer on a connection epoch; pinned cflib `cf.supervisor` getters may not run in parallel or act as a second safety oracle because they bypass #257's epoch-wide serialization/ambiguity guard. The passive cflib callback is not incorrectly treated as authority. The candidate also retains the integrated #256 transform, #257 fresh raw-state/deck-fault observer, same-port watchdog activation fence, one-way watchdog lifecycle, and the unproven teacher/effect/landing/execution boundaries without widening any physical-support claim.

The prior exact-head NO_GO findings are resolved on this Ready HEAD.

djibian (commented):
NO_GO df138f9fc871d3a489979bbfc26ab3fc8392da64

A newer authoritative #157 source finding now contradicts the remaining “controlled land/disarm” wording on this docs candidate. On stock brushed CF2.1 with AUTO_ARMING=1, completed high-level landing does not establish a persistent disarmed safety state: after the landed/reset path clears the arming request, healthy pre-flight checks can auto-arm the vehicle again and return it to ReadyToFly without a new host arm request. A host disarm request is therefore not a durable post-mission lockout either.

Please replace the normal-completion claim with the actual boundary: controlled high-level landing trajectory; fresh #257 evidence that physical flight has ended with no blocking fault; continued trusted watchdog/session handling; and teacher authorization continuing to gate every later flight-capable effect even if firmware auto-arms again. Keep HL stop / immediate e-stop exceptional motor-cut paths.

This is a docs-only correction from #157 comment 5600744407 and requires no motorized evidence.

PR #259, opened 2026-09-09
Refine normal landing semantics for stock auto-arming

djibian:
Projects the verified #157 normal-completion finding into the durable physical roadmap and capability matrix.

- replaces the misleading idea of persistent post-landing disarm on stock CF2.1;
- records that a controlled high-level landing must complete without using the high-level `stop()` motor-cut path;
- requires fresh #257 supervisor evidence that flight ended without a blocking fault while the trusted watchdog/session remains live;
- preserves explicit teacher authorization for every later flight-capable effect because stock auto-arming may return the vehicle to `ReadyToFly` after the landing/reset cycle.

Docs only; no physical command, execution authority or motorized evidence. Refs #157 and #72.

### DIFF
--- a/docs/REFERENCE_CAPABILITY_MATRIX.md
+++ b/docs/REFERENCE_CAPABILITY_MATRIX.md
@@ -82,7 +82,7 @@ assertion: a separately authorized physical effect consumer with explicit teache
 authorization before any flight-capable command/effect, firmware-independent
 arming semantics (stock brushed Crazyflie 2.1 auto-arms when pre-flight checks
 pass), an independent emergency-stop watchdog/liveness guard, controlled normal
-land/disarm behavior and proof of real execution continuity. Integrated #256
+landing/completion behavior and proof of real execution continuity. Integrated #256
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO 399f60d6159755535ef8a724ca2e3b1d51fd07e8

Independent bounded review against current `main@ad80c4425fab5d0671ace960680af9d5eb1aa705` and #157 comment 5600744407: this docs-only repair removes the incorrect persistent post-landing disarm assumption from both the roadmap and capability matrix. Normal completion is now the controlled high-level landing trajectory plus fresh #257 evidence that flight ended without a blocking fault while the trusted watchdog/session remains live; stock auto-arming may return the vehicle to ReadyToFly, so every later flight-capable effect remains behind explicit teacher authorization. Immediate e-stop and HighLevelCommander stop remain exceptional motor-cut paths. The repair adds no execution authority or motorized support claim and does not disturb the #256/#257/watchdog/exclusive-GET_STATE boundaries already integrated.

PR #260, opened 2026-09-09
Add fresh same-epoch physical yaw observer

djibian:
Adds the non-authority accepted-yaw prerequisite identified on #157 for later consumption by integrated #256.

- logs the firmware `stateEstimate.yaw` source instead of deriving yaw from synthetic planner state or the student program;
- binds observation to the live reconnect-sensitive connection epoch;
- uses the first newly received sample only as a timestamp baseline, then requires a later post-call firmware sample before returning an accepted yaw;
- handles the 24-bit firmware log timestamp wrap, rejects duplicate/stale timestamps and non-finite/malformed samples, and fails closed on timeout, log error, disconnect or epoch rotation;
- converts degrees to radians exactly once in the returned observation;
- keeps a continuous session log stream and exposes no commander, arming, watchdog, setpoint or other execution authority;
- wires deterministic regressions into the Runtime v2 core harness.

No physical command, motorized evidence or flight authority. Refs #157 and #72.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -53,6 +53,10 @@ def main():
     if physical_supervisor_test.returncode:
         print('FAIL connection-epoch physical supervisor freshness',file=sys.stderr);print(physical_supervisor_test.stdout,file=sys.stderr);print(physical_supervisor_test.stderr,file=sys.stderr);return physical_supervisor_test.returncode
     print(physical_supervisor_test.stdout.strip())
+    physical_yaw_test=subprocess.run([sys.executable,'tools/ci/test_physical_yaw_observer.py'],text=True,capture_output=True)
+    if physical_yaw_test.returncode:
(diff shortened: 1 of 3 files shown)


djibian (commented):
GO ea174486d58588e20e727fc08a8a3b9316c408cd

Independent bounded review against current `main@870bb5ba60b57dea84694689f15ce8c856197d77` and authoritative #157 yaw evidence 5600747704.

The candidate uses the firmware `stateEstimate.yaw` log source, binds the stream to the reconnect-sensitive live epoch, establishes the first received sample only as a timestamp baseline, and requires a strictly later post-call firmware sample before returning. Duplicate/stale timestamps are rejected, the 24-bit log timestamp wrap is handled, malformed/non-finite values and log/disconnect/epoch failures fail closed, and degrees are converted to radians exactly once for later #256 consumption.

The continuous observer remains non-authority: it exposes no commander, arming, watchdog or setpoint operation and does not infer yaw from synthetic planner state or student semantics. Timeout does not fabricate freshness, and deterministic regressions cover cached baseline rejection, duplicate timestamps, wrap, cleanup and failure paths. No motorized evidence is claimed or required.

PR #261, opened 2026-09-09
Record integrated fresh yaw observer

djibian:
Projects integrated #260 and the already-authoritative #157 teacher-authorization trust boundary into the durable physical roadmap/capability matrix.

- records the non-authority `stateEstimate.yaw` observer as established rather than future work;
- keeps the future #256 consumer responsible for taking one fresh same-epoch yaw sample immediately before the horizontal transform while preserving exact preflight/session binding;
- records that the existing browser-held capability/preflight bearer must remain non-authority and cannot become the teacher credential or gain effect methods merely for reuse;
- requires future teacher authorization to live on a distinct trusted host-side control path unavailable to the student/browser runtime;
- leaves #257 supervisor freshness, watchdog activation/lifecycle and physical effect authority as separate remaining gates.

Docs only; no command, execution authority or motorized evidence. Refs #157 and #72.

### DIFF
--- a/docs/REFERENCE_CAPABILITY_MATRIX.md
+++ b/docs/REFERENCE_CAPABILITY_MATRIX.md
@@ -75,11 +75,17 @@ preflighted against fresh live capability evidence, then both semantic
 requirements and the reconnect-sensitive connection epoch are re-asserted
 immediately before any later separately authorized physical effect. Workspace,
 profile or connection changes invalidate the prior binding. The bridge remains
-non-authority and exposes no WebeeBlocks flight command API.
+non-authority and exposes no WebeeBlocks flight command API. Its browser-held
+capability/preflight bearer is therefore not a teacher-authorization credential:
+future physical authority must stay on a distinct trusted host-side control path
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO 006d48180e2ef5894c7ecd896346855c2225d05b

Independent bounded review against current `main@41a30ff7218446a4afa6b9aa12c7ec4b0affc52a`, integrated #260 and authoritative #157 teacher-authorization trust evidence.

The candidate accurately projects #260 as a non-authority, same-epoch `stateEstimate.yaw` observer whose accepted sample is taken immediately before the #256 horizontal transform while preserving the exact preflight/session binding. It also preserves the distinct teacher trust domain: the existing browser-held capability/preflight bearer remains read-only/non-authority, future teacher authorization must live on a trusted host-side path unavailable to the student/browser runtime, and effect methods are not smuggled under the existing bearer merely for reuse.

The remaining #257 exclusive supervisor observation, watchdog activation/lifecycle, controlled completion and physical effect authority boundaries remain separate and unproven. No motorized support or execution authority is widened.

PR #262, opened 2026-09-09
Add fail-closed physical watchdog liveness guard

djibian:
Adds the non-student emergency-stop watchdog safety substrate for the remaining #157 physical boundary.

Candidate scope:
- causally fence the first stock supervisor watchdog keepalive with a later fresh #257 GET_STATE observation on the same unchanged connection epoch;
- bind that authoritative #257 reader to the exact Crazyflie object as well as the epoch;
- maintain keepalives on a dedicated trusted-host liveness thread with host-gap bounds below the 1 s firmware timeout;
- consume an externally established trusted powered-session watchdog lifecycle rather than minting fresh powered-session state from an arbitrary token inside this primitive;
- preserve terminal/lost watchdog certainty across guard and connection replacement through that external lifecycle, so reconnect/new epoch alone is never reset proof;
- allow a fresh external lifecycle only as a responsibility of the future trusted reset/session layer after separately proven STM+deck reset/power-cycle plus reconnect/re-preflight; this PR implements neither reset nor fresh-session minting;
- expose no arming, flight, HighLevelCommander, setpoint, landing, reset/power-cycle or browser/student API.

Deterministic regressions cover same-port activation ordering, exact Crazyflie/epoch binding, ambiguous fence failure, first/periodic enqueue deadline loss, periodic enqueue failure, host deadline loss, same powered lifecycle across reconnect, rejection of reconstructed non-fresh/unknown external lifecycle state, and simulated consumption of a fresh lifecycle supplied only by the test stand-in for the future reset layer.

This remains a bounded safety primitive. It does not provide teacher authorization, physical command authority, reset authority or a motorized checkpoint. Refs #157 and #72.


djibian (commented):
Draft review on `7b300b4e7cc23fb9e243b78c7affff30f4d8ea87` against the current #157 powered-session findings. The same-port activation ordering, protocol>=12 precondition, host-gap bounds and terminal-on-epoch-loss behavior are useful, but one authority-safety invariant remains unenforced across object/connection replacement:

After an ambiguous first keepalive/fence failure, periodic enqueue failure, missed deadline, or connection-epoch rotation, the class becomes terminal only **for that guard instance**. A caller can construct a new `EmergencyWatchdogLivenessGuard` after reconnect and call `activate()` on the same still-powered STM. That contradicts the durable #157 boundary: reconnect/new connection epoch does not clear firmware watchdog state; ordinary authority must remain poisoned until an explicit powered-session reset/reboot boundary has been established.

Smallest repair boundary: make watchdog terminality belong to a trusted powered-session identity/lifecycle shared across guard instances, not just the connection epoch/object. Add a deterministic regression proving `terminal/ambiguous -> reconnect/new epoch -> new guard` emits no watchdog command and remains rejected on the same powered-session identity; only a separately explicit simulated powered-session reset/new identity may permit activation again. Do not automatically power-cycle as part of this guard.

Minor cleanup while touching the file: remove the unreachable `Thread.current_thread if False else False` expression in `stop_for_terminal_reboot()`.

This is a Draft finding, not a GO/NO_GO verdict.

djibian:
Draft repair constraint for `0f9d11b616ef3533fe12cefcb8755fe9b87cdbf5`: **powered-session identity distinct from connection epoch** is still missing from the implementation.

Authoritative #157 findings 5601698143 and 5601727773 refine the earlier same-port fence: after the first watchdog send may have reached firmware, an ambiguous activation/maintenance outcome, missed liveness deadline, or loss of lifecycle certainty cannot be recovered by constructing a new guard after Crazyradio reconnect. The firmware watchdog state survives link reconnect; ordinary reuse requires a proven STM+deck power-cycle/reboot boundary, then reconnect/re-preflight.

The current guard keys all validity only to `bound_connection_epoch`. A fresh guard after reconnect can therefore send again on the same powered firmware session, which is the exact recovery path the newer finding forbids.

Smallest repair boundary:
- introduce an explicit powered-session identity separate from connection epoch and make a terminal/ambiguous watchdog lifecycle unrecoverable for that identity;
- permit a new identity only as input from the future trusted reset layer after the separately proven STM+deck reset boundary (this PR need not implement/authorize the reset itself);
- bind the #257 reader to the exact Crazyflie object as well as the epoch, so the causal fence cannot be assembled from a different connection object that happens to expose the same token;
- preserve #257 as the exclusive authoritative GET_STATE path and no flight/teacher/browser authority.

Because this PR is Draft, this is a repair constraint rather than a candidate verdict.

djibian (commented):
NO_GO 36965a97379bc868f56a00b5f2114bffd3971d60

The earlier reconnect/object-replacement finding is only partially resolved. The current implementation keeps lifecycle state in the process-local module registry `_POWERED_SESSION_STATES`, and `PoweredSessionWatchdogLifecycle(identity)` creates an unseen identity directly in state `new`. Therefore lifecycle certainty is still recoverable without the separately proven STM+deck reset required by current main/#157:

- after host/runtime process restart, the registry is empty even if the same Crazyflie STM remains powered and its firmware watchdog may already be active/locked/ambiguous;
- even in one process, a caller can choose a different arbitrary string and immediately obtain a fresh `new` lifecycle. The regression named “after-explicit-reset” proves only a different string was supplied; it proves no reset boundary.

That violates the integrated boundary on current main: after watchdog state may have changed, ordinary authority must remain fail-closed until an explicit STM+deck power-cycle/reboot is separately established, then reconnect/new epoch and complete re-preflight. A host-process restart is itself loss of lifecycle certainty, not reset proof.

Smallest repair boundary: do not let this watchdog primitive mint trusted NEW powered-session state from an arbitrary identity. Fresh lifecycle state must be supplied/established by a distinct trusted powered-session/reset boundary whose evidence cannot be recreated merely by reconnecting, reconstructing Python objects, restarting the host process, or choosing another token. If that reset layer is intentionally out of scope for this PR, make the interface fail closed until such externally established fresh-session state is provided, and add deterministic regression for prior/unknown watchdog state -> host-process/lifecycle reconstruction -> no watchdog send/no activation without reset proof.

The same-port #257 fence, exact Crazyflie binding, host-gap checks, and same-identity reconnect poisoning otherwise address the earlier Draft findings. The PR is also now based on pre-#263 main and will require normal base refresh/fresh candidate evidence after repair.

djibian:
Draft finding on `55398a9054b9e222406ca7e899210c2837ff05ca`: the activation fence can still leave the firmware watchdog too close to expiry before periodic maintenance starts.

`activate()` sends the first keepalive, performs the fresh #257 GET_STATE fence, verifies only that the elapsed host gap is still `<= max_host_gap_seconds`, marks the lifecycle active, starts the liveness thread, and that thread waits one full `keepalive_interval_seconds` before its next send. The constructor permits any `interval < max_gap < 1.0 s`; it does not require `max_gap + interval < 1.0 s`. Therefore a valid configuration such as `max_gap=0.9`, `interval=0.8` can return successfully after a ~0.85 s fence, while the next keepalive is not attempted until well after the firmware's fixed 1 s timeout. Even the defaults `0.75 + 0.25 = 1.0` leave no scheduling margin at that boundary.

Smallest repair boundary: after the successful same-port fresh GET_STATE fence and fault/gap checks, enqueue one immediate post-fence keepalive **before** marking the powered session active / starting the periodic thread. The maintained deadline then begins from that exact enqueue rather than from the older fenced keepalive. Add a deterministic fake-clock regression where the fence consumes almost the permitted host gap and verify that the post-fence keepalive is immediate; include an extreme but currently valid interval/max-gap pair that would otherwise cross the firmware 1 s deadline.

This preserves the current external reset-proof contract, exact Crazyflie/#257 binding, powered-session terminality and no-authority surface. This is a Draft finding, not a GO/NO_GO verdict.

djibian (commented):
NO_GO ff55de72397b78a66ad30c0e27d4419bf4105776

The prior powered-session and post-fence re-anchor findings are resolved on this exact Ready HEAD, but one liveness-certainty recovery hole remains in `assert_live()`.

`assert_live()` reads the external powered-session lifecycle with `_read_powered_session_state()` before entering the block that turns liveness uncertainty into terminal state. If that trusted external state becomes unavailable or invalid, `_read_powered_session_state()` raises `WatchdogLivenessError` directly and the guard does **not** call `_record_terminal()`. A later successful state read can therefore make the same guard usable again even though watchdog lifecycle certainty was temporarily lost. That contradicts the durable #157 rule that lost watchdog-lifecycle certainty is terminal until the separately proven STM+deck reset boundary.

Smallest repair boundary: make any external powered-session state/reason read failure encountered by the live guard poison the local/external watchdog lifecycle via `_record_terminal()` before returning an error. Add a deterministic regression where an active lifecycle becomes unreadable during `assert_live()`, then becomes readable/`active` again, and prove the guard remains terminal with no recovery on that powered session. Preserve the current abstract external reset authority, same-port #257 fence, immediate post-fence re-anchor, and no flight/reset/browser authority.

djibian (commented):
GO ba20604442072f5663aa340db1980c23b4711fae

Independent review against current main@6239728a4625e77684bb3aa0b55dafb66968791b and the applicable earlier refutations.

The exact Ready HEAD resolves the remaining NO_GO: loss or invalidity of the external powered-session lifecycle during `assert_live()` is now converted into terminal watchdog certainty via `_record_terminal()` before the error escapes, and the regression proves later readability/`active` state cannot resurrect the same powered session. The earlier findings also remain resolved: fresh watchdog state is supplied by an abstract external reset/session authority rather than minted from an arbitrary token; reconnect/process reconstruction alone cannot prove NEW; #257 is bound to the exact Crazyflie object and epoch; the same-port keepalive -> fresh GET_STATE causal fence is followed by an immediate maintenance re-anchor; host-gap/enqueue ambiguity fails closed; no arming, flight, reset/power-cycle, commander, setpoint or browser/student authority is exposed.

This GO is bounded to the non-authority watchdog activation/liveness safety primitive and its deterministic tests; it does not establish the future trusted reset/session implementation, teacher authorization, physical effect authority or motorized evidence.

PR #264, opened 2026-09-09
Add fresh controlled landing completion evidence

djibian:
Adds a non-authority completion observer for the remaining #157 physical boundary.

Candidate scope:
- capture a fresh same-epoch #257 supervisor baseline proving physical flight under active high-level control before the future separately authorized landing effect;
- bind that baseline to exact observer issuance and one completion attempt so forged, superseded, stale or consumed same-epoch evidence cannot be replayed;
- after that effect, accept completion only from later fresh #257 evidence with no blocking fault, `is_flying=false`, `hl_control_active=false`, and `hl_traj_finished=true`;
- enforce one monotonic total deadline across blocking reads, including rejecting an otherwise valid state that arrives after that deadline;
- preserve stock CF2.1 auto-arming semantics by not treating a later armed/ReadyToFly state as a completion failure;
- fail closed on epoch rotation, poisoned/failing supervisor freshness, malformed state, blocking fault, or timeout;
- issue no Crazyflie command and expose no arming, landing, HighLevelCommander, setpoint, watchdog, teacher-authorization or browser/student API.

Deterministic regressions cover the high-level-flight→finished/non-flying transition, low-level/stale baseline rejection, exact one-shot evidence binding, auto-arm after landing, blocking faults, epoch rotation, reader failure, total-deadline overrun, still-active high-level control, malformed state, and the no-authority source boundary.

This does not issue or acknowledge the landing command and does not by itself prove watchdog continuity or teacher authorization; those remain separate gates. Refs #157 and #72.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -53,6 +53,10 @@ def main():
     if physical_supervisor_test.returncode:
         print('FAIL connection-epoch physical supervisor freshness',file=sys.stderr);print(physical_supervisor_test.stdout,file=sys.stderr);print(physical_supervisor_test.stderr,file=sys.stderr);return physical_supervisor_test.returncode
     print(physical_supervisor_test.stdout.strip())
+    physical_landing_completion_test=subprocess.run([sys.executable,'tools/ci/test_physical_landing_completion.py'],text=True,capture_output=True)
+    if physical_landing_completion_test.returncode:
+        print('FAIL controlled physical landing completion evidence',file=sys.stderr);print(physical_landing_completion_test.stdout,file=sys.stderr);print(physical_landing_completion_test.stderr,file=sys.stderr);return physical_landing_completion_test.returncode
+    print(physical_landing_completion_test.stdout.strip())
     physical_yaw_test=subprocess.run([sys.executable,'tools/ci/test_physical_yaw_observer.py'],text=True,capture_output=True)
     if physical_yaw_test.returncode:
         print('FAIL fresh physical yaw observation',file=sys.stderr);print(physical_yaw_test.stdout,file=sys.stderr);print(physical_yaw_test.stderr,file=sys.stderr);return physical_yaw_test.returncode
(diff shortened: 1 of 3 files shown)


djibian (commented):
Draft review on `e76e5e3f45aed27187cf8880b536cb5a9444177b`: two completion-evidence invariants need repair before this can be a stable candidate.

1. `PreLandingFlightEvidence` is a public `NamedTuple`, while `await_completion()` validates only its type and connection epoch. A caller can forge same-epoch evidence without `capture_pre_land_flight()`, or reuse an old baseline from an earlier completion on the same epoch. The observer can then accept a landed/HL-finished state although this observer did not establish the claimed fresh pre-effect flight baseline for this completion attempt. Bind accepted evidence to exact observer issuance and one attempt (e.g. internally retain the exact opaque evidence object/token and reject forged, superseded, stale or consumed evidence), with regressions for same-epoch forgery and reuse.

2. The total timeout is checked before each blocking `_read_fresh()`, but an otherwise valid completion state returned after the total deadline is promoted before the clock is checked again. Re-check the monotonic deadline immediately after every fresh read and before accepting its state. Add a fake-reader/clock regression where the final read advances beyond the deadline while returning a valid landed state.

The no-command boundary, exclusive #257 observation path, blocking-fault rejection, epoch binding and stock auto-arm handling otherwise fit the current #157 direction. This is a Draft finding, not a GO/NO_GO verdict.

djibian (commented):
GO c0fba688ea0508c1be6b2c7aa93b4db8c10dc3e2

Independent bounded review against main@4ba87b971a7c8a40debe92b4b69f572f39c837c1 and the current #157 controlled-completion boundary. The prior Draft findings are resolved on this exact Ready HEAD: pre-land evidence is bound by exact observer-issued object identity, superseded/failed recapture invalidates older evidence, consumption is one-shot, and the monotonic total deadline is rechecked immediately after each fresh read before promotion. Completion remains exclusive #257 observation only, same-epoch, fail-closed on malformed/fault/reader/epoch evidence, requires fresh high-level flight before later finished/non-flying/HL-inactive evidence, and correctly does not reject stock post-landing auto-arm. Source and harness add no landing command, commander/setpoint, watchdog, arming, teacher/browser or other execution authority.

PR #266, opened 2026-09-09
Add trusted powered-session reset authority

djibian:
Implements the next bounded #157 safety prerequisite without adding flight authority.\n\n- adds an ephemeral concrete `PoweredSessionWatchdogAuthority` that cannot be reconstructed into `new` state from reconnect/process restart or serialized identity;\n- adds a trusted-host reset/postcondition factory that requires an explicit known-inactive gate, invalidates prior evidence before the STM+deck reset attempt, opens a fresh session, rejects reuse of the prior connection epoch, revalidates exact Crazyflie 2.1/self-test/protocol evidence, requires exact-bound preflight re-assertion, and requires one fresh non-blocking supervisor observation before minting authority;\n- adds a lazy explicit cflib `PowerSwitch.stm_power_cycle()` adapter, so merely constructing the helper has no physical effect;\n- adds deterministic regressions for safe ordering, one-shot lifecycle consumption, ambiguous reset, stale epoch, identity/self-test/protocol/preflight/supervisor failures, cleanup, and non-restorability;\n- wires the regression into the existing Runtime core CI harness.\n\nThis is trusted host safety/reset infrastructure only. It exposes no browser/student endpoint, teacher authorization, arming, HighLevelCommander, movement, landing or motorized checkpoint.\n\nRefs #157, #262, #72.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -65,6 +65,10 @@ def main():
     if physical_watchdog_test.returncode:
         print('FAIL physical watchdog activation/liveness guard',file=sys.stderr);print(physical_watchdog_test.stdout,file=sys.stderr);print(physical_watchdog_test.stderr,file=sys.stderr);return physical_watchdog_test.returncode
     print(physical_watchdog_test.stdout.strip())
+    powered_session_test=subprocess.run([sys.executable,'tools/ci/test_physical_powered_session_authority.py'],text=True,capture_output=True)
+    if powered_session_test.returncode:
+        print('FAIL trusted physical powered-session/reset authority',file=sys.stderr);print(powered_session_test.stdout,file=sys.stderr);print(powered_session_test.stderr,file=sys.stderr);return powered_session_test.returncode
+    print(powered_session_test.stdout.strip())
     physical_http_test=subprocess.run([sys.executable,'tools/ci/test_physical_capability_http_bridge.py'],text=True,capture_output=True)
     if physical_http_test.returncode:
         print('FAIL read-only physical capability HTTP bridge',file=sys.stderr);print(physical_http_test.stdout,file=sys.stderr);print(physical_http_test.stderr,file=sys.stderr);return physical_http_test.returncode
--- a/tools/ci/test_physical_powered_session_authority.py
+++ b/tools/ci/test_physical_powered_session_authority.py
@@ -288,6 +288,9 @@ def test_explicit_cflib_power_cycle_helper_is_lazy_and_uri_bound() -> None:
         def stm_power_cycle(self) -> None:
             calls.append("cycle")
 
+        def close(self) -> None:
+            calls.append("close")
+
     reset = authority.make_cflib_stm_deck_power_cycle(
         "radio://0/80/2M/E7E7E7E7E7",
         power_switch_factory=FakePowerSwitch,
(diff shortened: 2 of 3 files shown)


djibian (commented):
NO_GO 799c99b551a56f5ee1f4032dcc861855b966b68c

The reset/postcondition ordering and external-watchdog-authority composition are otherwise coherent, but the concrete cflib reset adapter leaves the `PowerSwitch` transport open on the normal modern cflib path. At the pinned cflib commit, `PowerSwitch.__init__()` opens a link driver and `PowerSwitch.close()` is the explicit release path; `stm_power_cycle()` itself does not close it. `make_cflib_stm_deck_power_cycle()` constructs the switch and calls `stm_power_cycle()` without a `finally` close, then the factory immediately attempts to open the post-reset live Crazyflie session. That can retain the Crazyradio/link resource across the reset boundary and makes the claimed fresh post-reset session establishment unreliable/ambiguous.

Smallest repair boundary: close the exact `PowerSwitch` instance in `finally` after the power-cycle attempt, and fail closed if cleanup cannot be established before opening the post-reset session. Add deterministic coverage that close occurs on success and reset failure, and that a close failure does not permit authority establishment. Preserve the current known-inactive gate, prior-evidence invalidation, new-epoch/preflight/supervisor postconditions, and no-flight-authority surface.

djibian (commented):
GO 29adf428d6d26e3e9e1ff31bf4dc833971cd9a6e

Independent review against current `main@109e24b35a1295be6924abe307db8dfae1d5e76b` and the applicable #157 reset/watchdog findings.

The prior `799c99b...` refutation is resolved on this exact candidate: the cflib `PowerSwitch` transport is explicitly closed after the reset attempt on success and failure, cleanup failure is fail-closed before post-reset session establishment, and deterministic regressions cover those paths.

The bounded reset-authority claim is otherwise preserved: fresh authority cannot be directly reconstructed/minted, the trusted factory gates reset on explicitly known-inactive physical state, invalidates old evidence before the potentially ambiguous power-cycle, requires a genuinely new connection epoch plus live exact-airframe/self-test/protocol evidence, exact-bound preflight and fresh non-blocking supervisor observation, and only then returns a one-shot `PoweredSessionWatchdogAuthority`. Reset ambiguity or any postcondition failure creates no authority. No browser/student, teacher, arming, commander, movement or landing authority is added.

The future effect consumer still has to compose this setup/recovery-only reset boundary with the same trusted-host exclusion/state discipline so reset cannot race flight-capable effects; that future composition requirement does not refute this isolated reset primitive.

djibian (commented):
GO cba8705bf47760fa1bdc1647a41608101ec6f329

Independent review against current `main@d00ccf43a9a35465e06e00665b67659c5b347b5c` and the still-applicable #157/#262 reset/watchdog findings.

The prior `799c99b...` refutation remains resolved on this exact candidate: the concrete cflib `PowerSwitch` transport is closed after the STM power-cycle attempt on both success and failure; cleanup ambiguity fails closed before any post-reset session can establish authority; deterministic regression coverage exercises successful close, reset failure cleanup, and close failure.

The broader bounded claim is coherent and deliberately non-commanding. Fresh powered-session authority cannot be directly constructed or restored, the only minting path first requires explicit known-inactive physical state, invalidates prior evidence before the potentially ambiguous reset, then requires a genuinely live post-reset session, a new connection epoch relative to the supplied previous epoch, exact Crazyflie 2.1/self-test/protocol evidence, exact-bound preflight re-assertion and a fresh non-blocking supervisor observation. Any failed/ambiguous postcondition closes the new session and produces no authority. The returned lifecycle is one-shot for watchdog activation and terminal state is latching.

The branch also contains current main's integrated teacher-run authorization through ancestry; the actual diff from current main remains only the powered-session authority, its deterministic test and core-harness registration. No browser/student bearer, arming, HighLevelCommander, movement or landing authority is introduced.

No applicable refutation found for this exact head. Integration still remains subject to the current V4 merge preconditions, including exact base/main observation.

PR #267, opened 2026-09-09
Add host-only teacher run authorization binding

djibian:
Adds the smallest non-command teacher-authorization primitive for the remaining #157/#72 physical boundary.

Bounded claim:
- one explicit trusted-host teacher decision authorizes one exact physical run bound to `profile_id + canonical ast_binding + live connection_epoch`;
- every later flight-capable effect boundary must re-check that exact binding; any profile/AST/epoch mismatch permanently invalidates the receipt so correcting inputs cannot resurrect stale authority;
- one authorization can cover the primitives of that same exact run, avoiding an impractical teacher click per primitive while still checking the invariant before every effect;
- authorization is one-active-run/process-local and has no restore/import surface, so reconnect, reset, host restart or a later run cannot silently inherit a prior teacher decision;
- the implementation exposes no browser/HTTP endpoint, reuses no capability bearer, imports no cflib command surface and emits no arming, setpoint, watchdog, reset or flight effect;
- teacher UI/decision transport, concrete powered-session/reset authority and the actual physical effect consumer remain separate future boundaries.

The Runtime v2 core CI harness runs deterministic regressions for explicit approval/denial, exact binding, permanent mismatch invalidation, single-run lifecycle and anti-forging/no-effect surface checks.

Refs #157 and #72.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -73,6 +73,10 @@ def main():
     if physical_submission_test.returncode:
         print('FAIL session-bound physical submission bridge',file=sys.stderr);print(physical_submission_test.stdout,file=sys.stderr);print(physical_submission_test.stderr,file=sys.stderr);return physical_submission_test.returncode
     print(physical_submission_test.stdout.strip())
+    teacher_authorization_test=subprocess.run([sys.executable,'tools/ci/test_teacher_run_authorization.py'],text=True,capture_output=True)
+    if teacher_authorization_test.returncode:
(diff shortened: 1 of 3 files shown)


djibian (commented):
GO e54a62e75eb40656acbbe8eb200b5d680fbb57c9

Independent review against main@109e24b35a1295be6924abe307db8dfae1d5e76b. The candidate is a bounded host-only authorization primitive, not physical effect authority: one explicit trusted decision mints one process-local receipt for the exact profile/AST/connection-epoch run; every effect boundary re-checks that binding and any mismatch permanently invalidates it. Denial, decision failure, concurrent authorization races, stale receipts and later runs fail closed. The code exposes no browser/HTTP/cflib command surface and does not reuse the capability bearer. The caller-supplied binding remains deliberately non-authority; a later effect consumer must still compose it with the independently proven preflight, powered-session/watchdog and completion gates. No applicable refutation found on this exact head.

PR #271, opened 2026-09-09
Add pure HighLevel acknowledgement freshness domain

djibian:
Implements the smallest no-effect prerequisite from the current #157 SETPOINT_HL acknowledgement findings.\n\n- adds a process-wide connection-epoch acknowledgement freshness domain with one-at-a-time transaction serialization;\n- models the pinned firmware reply contract exactly: four bytes, surviving request prefix bytes 0..2 plus result byte;\n- treats exact zero/nonzero firmware results as definitive acceptance/rejection without confusing rejection with transport ambiguity;\n- poisons the current epoch after malformed/wrong-prefix reply, timeout/disconnect/unknown post-send outcome, epoch change, or unresolved emitted transaction, and prevents reconstructed same-epoch domains from bypassing that poison;\n- allows a new epoch to clear reply-matching ambiguity only, explicitly without claiming that the prior physical effect outcome became safe;\n- keeps pre-effect validation/concurrency rejection non-poisoning;\n- deliberately imports no cflib/CRTP command surface and performs no physical send, so the future trusted transport must still register its callback, mark the boundary immediately before one plain send, wait for the reply and compose the #266/#267/#262 execution gates;\n- intentionally rejects requests shorter than the three-byte prefix the firmware can deterministically echo; this bounded slice therefore does not claim acknowledgement correlation for the two-byte high-level STOP/group-mask requests, and STOP remains an exceptional motor-cut path rather than an ordinary first-consumer command;\n- adds deterministic regressions and wires them into the Runtime v2 core harness.\n\nNo browser surface, physical command authority, automatic retry or motorized checkpoint is added. Refs #157, #257, #262, #266, #267 and #72.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -89,6 +89,10 @@ def main():
     if physical_high_level_timing_test.returncode:
         print('FAIL physical HighLevelCommander timing policy',file=sys.stderr);print(physical_high_level_timing_test.stdout,file=sys.stderr);print(physical_high_level_timing_test.stderr,file=sys.stderr);return physical_high_level_timing_test.returncode
     print(physical_high_level_timing_test.stdout.strip())
+    physical_high_level_ack_test=subprocess.run([sys.executable,'tools/ci/test_physical_high_level_ack.py'],text=True,capture_output=True)
+    if physical_high_level_ack_test.returncode:
+        print('FAIL pure SETPOINT_HL acknowledgement freshness domain',file=sys.stderr);print(physical_high_level_ack_test.stdout,file=sys.stderr);print(physical_high_level_ack_test.stderr,file=sys.stderr);return physical_high_level_ack_test.returncode
+    print(physical_high_level_ack_test.stdout.strip())
     browser=browser_binary()
     with socketserver.TCPServer(('127.0.0.1',0),QuietHandler) as server:
         port=server.server_address[1]; threading.Thread(target=server.serve_forever,daemon=True).start(); time.sleep(.05); url=f'http://127.0.0.1:{port}/{HARNESS.as_posix()}'
--- a/tools/ci/test_physical_high_level_ack.py
+++ b/tools/ci/test_physical_high_level_ack.py
@@ -1,0 +1,201 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+from pathlib import Path
+import sys
+
+ROOT = Path(__file__).resolve().parents[2]
+MODULE_PATH = ROOT / "tools" / "physical" / "high_level_ack.py"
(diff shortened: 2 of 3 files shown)


djibian (commented):
GO d40f0e06a37658b3c4ef190e289db77fb9ffdcc0

Independent review against current `main@d4bf6aece31b39627e4894d70ab392e7816b1256` and the exact #157 SETPOINT_HL acknowledgement/freshness findings.

This candidate is correctly bounded as a **no-effect prerequisite**, not the physical transport itself. It models the pinned four-byte firmware response contract (request bytes 0..2 plus status), serializes one acknowledgement transaction process-wide per connection epoch, and keeps pre-emission validation/concurrency failures non-poisoning. Once the caller marks the effect boundary, malformed or wrong-prefix response, explicit timeout/disconnect ambiguity, unresolved close, or epoch rotation poisons that epoch; reconstructing another domain on the same epoch cannot bypass the poison. Exact zero and nonzero firmware results remain definitive acceptance/rejection rather than being conflated with transport loss.

The boundary after an epoch rotation is also stated correctly: a new epoch clears only reply-matching ambiguity and does not convert an unknown prior physical effect into known inactivity or reset authority.

Source inspection confirms the slice imports no cflib/CRTP send surface, performs no packet emission, exposes no browser/HTTP authority and cannot accidentally use cflib `expected_reply` auto-retry. The future effect-bearing transport still has separate obligations before one physical send, including #266/#267/#262/exclusion checks and a link-level duplicate-suppression precondition; that next-layer requirement does not refute this pure freshness domain.

No applicable refutation found on this exact head.

djibian (commented):
GO d40f0e06a37658b3c4ef190e289db77fb9ffdcc0

Independent review against current main@d4bf6aece31b39627e4894d70ab392e7816b1256 and the exact #157 SETPOINT_HL acknowledgement/freshness findings.

The candidate keeps the new boundary deliberately pure: it adds no CRTP/cflib send surface, but establishes the process-wide same-epoch serialization and stale-reply poison semantics that the later trusted transport must consume. Exact four-byte replies are bound to request bytes 0..2 plus status, zero/nonzero status remains definitive rather than transport-ambiguous, and timeout/disconnect/malformed/unresolved post-emission outcomes poison the epoch so a reconstructed same-epoch domain cannot accept a delayed reply. A rotated epoch clears only reply-correlation ambiguity and does not claim that the prior physical effect became safe. Pre-effect validation/concurrency failure remains non-authority and non-effect.

The nonblocking reservation is fail-closed rather than a hidden command queue, and source-level regressions explicitly exclude cflib, physical send, callback and expected-reply retry surfaces. No browser authority, reset authority, command emission or motorized checkpoint is introduced. No applicable refutation found on this exact head.

PR #272, opened 2026-09-09
Add live SafeLink physical precondition

djibian:
Implements the smallest no-effect prerequisite from the current #157 SafeLink finding before any ordinary physical SETPOINT_HL command can be eligible.

- binds one exact Crazyflie object to one live connection epoch;
- requires the exact live cflib connection to be positively connected on an explicit `radio://` URI;
- accepts duplicate-suppression evidence only when the exact current link reports `needs_resending is False`, matching pinned cflib's post-SafeLink state;
- rechecks link identity, SafeLink state, connection liveness and epoch before returning evidence so one observation is internally same-link/same-epoch;
- fails closed for missing/unknown/true/non-boolean retry state, disconnect, non-radio link, link replacement or epoch rotation;
- exposes only non-authority evidence and deliberately imports no cflib/CRTP send surface, callback, acknowledgement retry, arming, watchdog or HighLevel command API;
- adds deterministic regressions and wires them into the Runtime v2 core harness.

The later trusted effect consumer must call this precondition immediately before entering #271's acknowledgement transaction and must still compose #266/#267/#262, exact preflight, reset/effect exclusion, fresh #257/#260 evidence, #256/#268 semantics/timing and #257/#264 completion. This slice does not make an ambiguous prior effect safe and authorizes no motorized checkpoint.

Refs #157, #271, #266, #267, #262 and #72.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -93,6 +93,10 @@ def main():
     if physical_high_level_ack_test.returncode:
         print('FAIL pure SETPOINT_HL acknowledgement freshness domain',file=sys.stderr);print(physical_high_level_ack_test.stdout,file=sys.stderr);print(physical_high_level_ack_test.stderr,file=sys.stderr);return physical_high_level_ack_test.returncode
     print(physical_high_level_ack_test.stdout.strip())
+    physical_safelink_test=subprocess.run([sys.executable,'tools/ci/test_physical_safelink_precondition.py'],text=True,capture_output=True)
+    if physical_safelink_test.returncode:
+        print('FAIL live radio SafeLink duplicate-suppression precondition',file=sys.stderr);print(physical_safelink_test.stdout,file=sys.stderr);print(physical_safelink_test.stderr,file=sys.stderr);return physical_safelink_test.returncode
+    print(physical_safelink_test.stdout.strip())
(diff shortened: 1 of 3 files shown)


djibian (commented):
GO 04b4fb89f917f6ed7fde2ebac4992b60f1b63ed8

Independent review against current `main@2f8bce0e242ce9f3e0e2df43ffab09b9940e5872` and the exact #157 SafeLink duplicate-suppression finding.

The candidate is correctly bounded as a no-effect precondition. Against pinned cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873`, the exact live `Crazyflie.link` is the RadioDriver that publishes `needs_resending = not _has_safelink`; requiring `needs_resending is False` therefore positively establishes the intended SafeLink state rather than inferring it from link type or connection success alone.

The guard binds one Crazyflie object to one connection epoch, requires a positively connected explicit radio link, and rechecks link identity, SafeLink state, liveness and epoch before returning evidence. Missing, malformed, non-boolean or changing evidence fails closed. The source imports no cflib/CRTP send surface, emits no packet, grants no browser/teacher/watchdog authority and does not weaken #271's separate acknowledgement-freshness requirement or the unknown-effect/reset boundary.

The exact candidate's canonical CI Gate run completed successfully, including the final `CI Gate` job. No applicable refutation found on this exact head.

PR #273, opened 2026-09-09
Add trusted physical reset/effect exclusion domain

djibian:
Implements the smallest no-effect execution-order prerequisite from the current #157 physical backend boundary.\n\n- starts every host process in `recovery-required`, so reconstructed process state cannot mint flight authority;\n- runs the complete #266 reset-establishment transaction under the same process-local exclusion used by command boundaries;\n- runs all immediate command assertions inside that exclusion before an explicit `mark_emitted()` boundary;\n- blocks reset and later effects after positive acknowledgement until a separately supplied fresh completion proof establishes `flying` or `inactive`;\n- converts unresolved/exceptional post-emission outcomes and unavailable completion proof to `recovery-required`;\n- restores the exact prior stable phase only after a definitive firmware rejection;\n- proves reset/effect mutual exclusion with deterministic lifecycle/concurrency regressions and wires them into Runtime v2 core CI;\n- imports no cflib/CRTP send surface and emits no physical effect.\n\nThe future trusted effect consumer must compose this domain with exact preflight, #267 teacher binding, #266/#262 powered-session/watchdog state, #272 SafeLink, #271 acknowledgement freshness, fresh #257/#260 evidence where applicable, #256/#268 semantics/timing, one one-shot SETPOINT_HL send, and fresh #257/#264 completion. This PR grants no browser/HTTP authority and authorizes no motorized checkpoint.\n\nRefs #157, #266, #267, #272, #271, #262 and #72.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -97,6 +97,10 @@ def main():
     if physical_safelink_test.returncode:
         print('FAIL live radio SafeLink duplicate-suppression precondition',file=sys.stderr);print(physical_safelink_test.stdout,file=sys.stderr);print(physical_safelink_test.stderr,file=sys.stderr);return physical_safelink_test.returncode
     print(physical_safelink_test.stdout.strip())
+    physical_execution_domain_test=subprocess.run([sys.executable,'tools/ci/test_physical_execution_domain.py'],text=True,capture_output=True)
+    if physical_execution_domain_test.returncode:
+        print('FAIL trusted physical reset/effect exclusion domain',file=sys.stderr);print(physical_execution_domain_test.stdout,file=sys.stderr);print(physical_execution_domain_test.stderr,file=sys.stderr);return physical_execution_domain_test.returncode
+    print(physical_execution_domain_test.stdout.strip())
     browser=browser_binary()
     with socketserver.TCPServer(('127.0.0.1',0),QuietHandler) as server:
         port=server.server_address[1]; threading.Thread(target=server.serve_forever,daemon=True).start(); time.sleep(.05); url=f'http://127.0.0.1:{port}/{HARNESS.as_posix()}'
--- a/tools/ci/test_physical_execution_domain.py
+++ b/tools/ci/test_physical_execution_domain.py
@@ -1,0 +1,332 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+from pathlib import Path
+from threading import Event, Thread
+import sys
+
+ROOT = Path(__file__).resolve().parents[2]
+MODULE_PATH = ROOT / "tools" / "physical" / "physical_execution_domain.py"
+sys.path.insert(0, str(MODULE_PATH.parent))
+import physical_execution_domain as domain  # noqa: E402
+
+
+def require(condition: bool, message: str) -> None:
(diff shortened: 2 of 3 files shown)


djibian (commented):
NO_GO 0a0b467e15bcbd6e42e5a4ce21bb2c05fb29fd98

The per-instance `RLock` does not yet establish the durable design conclusion that reset establishment and flight-capable emission are mutually exclusive in one process-local execution domain.

Two independently constructed `PhysicalExecutionDomain` instances own different locks. While one instance is inside an effect boundary, a second instance can enter `run_reset_establishment()` concurrently. That re-opens exactly the reset/effect race this slice is intended to close; the current concurrency regression exercises only one instance.

Smallest repair boundary: make the exclusion process-wide for all domain instances that can target the physical host lifecycle (for example one module-global lock/state domain or construction constrained behind a single trusted factory), and add a deterministic regression using two separately constructed handles proving that reset cannot enter while the other handle owns an effect boundary. Preserve the existing fresh-process fail-closed and no-effect properties.

djibian (commented):
GO 9ec9310e546624382c4199e897958fce998ec868

Independent review of the exact Ready HEAD. The earlier NO_GO on 0a0b467e15bcbd6e42e5a4ce21bb2c05fb29fd98 is resolved: reset/effect lifecycle state and exclusion are now shared process-wide across independently constructed PhysicalExecutionDomain handles, and the added deterministic two-handle regression proves a reset cannot enter while another handle owns the effect boundary. The fresh-process recovery-required invariant, post-emission fail-closed transitions, exact prior-phase restoration only on definitive rejection, separate fresh-completion requirement, and no-cflib/no-CRTP/no-browser effect surface remain preserved. No broader flight authority is introduced by this slice.

djibian (commented):
GO 9ec9310e546624382c4199e897958fce998ec868

Independent review against current `main@667010ca4a6b55252013c23766ef87c2107ab955` and the still-applicable NO_GO on `0a0b467e15bcbd6e42e5a4ce21bb2c05fb29fd98`.

The prior refutation is resolved on this exact head. `PhysicalExecutionDomain` handles now share one module-level `_PROCESS_STATE` containing the single `RLock`, phase and active-section ownership. A separately constructed handle therefore cannot acquire an independent reset lifecycle while another handle owns an effect boundary. The added two-handle regression holds an effect through one handle and proves the other handle's reset callback cannot enter until that shared exclusion is released; both handles then observe the same stable phase.

The remaining lifecycle is fail-closed for the bounded no-effect claim: a fresh process starts `recovery-required`; immediate preconditions execute only after the shared exclusion is held; `mark_emitted()` moves to unresolved before caller-side emission; unresolved/exceptional post-emission state and uncertain completion require recovery; positive acknowledgement alone blocks reset/later effects until a fresh completion proof; definitive firmware rejection alone restores the exact prior stable phase. Reset establishment while flying or awaiting completion is rejected.

The module still imports no cflib/CRTP command surface and exposes no browser/HTTP effect path. Its reset callback remains explicitly the later trusted #266 composition point rather than an independent flight authority; the actual effect consumer must still supply the already-established #266/#262/#267/#272/#271 and action-specific fresh gates described by #157.

No blocking counterexample found for this exact Ready candidate. CI evidence remains a separate integration prerequisite.

PR #268, opened 2026-09-09
Add pure HighLevel trajectory timing policy

djibian:
Implements the bounded no-effect timing slice established on #157 before direct physical command consumption.

- models the pinned smooth rest-to-rest seventh-order HighLevelCommander planner peak/average speed factor (`35/16`);
- keeps the already-proven Runtime v2 `set_speed` meaning as horizontal-move speed only, default 0.35 m/s and bounded 0.1–0.35 m/s;
- derives horizontal `go_to` duration so the smooth trajectory peak speed does not exceed the selected student speed;
- derives turn duration against the existing Runtime v2 0.7 rad/s yaw-rate ceiling, independent of horizontal `set_speed`;
- deliberately leaves vertical/takeoff/landing duration policy unimplemented because the product has no established student speed semantic for those axes/actions;
- imports no cflib command surface and emits no effect or authority;
- adds deterministic math/bounds regressions to the Runtime v2 core harness.

The future effect consumer must still serialize commands from fresh supervisor completion and compose exact preflight, teacher authorization, powered-session/watchdog, fresh yaw, #256 geometry and #257/#264 completion. No motorized checkpoint is authorized.

Refs #157 and #72.

PR #276, opened 2026-09-09
Add trusted one-shot SETPOINT_HL effect transport

djibian:
Introduces the first bounded ordinary physical SETPOINT_HL effect surface for #157, composed inside the integrated #283/#280 physical-host trust root and reconciled with the exact-program sequencing boundary from #295.

The importable `TrustedSetpointHlTransport` remains fail-closed by itself: it accepts no bridge, responder credential, provenance object/client, teacher decision, reset-safety assertion or caller-selected raw command. Production `serve_physical_host.py` owns the one live reset-aware #278 bridge and passes it only into the trusted `physical_run_activation.py` adapter. That adapter defines the process-local `_HostBoundInflightTransport`, whose current-program hook performs a fresh exact #249 profile/AST/epoch assertion through the host-owned bridge immediately inside each effect transaction.

The same trusted activation first validates the complete #295 canonical physical-program envelope before reset or flight effect, then uses the integrated causal takeoff lifecycle to establish the exact same-epoch #266 powered session, #267 teacher authorization, #262 watchdog, #257 supervisor reader, #272 SafeLink, #271 acknowledgement domain and #273 execution domain. Only after causal takeoff has completed into `FLYING` does it expose a process-local controller for in-flight motion.

Ordinary caller IPC can request only parameter-free `execute-next-inflight`. It cannot supply direction, distance, angle, bridge/provenance data or any effect authority. The trusted controller reserves the next eligible move/turn from the exact #267 canonical AST through `PhysicalProgramSequence`; accepted completion advances the sequence, definitive pre-effect/rejected outcomes do not, and ambiguous outcomes poison the sequence rather than manufacturing a retry. The exact terminal landing is validated by #295 but remains outside this bounded #276 slice.

The ordinary motion surface is narrow: already-flying relative `GO_TO_2` horizontal moves and yaw turns only. Horizontal motion derives geometry from #256 using one fresh same-epoch #260 yaw sample and duration from #268; turns use #256/#268 directly. TAKEOFF_2, LAND_2, vertical movement, STOP, trajectory/spiral and raw SETPOINT_HL bytes remain unavailable on the #276 transport.

Immediately before emission, fresh host-local #278/#249 provenance is followed by exact teacher/watchdog rechecks, fresh #257 no-fault/flying/high-level/previous-trajectory-finished evidence, another mutable-authority recheck and live #272 SafeLink. The SETPOINT_HL reply callback is installed before crossing the effect boundary, followed by exactly one plain `send_packet(packet)` call with no application retry. A positive acknowledgement yields only the private #279 completion permit; completion requires causally fresh #257 evidence and returns #273 to `FLYING`, while timeout/disconnect/malformed reply/send or completion uncertainty fails closed.

Deterministic coverage exercises the actual production host control flow: caller-selected motion fields are rejected, parameter-free steps execute the exact post-takeoff AST turn/move order, #260 yaw opens only for horizontal movement, fresh #278/#249 guards the in-flight effect, malformed terminal structure fails before takeoff, and the earlier process-boundary forgery regressions remain active. The direct importable effect core still cannot emit without trusted production composition.

No motorized checkpoint is requested or authorized.

Refs #157. #280/#283/#293/#295 are integrated; #281 is superseded.


djibian (commented):
NO_GO 70731c9443071a8988497e52469ba6d85855d3c8

Reviewed against current main@38e3229a6efbbe59444efcea747d985957f7c146. The candidate correctly establishes the one-shot callback/acknowledgement mechanics, SafeLink re-check, no-`expected_reply` send path and fail-closed ambiguity handling, but the first ordinary physical-effect surface still lets its caller self-supply the very authorities it is supposed to enforce.

`send_once()` accepts arbitrary callables for exact preflight, teacher binding, powered-session and watchdog assertions; `lambda: None` satisfies all four. It also accepts an arbitrary raw-byte `request_builder`, so the same caller can bypass #256/#268/action-specific semantic and timing bounds while still producing a syntactically allowed TAKEOFF_2/LAND_2/GO_TO_2 payload. The module can therefore emit a flight-capable packet without a concrete #267 receipt, #266/#262 lifecycle object, or structurally validated semantic/timing input. Because this is the first trusted effect surface, relying on callers to voluntarily provide the correct checks makes the authority boundary bypassable rather than composed.

Smallest repair boundary: bind concrete trusted evidence/objects rather than assertion-shaped callables for the mandatory authorities, and stop accepting arbitrary command bytes as the ordinary motion API. The effect method should derive or consume a narrowly typed/validated command representation whose construction enforces the integrated #256/#268 semantics and required fresh action evidence, while preserving the existing #272 -> #271 -> callback-before-one-send mechanics. Add a deterministic counterexample proving no-op/fake authority inputs and out-of-policy-but-correct-size packets cannot reach `send_packet`. No motorized test is needed for this repair.

djibian:
Draft finding f5c75d9d7642ef5b162bd70b1da9c4d667630d96

The concrete-authority repair is moving in the right direction, but two bypasses remain on this exact Draft head and should be closed before Ready:

1. `current_preflight_binding_reader` is still an arbitrary callable. A caller can supply `lambda: teacher_authorization.binding`; that satisfies the exact `PhysicalRunBinding` equality without re-asserting the live #249 preflight/workspace state. This recreates the old authority-by-callback problem specifically for exact preflight. The first effect surface needs a concrete trusted host preflight/session object or another non-forgeable live re-assertion boundary, not a caller-selected function returning the expected tuple.

2. Fresh #257 supervisor safety is still optional via arbitrary `action_preconditions`. `_authority_checks()` only tests the cached #273 phase `FLYING`; that phase was established by an earlier completion proof and does not prove that no blocking supervisor fault appeared before the next motion. Bind the exact same-epoch `FreshSupervisorStateReader` and require a fresh non-blocking/flying observation immediately before emission (after request/fresh-yaw construction), alongside the final #267/#262 re-check already required by issue #157 comment 5608065756.

Also bind `powered_session.session is crazyflie` rather than only checking that `powered_session.session` is non-null; `EstablishedPoweredSession` itself is a public dataclass and its live-session association should be exact at this first effect boundary.

These are deterministic host-side repairs only; no motorized checkpoint is implicated.

djibian:
Draft finding 7637900c5c7437d339748c21054eb083f67193f2

The production transport has advanced beyond the original raw-callback design, but `tools/ci/test_physical_setpoint_hl_transport.py` is still the old 430-line fixture for `send_once(...)`: it constructs `TrustedSetpointHlTransport` without the now-required concrete #267/#266/#262/#257 objects and exercises the removed public raw-send API. The successful workflow on this Draft is only `CI Gate (Draft)`; Runtime/Webots suites are skipped, so it does not execute this stale regression file.

Before Ready, replace the test with fixtures for the current exact constructor/public `send_horizontal_move`/`send_turn` surface and cover the already-recorded refutations: concrete authority identity/epoch/CF binding, final post-construction teacher/watchdog recheck, fresh #257 flying/non-blocking state, lost SafeLink, one plain send/no retry, definitive reject vs ambiguous reply, and absence of any public raw request path. The PR body should also be refreshed because it still claims TAKEOFF_2/LAND_2 and caller-supplied mandatory assertions that are no longer present.

No physical checkpoint is needed for this repair.

djibian:
Draft deterministic finding 308caaa8783d27453ab0c9dc4a204593f0e212d2

The rewritten transport regression is now aligned with the concrete API, but its final source-surface oracle is internally inconsistent with this exact HEAD and will fail when Runtime CI is actually enabled:

- it requires `source.count("self._cf.send_packet(packet)") == 1`, while production now deliberately resolves `send_packet = getattr(self._cf, "send_packet", None)` and calls `send_packet(packet)`, so that count is zero;
- it forbids the substrings `TAKEOFF_2` and `LAND_2`, while the module docstring explicitly states that those commands are *not* exposed, so both forbidden strings are necessarily present.

Make the oracle structural rather than matching those stale text fragments (for example assert the single local `send_packet(packet)` effect site and absence of public takeoff/land methods/raw command API). This is deterministic only; no physical test is implicated.

djibian:
Draft refutation 71356cacda54bcca9b2a2867164889672c74c013

The new `LiveCurrentProgramPreflightGuard` does not yet close the recorded #249/preflight bypass. The private mint key protects direct construction of the guard, but the public `TrustedCurrentProgramPreflightFactory` can itself be constructed by any caller with an arbitrary `Callable[[], PhysicalRunBinding]`. Consequently the exact counterexample from the earlier finding still works one layer higher: construct the factory with `lambda: teacher_authorization.binding`, call `bind(teacher_authorization.binding)`, and the resulting privately-minted guard will positively re-assert forever without invoking the production #249 workspace/preflight path.

Calling that callable a “trusted-host adapter” in the type/docstring does not make the trust boundary non-forgeable. Before Ready, the effect path needs a concrete production object/state that itself owns or verifies the live #249 exact-program/session assertion, or a minting path that ordinary effect callers cannot manufacture from an assertion-shaped callback. A deterministic regression should construct the lambda-backed factory counterexample and prove it cannot yield an effect-eligible preflight guard.

This is a host-side design refutation only; no physical checkpoint is implicated.

djibian:
Draft deterministic finding 9ddaa824258a975afe6bf7dbd38b296e0fd25523

The transport is now wired to the new concrete preflight-guard parameter names, but the rewritten regression fixture was not updated with that transition. On this exact HEAD, `Fixture` still constructs `TrustedSetpointHlTransport` with `safelink_precondition=` and `current_preflight_binding_reader=`; production now requires `safelink_guard=` and `current_preflight_guard=`. Runtime CI will therefore fail at fixture construction before exercising any of the intended transport oracles.

Update the fixture to the exact constructor and add the guard object explicitly. When doing so, do not hide the already-recorded 71356ca refutation: a guard minted through the public lambda-backed factory is not sufficient evidence that the real #249 current-program path was re-asserted.

This is deterministic host-side repair only; no physical checkpoint is implicated.

djibian:
NO_GO 1012bee4d2f75276bb37461ea3de31594b0bb3ac

Independent review against current main@38e3229a6efbbe59444efcea747d985957f7c146.

The candidate resolves the original raw-command/caller-supplied teacher-powered-watchdog surface, adds the required post-construction #267/#262 recheck, fresh same-epoch #257 flying/no-fault evidence, exact powered-session/Crazyflie binding, narrow #256/#268 GO_TO_2 semantics, SafeLink immediately before #271, and one plain no-retry send. The stale transport regression and source-oracle findings are also repaired on this HEAD.

One applicable authority refutation remains unresolved: `TrustedCurrentProgramPreflightFactory` is a public constructor that accepts an arbitrary `Callable[[], PhysicalRunBinding]`, and the current deterministic fixture itself demonstrates the bypass by constructing it with `lambda: self.current_binding`. That public factory can mint a valid `LiveCurrentProgramPreflightGuard` without invoking the production #249 workspace/profile/session re-assertion. The private guard mint key therefore moves, but does not close, the assertion-shaped preflight trust boundary. A flight-capable effect caller can still manufacture the required preflight guard from the already-known teacher binding.

Smallest repair boundary: make effect-eligible current-program evidence originate from a concrete production #249 state/object or a minting path ordinary effect callers cannot manufacture from an arbitrary binding-returning callback. Add the recorded counterexample as a negative regression: a lambda-backed/publicly fabricated adapter must not be able to produce a guard accepted by `TrustedSetpointHlTransport`. No motorized test is needed.

djibian:
Draft architecture finding f89825ac0f09cb9dcb68c59011be1ca534c31df8

The remaining preflight provenance problem is not local to `current_program_preflight.py`: on current main, the concrete #249 exact-current-program re-assertion exists only in the browser path (`physical_preflight_runtime.js -> physical_submission_bridge.assertCurrentWorkspace()`). The Python host surface deliberately exposes only read-only capability/epoch reads; there is no production host object or protocol today that can invoke/consume that exact #249 assertion. Code search outside this PR consequently finds no production source for `TrustedCurrentProgramPreflightFactory`; the only source on the candidate is the deterministic lambda fixture.

That means wrapping the callback in another private-mint Python type cannot close 71356ca/1012bee: provenance has to cross the existing browser/host boundary explicitly, or this effect slice is premature. The smallest sound next boundary is therefore a non-authority #249 handoff owned by production integration (exact profile/AST/epoch, fresh re-assertion, fail-closed invalidation) whose host-side receipt/guard cannot be minted from an arbitrary effect-caller callback. Keep that handoff distinct from the browser-held capability bearer and from teacher/effect authority. If adding that production handoff would materially broaden #276, split it as a prerequisite and leave the transport Draft rather than weakening the trust claim.

A negative regression should demonstrate the exact current counterexample: a lambda returning the already-authorized `PhysicalRunBinding` must not be able to create any object accepted as fresh #249 evidence by the effect transport. No physical checkpoint is implicated.

djibian:
Draft refutation f89825ac0f09cb9dcb68c59011be1ca534c31df8

A second raw-effect bypass remains in the public `packet_factory` test seam. `_validate_packet()` reads `packet.port` and `packet.data` once, then several later safety/acknowledgement operations occur before the same caller-supplied object is handed to `Crazyflie.send_packet(packet)`. The module therefore does not actually guarantee the claimed byte-for-byte request at the physical boundary.

Deterministic counterexample: a caller-supplied packet object can expose port 0x08 and return the validated GO_TO_2 bytes on the first `data` read, then return different bytes (for example a STOP/TAKEOFF-shaped payload) when `send_packet` later consumes the object. The same issue exists with an ordinary mutable packet changed after `_validate_packet()`. No concurrency or malformed authority object is required; the constructor explicitly accepts the factory.

Smallest repair boundary: do not let a caller-supplied packet object cross the physical effect boundary. Keep any deterministic seam before byte validation (for example build validated request bytes only), then construct the exact cflib `CRTPPacket` internally at the last boundary from an immutable local `bytes` value, or otherwise prove/copy the exact bytes into an internally-owned packet immediately before the one send. Add a flipping/mutable-packet counterexample proving alternate command bytes cannot reach `send_packet`.

This is deterministic host-side only; no motorized checkpoint is implicated.

djibian:
Draft architecture refinement 09c9b5be6192b61a287a5e2c4b7b1801f7675695

A one-way “browser posts a fresh #249 receipt to the host” handoff is not sufficient for this transport’s existing freshness claim. `_send_go_to_once()` deliberately rechecks `_assert_current_authority()` both on entry and again **after** request construction/fresh yaw, because workspace/profile/watchdog/teacher state may change while that construction is in progress. A cached browser assertion obtained before the builder would therefore be stale at the second boundary by construction.

The production handoff needed to close 71356ca/1012bee must support a fresh host-initiated round trip at the actual effect boundary (or an equivalent control-flow structure): host issues a one-shot nonce/challenge, the already-integrated #249 browser runtime runs its exact `assertCurrentProgram()` against the then-current profile/workspace/session, and only that matching response can satisfy the pending host check. The browser-held non-authority bearer may answer a host challenge, but must not be able to create the challenge or invoke an effect. Timeout, mismatch, reconnect, changed workspace/profile, duplicate/late response or responder loss must fail closed.

This is a prerequisite/protocol-shape conclusion, not a request to broaden #276 blindly. If the round-trip cannot remain a small bounded repair, keep #276 Draft and integrate the non-authority host↔#249 handoff separately first. No physical checkpoint is implicated.

djibian:
Draft completion-boundary finding 09c9b5be6192b61a287a5e2c4b7b1801f7675695

The transport now blocks a new motion until #273 is back in `FLYING`, but the only way out of `AWAITING_COMPLETION` is still the public `PhysicalExecutionDomain.complete_accepted_effect(next_phase, prove_completion)` API, whose proof is an arbitrary caller-supplied callable. The deterministic fixture currently demonstrates the bypass directly with `complete_accepted_effect(FLYING, lambda: True)`.

Once this PR adds the first ordinary flight-capable send surface, that is no longer a harmless lower-level test seam: after one positive SETPOINT_HL acknowledgement, caller code can manufacture “completion”, restore `FLYING`, and issue another `GO_TO_2` while the prior high-level trajectory may still be active. The fresh pre-send #257 check currently requires only `blocking_fault is False` and `is_flying is True`; it does not require `hl_traj_finished`, so it does not close the overlap counterexample. This conflicts with the durable #157 requirement that high-level effects be serialized and fresh #257 completion be established before the next motion.

Smallest repair boundary before this effect surface can become Ready: make accepted-motion completion concrete rather than assertion-shaped. Either the transport owns a same-epoch #257 completion observer and only restores `FLYING` after a fresh state proves the previous high-level trajectory finished while flight remains healthy, or #273 gains a non-forgeable typed completion evidence path that this transport alone consumes. Add a deterministic regression showing that `is_flying=True` with `hl_traj_finished=False` cannot authorize the next send and that a caller-supplied `lambda: True` cannot restore motion eligibility.

Separately, after #277 is integrated, keep the slow current-program round trip before the final fresh supervisor observation where possible; today `_read_fresh_flying()` runs before a second potentially blocking preflight re-assertion, so the supervisor observation can age while that round trip waits. No physical checkpoint is implicated.

djibian:
Draft refutation ab7bd8a65fd367f0e6295595ff3db7a827bb65b6

The opaque accepted-effect claim closes the old `lambda: True`-without-claim bypass, and automatic post-ack completion is the right shape, but the new completion oracle can still accept the **previous** trajectory's finished bit as proof of the newly accepted GO_TO_2.

On pinned firmware `54f31e243a0b28b67efef5ba20dbb6d9890a5478`, `plan_go_to_from()` installs a new planner trajectory/begin time immediately, while supervisor bit 9 is refreshed later in `updateLogData()` from `crtpCommanderHighLevelIsTrajectoryFinished()`. Before the command, a completed prior GO_TO naturally has `hl_traj_finished=true`. The first fresh GET_STATE after the SETPOINT_HL application reply can therefore still expose that prior true bit before the supervisor loop reflects the newly running trajectory.

Current `_await_motion_completion()` accepts the first `is_flying=true && hl_traj_finished=true` observation and restores #273 to `FLYING`. It also does not require `hl_control_active=true`. This leaves a deterministic overlap counterexample despite the private claim.

Smallest repair: after positive acknowledgement, require fresh same-epoch #257 evidence of the newly accepted trajectory **in progress** at least once (`blocking_fault=false`, `is_flying=true`, `hl_control_active=true`, `hl_traj_finished=false`), and only then accept a later fresh healthy `is_flying=true`, `hl_control_active=true`, `hl_traj_finished=true` state. If the unfinished phase is never observed before the bounded timeout, fail closed rather than treating an already-true bit as completion. Add the exact stale-true -> false -> true regression and a stale-true-only timeout regression.

This is deterministic host-side evidence; no physical checkpoint is implicated.

djibian:
Draft trajectory-completion freshness finding ab7bd8a65fd367f0e6295595ff3db7a827bb65b6

The new private-claim completion path closes the unusable `AWAITING_COMPLETION` state, but the first post-ack supervisor sample can still falsely complete the just-issued motion.

Immediately before emission this candidate requires `hl_traj_finished=True` for the previous trajectory. After a positive firmware acknowledgement, `_await_motion_completion()` accepts the very first fresh #257 sample with `hl_traj_finished=True`. There is no causal fence proving that this bit belongs to the newly acknowledged GO_TO_2 rather than the pre-command finished state persisting briefly until the supervisor bitfield refreshes. The host-planned duration is used only as a timeout, so it does not close that race.

Smallest repair: after acknowledgement, require either (a) at least one fresh same-epoch observation with `hl_traj_finished=False` before later accepting `True`, or (b) if the motion completes before that in-progress state can be observed, do not accept `True` until the exact planned duration has elapsed and a fresh #257 read still reports finished. Duration is only a causal lower bound in case (b); the final completion evidence remains the fresh supervisor bit.

Also include `hl_control_active=True` and `self._watchdog.assert_live()` in the post-ack completion loop. A lost watchdog lifecycle or loss of high-level control must fail closed even if `is_flying=True` and `hl_traj_finished=True`.

Deterministic counterexample: pre-send finished=True; first post-ack read finished=True at t≈0 must not complete. Then either observe false→true, or advance to the planned-duration bound and require a new finished=True read. No motorized test is needed.

djibian (commented):
NO_GO d1cc07dcaaa253507053d15acee631d4a0a22be4

Independent review against current main@c5c3ab23fda657aef99e00bec44ece8a33387557 and the applicable #276 authority findings.

The completion work is now substantially better: this candidate consumes current main's #279 permit, requires active high-level control, keeps watchdog liveness in the completion loop, and causally fences the post-ack finished bit with false→true or a conservative full-duration lower bound. The latest mutable-authority ordering also keeps the final supervisor read between fast teacher/watchdog checks.

One integration-blocking authority refutation is still unchanged: the transport still imports candidate-local `current_program_preflight`, requires `LiveCurrentProgramPreflightGuard`, and the deterministic fixture still mints that guard through public `TrustedCurrentProgramPreflightFactory(lambda: self.current_binding)`. That is the exact 1012bee/71356ca counterexample: the effect caller can manufacture “current program” evidence without running the production #249 browser re-assertion.

Current main now contains #278's concrete trusted-host boundary, `ReadOnlyCapabilityHttpBridge.assert_current_program(profile_id, ast_binding, connection_epoch)`, which performs the host-initiated one-shot #249 round trip and returns private immutable `CurrentProgramPreflightEvidence`. #276 must consume that integrated concrete provenance (or an equally non-forgeable production object), not preserve the superseded callback/factory layer.

Smallest repair: remove the candidate-local preflight factory/guard from the effect boundary and tests; bind the exact integrated #278 host bridge/evidence to the #267 run binding, with the slow #249 round trip before the final fresh #257 observation and the fast teacher/watchdog recheck afterward. No motorized checkpoint is required.

djibian:
Draft authority refutation f6f7ce133bbcbe45e44a32bea8107cd88e7bb92e

The candidate now exercises the integrated #278 host-initiated round trip instead of the old lambda-backed preflight fixture, but the effect boundary is still forgeable one layer lower.

`TrustedCurrentProgramPreflightFactory` accepts any exact `ReadOnlyCapabilityHttpBridge`. That bridge is itself publicly constructible, does not require an exact production `ReadOnlyCapabilitySession`, exposes its generated `preflight_responder_token`, and exposes the host-side challenge/assertion methods on the Python object. An ordinary host effect caller can therefore construct its own bridge around an arbitrary session-like object whose `read_connection_epoch()` returns the authorized epoch, create a guard for the known #267 `PhysicalRunBinding`, and answer the bridge's own challenge with matching profile/AST/epoch values without ever invoking the production browser #249 `assertCurrentProgram()` path. The resulting `LiveCurrentProgramPreflightGuard` is accepted by `TrustedSetpointHlTransport`.

So f6f7 closes the old arbitrary-callback shape in the test fixture, but it does not yet establish non-forgeable production provenance for effect-eligible current-program evidence. This is the same trust invariant as 71356ca/1012bee, now moved to a publicly constructible bridge object.

Smallest repair boundary: the object accepted by the effect transport must be minted/bound by a production integration path ordinary effect callers cannot reproduce from public constructors or responder credentials. A deterministic regression should construct a separate public `ReadOnlyCapabilityHttpBridge` with a fake epoch source and self-answering responder and prove that it cannot yield any object accepted as fresh #249 evidence by the transport. Preserve the host-initiated nonce, exact profile/AST/epoch checks, final pre-effect ordering, and non-authority browser bearer.

No motorized checkpoint is implicated.

djibian:
Repair evidence cc29841ecdbfbe7727ab92c4318dc9c5334ada10

This exact candidate addresses Draft authority refutation 5609091379 without reintroducing the superseded callback/factory layer.

The integrated #278 module now exposes a private-minted, non-authority `CurrentProgramEffectPreflightHandle`. `TrustedSetpointHlTransport` accepts only that exact handle, not a raw/public `ReadOnlyCapabilityHttpBridge` or responder credential. The handle delegates every effect-boundary assertion to the underlying integrated host-initiated #249 round trip and exposes no bridge/token surface.

The deterministic regression reproduces the recorded counterexample rather than assuming it away: it constructs a separate public #278 bridge around an arbitrary epoch source, self-answers a host challenge and proves that this bridge can indeed mint ordinary #278 evidence. Direct construction of the effect handle with a caller key fails, and passing the raw forged bridge to the transport is rejected before any physical effect.

The earlier completion and freshness repairs remain unchanged: final current-program round trip precedes the final #257 observation; fast teacher/watchdog authority is rechecked afterward; accepted motion keeps the #279 permit private; #257 completion is causally fenced with in-progress→finished or the conservative planned-duration lower bound, while watchdog/high-level-control/fault evidence remains live.

This execution mutated the candidate and therefore supplies no independent GO for this HEAD.

djibian:
Draft authority refutation cc29841ecdbfbe7727ab92c4318dc9c5334ada10

The new handle does not yet close the recorded f6f7ce trust invariant, because its minting seam is still directly callable by the same ordinary Python code the handle is meant to exclude.

`serve_reference_capabilities._bind_effect_current_program_bridge(bridge)` is a module-level function. The leading underscore is only a naming convention; any caller that can import this module can call it. The deterministic fixture itself demonstrates that exact operation to obtain the effect-eligible handle.

Therefore the f6f7 counterexample still composes end-to-end: an ordinary host caller can construct a public `ReadOnlyCapabilityHttpBridge` around an arbitrary epoch source, self-answer its own #249 challenge, then call `_bind_effect_current_program_bridge(forged_bridge)` and obtain the exact `CurrentProgramEffectPreflightHandle` accepted by `TrustedSetpointHlTransport`. Rejecting direct constructor use and rejecting the raw bridge do not help while this callable mint path remains reachable.

Smallest repair boundary: effect-eligible current-program provenance must be created only by the actual production composition path and must not expose any callable/module attribute that can turn an arbitrary public bridge into the accepted handle. A deterministic regression should take the self-answering forged bridge already used by this candidate, then attempt every exposed binding/mint path available to normal imported Python code and prove none can produce an object accepted by the transport.

The completion/freshness work on this HEAD remains useful and is not refuted by this finding. No motorized checkpoint is implicated.

djibian (commented):
NO_GO a53ec004140033b6750861f14ab3f32e927bd904

Independent review against current main@c5c3ab23fda657aef99e00bec44ece8a33387557 and all still-applicable #276 findings.

The only delta from cc29841 is a deterministic supervisor-regression variable-shadowing fix. It does not touch the current-program effect provenance path, so Draft authority refutation cc29841ecdbfbe7727ab92c4318dc9c5334ada10 remains fully applicable.

The effect handle is still mintable by ordinary imported Python code through the module-level callable `serve_reference_capabilities._bind_effect_current_program_bridge(bridge)`. A leading underscore is not an access boundary. The recorded counterexample therefore remains end-to-end: construct a public `ReadOnlyCapabilityHttpBridge` on a fake epoch source, self-answer its challenge, call that binder, then pass the returned exact `CurrentProgramEffectPreflightHandle` to `TrustedSetpointHlTransport`.

The candidate must remain Draft until no module-reachable constructor/binder can turn an arbitrary caller-created/self-answered bridge into effect-eligible current-program provenance. The regression must exercise the binder itself, not only wrong-key direct construction and raw-bridge rejection.

The unrelated transport repairs remain useful and are not refuted: typed GO_TO_2 only, exact concrete teacher/powered/watchdog/SafeLink/#271 composition, one plain no-retry send, private #279 completion permit, and causal fresh #257 completion with watchdog/high-level-control/fault checks. No motorized checkpoint is required.

djibian (commented):
NO_GO d57629e9aef79e93e7709c99baf364a0e84ec4d4

Independent review against current main@c5c3ab23fda657aef99e00bec44ece8a33387557 and the still-applicable current-program provenance refutation.

The new `effect_preflight_production_backed` flag does not establish production provenance. It is derived only from whether optional constructor seams were omitted. A normal imported caller can construct `ReadOnlyCapabilitySession("radio://...")` with no seams, obtaining `effect_preflight_production_backed=True`, then populate the same reachable Python state used by this candidate's deterministic fixture (`_scf`, `_connection_epoch`) and build/self-answer a public `ReadOnlyCapabilityHttpBridge`. The module-level `_bind_effect_current_program_bridge()` then accepts that caller-created session and returns the exact handle consumed by `TrustedSetpointHlTransport`.

In fact the candidate test fixture demonstrates the critical construction pattern itself: uninjected public session constructor followed by direct internal-state injection. The current negative test uses a different `FakeCapabilitySession`, so it does not exercise the strongest available counterexample against the new flag.

This is not a naming/privacy problem that another boolean can close in the same importable object graph. The smallest sound boundary is the one already identified by the prior refutation: production effect composition must own the real #278 bridge/session path without exposing a callable conversion from caller-selected bridge/session state into effect-eligible provenance. If ordinary host code and the #249 responder/mint path remain in the same freely importable Python authority domain, private names, constructor-shape flags and mutable internal fields cannot establish that provenance.

Smallest next result: split a bounded production-composition prerequisite. The effect consumer should receive only a result from that production composition boundary; no public/importable API may accept an arbitrary bridge/session or convert one into effect eligibility. If necessary, make the #278 assertion service a separate authority process/channel so the effect caller may request a fresh assertion but cannot answer or mint it. Add the exact uninjected-session + injected-state + self-answer + binder counterexample before claiming closure.

The typed GO_TO_2/no-retry transport, #279 permit, and causal #257 completion work remain useful. No motorized checkpoint is implicated.

djibian:
Draft authority finding d57629e9aef79e93e7709c99baf364a0e84ec4d4

The new “production-backed” flag still does not close the exact module-reachable mint counterexample.

`ReadOnlyCapabilitySession("radio://...")` with no injected constructor seams sets `_effect_preflight_production_backed=True` before any real connection is established. Both pieces the binder subsequently trusts are ordinary mutable Python attributes: the fixture itself sets `session._scf = OpenSyncSession(cf)` and `session._connection_epoch = ...` directly to manufacture an “active production session” without `session.open()`. Because the authoritative a53 NO_GO explicitly rejects leading-underscore naming as an access boundary, ordinary imported host code can perform the same writes.

The end-to-end counterexample therefore remains: create an uninjected `ReadOnlyCapabilitySession`, assign a fake open sync session and authorized epoch exactly as this candidate's test does, wrap it in a public self-answering `ReadOnlyCapabilityHttpBridge`, then call the still-module-reachable `_bind_effect_current_program_bridge()`. Its exact-type, production-backed and live-epoch checks all pass, yielding the exact handle accepted by `TrustedSetpointHlTransport` without a real #249 browser assertion or real live capability-session open.

Smallest repair boundary: production provenance cannot be inferred from a constructor-time boolean plus mutable session internals. The accepted capability must be rooted in an already-trusted production composition/lifecycle whose minting fact cannot be recreated by normal imported code through writable attributes or an exposed binder. The deterministic regression should reproduce the fixture's own direct `_scf`/`_connection_epoch` injection on a nominally uninjected session and prove that it still cannot reach an effect-eligible handle.

No motorized checkpoint is implicated.

djibian:
Draft authority finding d57629e9aef79e93e7709c99baf364a0e84ec4d4

The production-session marker narrows the cc298/a53 counterexample, but it still does not establish the required #249 browser provenance. `_bind_effect_current_program_bridge()` remains module-callable and it now asks only whether the bridge wraps an exact `ReadOnlyCapabilitySession` constructed without injection plus a live epoch. Those facts prove a real capability-session shape; they do not prove that the pending current-program challenge was answered by the production browser `assertCurrentProgram()` path.

The end-to-end bypass therefore still exists without an arbitrary session-like class: ordinary host code can construct/open an uninjected `ReadOnlyCapabilitySession` on a real live link, construct its own public `ReadOnlyCapabilityHttpBridge` with a responder credential it knows, self-answer the challenge with the already-known #267 profile/AST/epoch, call the still-publicly-reachable binder, and obtain the exact handle accepted by the transport. The deterministic fixture also shows why the construction-time boolean itself is not an authority boundary: it creates an uninjected exact session and then writes the live-state fields directly before binding.

Smallest repair is not another session-shape marker. The accepted effect provenance must carry something the effect caller cannot manufacture by self-answering: either the actual production composition must own the bridge/responder and hand the effect consumer an already-bound capability without exposing any binder/response secret to that consumer, or the #249 browser assertion must contribute cryptographic/opaque evidence bound into the exact run authorization. Until that boundary exists, a live Crazyradio epoch plus known profile/AST is insufficient proof of current browser state.

This is deterministic architecture evidence only; no physical checkpoint is implicated.

djibian:
Dependency split after NO_GO d57629e9aef79e93e7709c99baf364a0e84ec4d4

Issue #281 now captures the production-composition prerequisite exposed by the repeated #249 provenance counterexamples.

#276 should remain Draft while #281 is unresolved. Its useful transport work remains preserved, but no additional same-process private-key / underscore-binder / constructor-shape wrapper should be treated as closing the authority boundary.

The dependency is technical, not plan-ordering: #276 cannot truthfully claim a trusted current-program effect precondition until the later effect consumer can request the integrated #278/#249 assertion without possessing any exposed path that can answer or mint that assertion from caller-selected bridge/session state.

No motorized checkpoint is implicated.

djibian (commented):
NO_GO 9f3363f06a2abe146bca0258398868ab74974e73

Independent review against current main@c5c3ab23fda657aef99e00bec44ece8a33387557 and #280's explicit acceptance counterexample.

Moving the mint key into a closure removes the module-level binder/key, but it still does not make `CurrentProgramEffectPreflightHandle` non-forgeable in the same Python authority domain. The exact class remains a module attribute, the transport accepts it by exact type, and its only security-relevant state is a reachable Python slot containing the bridge.

This candidate's own deterministic fixture demonstrates the bypass directly:

```
handle = object.__new__(CurrentProgramEffectPreflightHandle)
object.__setattr__(
    handle,
    "_CurrentProgramEffectPreflightHandle__bridge",
    caller_selected_bridge,
)
```

That object is the exact type accepted by `TrustedSetpointHlTransport` and its `assert_current_program()` method delegates to the injected caller-selected/self-answerable bridge. Therefore the strongest #280 counterexample is not merely possible; the candidate uses it to exercise its positive tests.

A closure-private constructor key cannot authenticate an object whose accepted type and provenance-bearing state can be instantiated/populated without running `__init__`. More generally, same-process Python object encapsulation is not the authority boundary required here.

Smallest repair remains #280: move the positive #249 responder/mint authority outside the ordinary effect caller's importable object domain (or use an equivalent boundary whose positive evidence is independently verifiable). The effect caller may request a fresh exact assertion, but cannot manufacture its response by constructing/mutating Python objects. Do not make #276 Ready by adding another same-process wrapper.

The typed/no-retry GO_TO_2 transport and causal #257 completion work remain useful. No motorized checkpoint is implicated.

djibian (commented):
NO_GO 6836ef4e00675849be0c21e1d6423b4597ff3ae6

Independent review against current main@9d63bdf7247e5bbf2fb73b79d28a6314e5fe8326, integrated #283 / closed #280, and the authoritative #280 correction 5609326083 plus clarification 5614297255.

Merging current main into the candidate establishes the missing production trust root, but it does not yet compose this effect consumer into that root. The #276 delta still adds `CurrentProgramEffectPreflightHandle` to `serve_reference_capabilities.py`; `TrustedSetpointHlTransport` still accepts that obsolete same-interpreter handle; and the deterministic fixture still forges the accepted exact type with `object.__new__` plus direct bridge-slot injection. Meanwhile integrated `serve_physical_host.py` remains deliberately effect-free and never instantiates `TrustedSetpointHlTransport` inside `_run_physical_host()`.

After #283, the required repair is no longer to invent a stronger Python handle. The corrected trust topology makes the co-located #276 implementation part of the trusted physical-host TCB, while the external caller/UI is the adversarial side. Therefore current-program provenance for the effect must stay lexical to that host composition: the host-owned bridge performs the fresh exact #249 assertion immediately before the co-located effect, and caller IPC must not supply/select a bridge, responder credential, provenance client, mint/binder, or caller-returned evidence.

Smallest repair: remove the candidate-local `CurrentProgramEffectPreflightHandle` contract (restore the integrated read-only bridge surface), compose the transport from inside `serve_physical_host.py` using the host-local live bridge/session and same-epoch safety/effect objects, and update deterministic regressions so the strongest external caller bridge/socket substitution cannot influence the provenance used by the effect. Preserve the already-useful concrete #267/#266/#262/#257/#272/#271/#273/#279 bindings, typed #256/#268 GO_TO_2 surface, final fresh supervisor ordering, one plain no-retry send, and causal completion proof.

No motorized checkpoint is required for this repair.

djibian (commented):
NO_GO ac7733a763461ff18bd1da57b13d14648abc51f2

Independent review against current main@ec0c98a68744dd945d8ea53dc6261ab478e715a7, integrated #283/#280, and the current roadmap plus the still-applicable NO_GO 6836ef4e00675849be0c21e1d6423b4597ff3ae6.

This HEAD moves one piece into the correct TCB but still does not compose the claimed physical effect consumer. `serve_physical_host.py` now reflectively creates a `CurrentProgramEffectPreflightHandle` from its host-local bridge, but it never imports or instantiates `TrustedSetpointHlTransport`, never composes #267/#266/#262/#257/#260/#272/#271/#273 with the host-owned live session, and its own executable contract still states that it emits no flight/setpoint command. The caller IPC continues to perform only `validate-run-context` and returns non-authority diagnostic data. So the production path has fresh #249 provenance but no co-located first effect consumer.

The obsolete candidate-local handle layer also remains in `serve_reference_capabilities.py` and as the exact constructor dependency of `TrustedSetpointHlTransport`. Inside the corrected #283 trust topology Python privacy is not the security boundary; the important invariant is that the actual transport and provenance decision are composed lexically inside the physical-host TCB and cannot be selected or supplied by the external caller. Creating an otherwise-unused reflective handle in that host does not establish that composition.

Smallest repair boundary: remove the candidate-local `CurrentProgramEffectPreflightHandle` production contract, make the transport's fresh-current-program operation consume the host-local #278 bridge only through the co-located host composition, and instantiate the bounded #276 transport inside `_run_physical_host()` with the exact same live Crazyflie/session/epoch safety objects. The external caller may provide only bounded semantic/run requests; it must not provide a bridge, responder credential, provenance object/client or effect authority. Add deterministic host-composition coverage proving the actual effect transport uses the host-owned bridge and that caller-side bridge/socket substitution cannot influence the pre-effect assertion. Preserve the existing typed #256/#268 GO_TO_2 semantics, final #267/#262/#257/#272 ordering, #271 one-send/no-retry behavior, #273/#279 effect lifecycle and causal completion proof.

No motorized checkpoint is required for this repair.

djibian (commented):
NO_GO 2f075dd7fbb91a049f2543e016201b78ce9f30ee

Independent review against current main@ec0c98a68744dd945d8ea53dc6261ab478e715a7, the integrated #283/#280 trust boundary now recorded in current ROADMAP, and the still-applicable ac7733/6836 refutations.

The only delta after ac7733 is the merge of the documentation-only #284 result; it does not change the #276 implementation. The candidate still exports the obsolete `CurrentProgramEffectPreflightHandle` from `serve_reference_capabilities.py`, `TrustedSetpointHlTransport` still requires that exact handle, and `serve_physical_host.py` only reflectively injects its host-local bridge into an otherwise-unused handle. The production host still never instantiates `TrustedSetpointHlTransport` or composes the exact #267/#266/#262/#257/#260/#272/#271/#273 safety/effect objects around its one live Crazyflie/session.

Therefore the current roadmap requirement remains unmet: the actual first physical effect consumer must be co-located inside the #283 trusted physical-host TCB, perform fresh #278/#249 assertion there immediately before an effect, and accept no caller-returned bridge/provenance/effect authority. The external caller must remain bounded semantic/run IPC only.

Smallest repair remains to remove the candidate-local effect-preflight handle contract, keep `serve_reference_capabilities.py` on its integrated read-only surface, and compose the bounded transport inside `_run_physical_host()` from the host-owned live bridge/session and same-epoch safety objects. Deterministic host-composition evidence must show caller-side bridge/socket substitution cannot influence the provenance used by the actual effect path. Preserve the typed #256/#268 GO_TO_2 surface, final #267/#262/#257/#272 ordering, #271 one-send/no-retry behavior, #273/#279 lifecycle and causal completion proof.

No motorized checkpoint is required.

djibian (commented):
NO_GO 2f075dd7fbb91a049f2543e016201b78ce9f30ee

Independent review against current main@ec0c98a68744dd945d8ea53dc6261ab478e715a7, the newly integrated roadmap boundary, and the exact prior refutation on ac7733a763461ff18bd1da57b13d14648abc51f2.

This HEAD only merges current main's roadmap update into the Draft candidate. It does not change the physical composition code, so the ac7733 refutation remains applicable unchanged: `serve_physical_host.py` still creates only an otherwise-unused reflective `CurrentProgramEffectPreflightHandle`; it still does not instantiate `TrustedSetpointHlTransport` with the host-owned live Crazyflie/session and same-epoch #267/#266/#262/#257/#260/#272/#271/#273 state; and the candidate-local handle remains the transport's production current-program dependency.

The current roadmap now makes the repair boundary explicit: #280 is integrated and no longer the blocker; #276 itself must be the first co-located effect consumer inside the #283 trusted physical-host process, perform fresh #278/#249 assertion internally immediately before the effect, and accept no caller-returned provenance. The merge commit does not establish that claim.

Smallest repair remains: remove the candidate-local effect-preflight handle contract, bind the transport's fresh assertion directly to the host-local #278 bridge only through `_run_physical_host()` composition, and actually compose the bounded transport there with the exact live-session safety/authority objects. Preserve the already-useful typed #256/#268 GO_TO_2, final #267/#262/#257/#272 ordering, #271 one-send/no-retry behavior and #273/#279 completion lifecycle. No motorized checkpoint is required.

djibian:
Composition-timing refinement for the current `main@ec0c98a68744dd945d8ea53dc6261ab478e715a7` trust topology and Draft HEAD `2f075dd7fbb91a049f2543e016201b78ce9f30ee`.

The required co-location inside `serve_physical_host.py` must **not** be implemented by eagerly fabricating a fully active `TrustedSetpointHlTransport` at host startup. Two integrated prerequisites are intentionally run-scoped:

- #266 mints `EstablishedPoweredSession` only after an explicit fresh `flight is known inactive` gate, prior-evidence invalidation, STM+deck power-cycle, a new live epoch, fresh capabilities/preflight and fresh non-fault #257 state. Host startup or an ordinary caller request is not evidence that this reset is safe.
- #267 mints `TeacherRunAuthorization` only from one explicit trusted-host teacher decision for the exact profile/AST/epoch. The ordinary caller/UI channel established by #283 is explicitly not that decision source.

So the smallest correct production composition is **lazy and host-local**: `_run_physical_host()` owns/captures the real #278 bridge/session; once a genuine run activation has separately established the exact #266/#267/#262/#257/#260/#272/#271/#273 objects on that same live epoch, a lexical host-only composition operation constructs the #276 transport while injecting the already-captured host provenance source. No bridge/provenance object, teacher decision, powered-session authority or reset-safety assertion may be supplied by the ordinary caller IPC merely to satisfy the constructor.

This is implementation ordering, not a new product dependency: #266 and #267 are already integrated prerequisites. Deterministic repair evidence should therefore prove both sides: (1) the host-local composition cannot be reached without genuine run-scoped authorities, and (2) when those authorities are supplied by trusted host fixtures, the resulting transport's fresh #249 assertion is rooted only in the host-owned bridge and cannot be substituted from caller IPC.

No real reset, motorized test or `TEST_REQUIRED` checkpoint is needed for this deterministic repair.

djibian (commented):
NO_GO 04c90771f4930996fa642fb4da15a20a6b5b9550

Independent falsification review against current main@ec0c98a68744dd945d8ea53dc6261ab478e715a7, integrated #283/#280, current ROADMAP, and the still-applicable production-composition boundary from the 6836/ac7733/2f075dd refutations.

This HEAD correctly removes the forgeable `CurrentProgramEffectPreflightHandle` from the bridge and transport constructor. The base transport now fails closed without host-local provenance, and `_HostBoundSetpointHlTransport` closes directly over the lexical #278 bridge. Those repairs resolve the same-process handle/binder counterexample.

The remaining integration-blocking gap is that production still does not actually compose or expose the claimed first effect consumer. `_run_physical_host()` defines `_compose_inflight_setpoint_transport(...)`, assigns it to `compose_inflight_setpoint_transport`, and only checks that it is callable. No production path establishes the exact #266 powered session / #267 teacher receipt / #262 watchdog / #257 supervisor / #272 SafeLink / #271 ack / #273 execution objects and calls that closure; no `TrustedSetpointHlTransport` instance is created by the running host. The only caller operation remains `validate-run-context`, which returns non-authority diagnostic data and never reaches a bounded motion request.

The deterministic test mirrors this missing link rather than testing it: `TestHostBoundTransport` is a test-only analogue instantiated directly with fake authorities, while `test_production_source_roots_provenance_only_in_host_tcb()` only inspects source strings and import leakage. Thus it proves the core mechanics and lexical shape, but not that the actual production host can consume a bounded semantic request through the exact live-session safety lifecycle and reach exactly one SETPOINT_HL send.

Smallest repair: complete one deterministic production-shaped host path that constructs the exact same-session safety/authority objects, invokes the host-local composition closure, and routes one bounded already-flying horizontal/turn semantic request to that composed transport. Keep startup and mere `validate-run-context` effect-free; do not make teacher/reset authority caller-selectable. The regression should execute that actual host composition with injected/fake hardware seams or an equivalent deterministic TCB fixture, proving fresh #249 provenance immediately before the effect and exactly one no-retry send, while caller-side bridge/provenance substitution remains irrelevant.

No motorized checkpoint is required.

djibian (commented):
NO_GO 1425a2cd89a991c5ef7ca4a4676aeaf74494b244

Independent falsification review against current main@ec0c98a68744dd945d8ea53dc6261ab478e715a7, current ROADMAP, and the still-applicable production-composition refutation recorded on 04c90771f4930996fa642fb4da15a20a6b5b9550.

The current HEAD changes only the physical-host provenance contract wording after 04c907; it does not repair that refutation. The candidate now has the correct lexical #278/#249 provenance shape and removes the obsolete forgeable effect-preflight handle, but `_run_physical_host()` still only defines `_compose_inflight_setpoint_transport(...)`, assigns it to a local name, and checks that it is callable. No production path establishes the exact run-scoped #266 powered session, #267 teacher authorization, #262 watchdog, #257 supervisor reader, #272 SafeLink, #271 acknowledgement domain and #273 execution state and then invokes that closure. Consequently the running host still creates no `TrustedSetpointHlTransport` instance and its only ordinary caller operation remains effect-free `validate-run-context`.

The deterministic regression likewise instantiates a test-only `TestHostBoundTransport` directly with fake authorities and source-inspects the production host; it never executes a production-shaped host composition path. So it proves the transport core and provenance topology, but not the bounded claim that the first co-located physical effect consumer actually exists in the #283 host process.

Smallest repair boundary: keep startup and `validate-run-context` effect-free, but add one deterministic production-shaped trusted-host run path that establishes/receives the genuine host-side run-scoped authorities, invokes the lexical composition closure, and routes a bounded already-flying horizontal/turn semantic request through the composed transport. Teacher/reset authority must remain host-side and not caller-selectable. Exercise that actual path with deterministic fake hardware/seams, proving fresh #249 provenance immediately before exactly one no-retry SETPOINT_HL send and proving caller-side bridge/provenance substitution cannot affect it.

No motorized checkpoint is required.

djibian (commented):
NO_GO 1425a2cd89a991c5ef7ca4a4676aeaf74494b244

Independent falsification review against current main@ec0c98a68744dd945d8ea53dc6261ab478e715a7, current ROADMAP, and the still-applicable production-composition refutation on 04c90771f4930996fa642fb4da15a20a6b5b9550.

This HEAD is only a four-line docstring/source-oracle wrapping change in `tools/physical/serve_physical_host.py`; it does not change the physical composition path. The 04c907 blocker therefore remains unresolved on this exact Ready candidate.

`_run_physical_host()` still only defines `_compose_inflight_setpoint_transport(...)`, assigns/checks the closure, and never calls it with an established exact #266 powered session / #267 teacher receipt / #262 watchdog / #257 supervisor / #272 SafeLink / #271 ack / #273 execution domain. The running host still creates no `TrustedSetpointHlTransport` instance, and ordinary caller IPC still supports only `validate-run-context`; there is no bounded horizontal/turn semantic request routed through the co-located transport.

The deterministic test still uses test-only `TestHostBoundTransport` and source-string checks rather than executing the actual host composition, so it cannot establish the missing production reachability or prove the actual effect path uses the host-owned lexical #278/#249 provenance immediately before the one no-retry SETPOINT_HL send.

Smallest repair remains the prior one: complete one deterministic production-shaped host path that constructs the exact same-session safety/authority objects, invokes the lexical composition closure, and routes one bounded already-flying horizontal/turn request to the actual composed transport. Startup and `validate-run-context` must stay effect-free; teacher/reset authority must remain outside caller selection; deterministic coverage should execute this actual host path with fake/injected hardware seams (or equivalent TCB fixture) and prove caller-side bridge/provenance substitution cannot affect the fresh pre-effect assertion.

No motorized checkpoint is required.

djibian (commented):
NO_GO c4d983ad7b2835458616a74d24b36fe434bbd597

Independent falsification review against current main@ec0c98a68744dd945d8ea53dc6261ab478e715a7, current ROADMAP, and the still-applicable production-reachability finding from 04c90771f4930996fa642fb4da15a20a6b5b9550 / 1425a2cd89a991c5ef7ca4a4676aeaf74494b244.

The candidate now defines a host-local `_ActiveInflightRun` and `_activate_inflight_run(...)`, but the running production host still never calls that activation path. `_run_physical_host()` starts the capability bridge, publishes browser bootstrap, then services only ordinary `validate-run-context` IPC. It does not establish/receive the exact run-scoped #266 powered session, #267 teacher authorization, #262 watchdog, #257 supervisor reader, #272 SafeLink, #271 acknowledgement domain and #273 execution domain and pass them through `_activate_inflight_run`; consequently no `TrustedSetpointHlTransport` instance is created by the actual running host and no bounded horizontal/turn request can reach the effect path in production.

The new deterministic test does not close this gap. `production_activation()` parses `serve_physical_host.py`, extracts the nested production definitions with `ast.walk`, recompiles them into a synthetic wrapper, and invokes that wrapper with test-fixture authorities. That proves the extracted definitions can compose when externally supplied with fixtures, but it bypasses the production `_run_physical_host()` control flow whose reachability was the prior blocker. A source-extracted function is not evidence that the actual host establishes the authority lifecycle and reaches it.

Smallest repair boundary: make one real trusted production control-flow path inside `_run_physical_host()` establish or consume the genuine host-side run-scoped authorities and invoke `_activate_inflight_run`, then route one bounded already-flying horizontal/turn semantic request through that activated surface without making teacher/reset/provenance authority caller-selectable. The deterministic regression should exercise that actual production control flow through explicit deterministic seams or an equivalent injectable trusted-host runner, rather than AST-extracting dead definitions, and prove fresh #278/#249 provenance immediately before exactly one no-retry SETPOINT_HL send. Preserve startup and mere `validate-run-context` as effect-free.

No motorized checkpoint is required.

djibian:
Production-reachability refinement on exact Ready HEAD `c4d983ad7b2835458616a74d24b36fe434bbd597`.

Inspection of the integrated #273 execution lifecycle exposes a stricter prerequisite behind the current NO_GO. `PhysicalExecutionDomain` starts `recovery-required`; successful #266 reset establishment moves it only to `inactive`; and the only legitimate transition to `flying` is a positively acknowledged physical effect followed by an exact completion permit and fresh completion proof. Meanwhile #276's current `TrustedSetpointHlTransport` deliberately rejects ordinary GO_TO_2 unless the shared domain is already `flying`, and the candidate deliberately exposes no TAKEOFF_2 effect.

Therefore the current production host cannot legitimately make its new `_activate_inflight_run()` usable merely by wiring another request path to it: there is no existing production path that can establish the required `flying` phase first. The deterministic fixtures can seed `flying` by synthesizing an accepted effect, but doing that in production would fabricate exactly the #273 causal state that the domain exists to protect.

This narrows the repair decision. Before #276 can claim a genuinely reachable production in-flight effect path, one of two things must be established durably: either (a) a trusted, causally proven takeoff/initial-flight effect path that advances #273 from `inactive` to `flying`, or (b) an independently justified production mechanism that can adopt an already-flying vehicle into #273 without fabricating effect/completion evidence. No such mechanism is currently integrated. Merely adding a caller/trusted IPC operation around `_activate_inflight_run` is insufficient.

Do not weaken the `FLYING` precondition or mint a synthetic completion permit to close this gap. This is a deterministic architecture finding only; it does not authorize a motorized checkpoint.

djibian (commented):
NO_GO 2c984d0f28763912333e4877fe014da857928dc8

Independent falsification review against current main@b4633d92f173aef168102da065c1cecff233aa7d, including integrated #288, current ROADMAP, and the still-applicable production-reachability finding on c4d983ad7b2835458616a74d24b36fe434bbd597.

Merging #288 supplies the previously missing distinct trusted teacher-decision source and keeps the resulting exact #267 receipt process-local. It does not by itself close #276's remaining reachability gap. On this HEAD, `_run_physical_host()` still defines `_activate_inflight_run(...)` and `_ActiveInflightRun`, but no actual production control-flow path calls `_activate_inflight_run`. The teacher worker stores `teacher_state["authorization"]`; shutdown later closes that receipt, but normal host execution never consumes it to compose #266/#262/#257/#272/#271/#273 state into `TrustedSetpointHlTransport`, and ordinary IPC still accepts only `validate-run-context`. Therefore the running host still cannot reach one bounded horizontal/turn SETPOINT_HL effect through the claimed first physical effect consumer.

The new `test_physical_host_effect_composition.py` still proves only extracted-definition behavior: `production_activation()` parses the host AST, recompiles the nested classes/functions into a synthetic wrapper, and invokes that wrapper with test-fixture authorities. That does not execute the real `_run_physical_host()` control flow or prove that its live teacher receipt and same-session lifecycle objects actually reach activation.

Smallest repair boundary: use integrated #288 as the real teacher source, then complete one actual trusted host control-flow path that consumes the process-local exact receipt and genuine same-session #266/#262/#257/#272/#271/#273 objects, opens #260 as required, invokes `_activate_inflight_run`, and routes one bounded semantic horizontal/turn request through it. Keep startup and mere `validate-run-context` effect-free, and keep reset/teacher/provenance authority off caller selection. Replace the AST-extracted positive regression with a deterministic seam that executes the actual runner/control flow and proves fresh #278/#249 immediately before exactly one no-retry send.

No motorized checkpoint is required.

djibian:
DRAFT_FINDING — current #293 integration resolves the old `flying` reachability prerequisite, but exposes an exact-program sequencing gap on `2c984d0f28763912333e4877fe014da857928dc8` against `main@84c557993c19d84106a14e4715351924a139d519`.

The integrated #293/#287 path now legitimately advances the shared #273 domain from reset-established `inactive` to causally completed `flying` while retaining the exact #267 teacher binding and same-session #257/#262/#272/#271 authorities. That makes the current branch's bounded GO_TO_2 mechanics reusable rather than unreachable.

However, the branch's positive surface still accepts a caller-selected semantic `turn` or `horizontal` request and validates only that it is bounded plus that the surrounding teacher-authorized/current #249 AST binding is unchanged. The now-integrated #287 contract explicitly established that later in-flight execution needs exact-program sequencing semantics and must not infer that an arbitrary bounded motion is authorized merely because the containing AST is current. A caller can therefore request a different bounded move/turn from the next statement in the exact teacher-authorized canonical AST and still reach the effect core.

Smallest repair boundary: rebase/reconcile the useful one-shot SETPOINT_HL core onto current #293 host composition, then derive/consume the next eligible in-flight statement from the exact canonical AST binding (with explicit trusted sequencing state) rather than accepting caller-selected motion as effect authority. Preserve the current #256/#260/#268 geometry/timing, fresh #278/#249 + #267/#266/#262/#257/#272/#271 checks, one-send acknowledgement and private #279 completion semantics. Deterministic actual-host evidence should prove a caller-substituted bounded motion cannot change the next authorized effect and that successful takeoff hands the same exact run/session/epoch into this sequenced in-flight consumer.

No physical test or motorized checkpoint is required for this repair.

djibian:
DRAFT_FINDING — the exact-program cursor is now present on `01cf873c9134a68a31a59896686ef02335cc0dee`, but it is not yet connected to the production #293 host lifecycle.

`tools/physical/inflight_program_sequence.py` correctly removes caller-selected move/turn semantics from the cursor itself, and its focused test proves the next direct move/turn is derived from the #267 binding. However, the current PR diff still does not modify `physical_run_activation.py` or `serve_physical_host.py`: successful #293 activation returns the causal takeoff controller, the host stores it only for teardown, and no production path constructs this sequence from that controller's exact `active_run` bundle. As a result the new sequence currently has no production-reachable effect path, and there is no deterministic witness that the causally completed takeoff hands the same CF/session/epoch/#267/#262/#257/#272/#271/#273 authorities into #276.

The new regression is also not part of `run_runtime_v2_core_harness.py`; on this Draft HEAD the Runtime suite is skipped and only `CI Gate (Draft)` ran, so the green Draft workflow does not exercise either the existing SETPOINT_HL transport regression or the new sequencing regression.

Smallest repair before Ready: compose the sequencer inside the existing trusted `activate_validated_run()` result after successful causal takeoff, using a host-local SETPOINT_HL transport rooted in the same lexical #278/#249 bridge and the exact `ProductionTakeoffRunController.active_run` authorities; expose only a no-semantic `execute_next...()` trusted-process operation, never caller-selected direction/distance/angle. Add an actual-host deterministic regression showing takeoff -> same exact run/session/rotated epoch -> fresh current-program assertion -> exact next AST motion, and wire both #276 regressions into Runtime CI. No physical test or motorized checkpoint is required.

djibian:
DRAFT_FINDING — two production-control-flow edges remain on exact head `679d0eafa7a72264341e2179128df294246534a2` after the useful exact-sequence integration.

1. `execute-next-inflight` can still fail spuriously immediately after a successful `validate-run-context`. Validation sets `staged_ready` and then releases `lifecycle_lock`; if the caller thread reacquires that lock for the next request before the activation worker has acquired it, `activation_state["active_run"]` is still `None` and the exact request is rejected as unavailable. The later activation may then complete successfully, so the caller-visible failure is a scheduler race rather than a product precondition. The host should wait for the already-started bounded activation transaction before deciding whether the in-flight surface exists, while preserving caller-only/no-teacher effect-free behavior.

2. `_ActivatedRunController.execute_next_inflight()` unconditionally calls `_ensure_yaw_reader()` before `ExactInflightProgramSequence` decides whether the next authorized statement is a move or a turn. That makes an ordinary turn depend on opening the #260 yaw stream even though the established #276 boundary requires #260 only for horizontal geometry and says turns use #256/#268 directly. A yaw-stream failure can therefore reject an otherwise eligible exact turn. The sequence should expose/read its next exact kind without advancing, and the controller should open #260 lazily only for `move`.

Both repairs are deterministic and preserve the current authority topology, exact-AST sequencing, and no-motorized-test boundary.

djibian:
Draft finding bc5a456470d8fe8d5b960cd3dcd18d87527b0156

The production host/activation composition now closes the earlier #249 authority bypasses and the activation handoff/lazy-yaw path is materially stronger, but the branch still carries a second, weaker AST cursor in `tools/physical/inflight_program_sequence.py` instead of the exact fail-closed sequencing boundary being established by #295.

On this exact Draft HEAD, `ExactInflightProgramSequence.__init__()` accepts the terminal statement whenever `program[-1].get("kind") == "land"`. It does not require the exact terminal form `{kind: "land"}`. Therefore a canonical teacher-bound program ending in e.g. `{"kind":"land","extra":true}` is silently treated as the landing boundary and never inspected by `execute_next()`. The already-causal takeoff path proves the first statement, but it does not validate that terminal landing object. This recreates the terminal-boundary counterexample already found and repaired in #295.

Smallest repair boundary: do not maintain a parallel sequencing authority here. Once #295 is integrated, reconcile #276 onto current `main` and reuse its exact canonical sequencing primitive (or prove byte-for-byte equivalent terminal/caller-nonselection semantics), deleting the duplicate weaker cursor if possible. Add/retain the malformed-terminal regression through the actual production host composition so an unsupported terminal statement cannot be skipped as a landing boundary.

This is deterministic host-side evidence only; no motorized checkpoint is implicated.

djibian:
DRAFT_FINDING — exact head `bc5a456470d8fe8d5b960cd3dcd18d87527b0156` repairs both prior production-control-flow findings, but the candidate-local `ExactInflightProgramSequence` still has one exact-program gap that should not survive reconciliation with #295.

Its constructor accepts the terminal statement whenever it is a dict whose `kind` is `land`; it does not require the exact supported landing form. The physical preflight contract canonicalizes JSON and derives capability facts, but it does not reject extra fields on action statements. Therefore a bound canonical program ending in e.g. `{"kind":"land","extra":true}` can be silently treated as the final landing boundary by this sequencer. The same class also re-parses/owns cursor semantics independently of the stronger no-effect reservation/ambiguity state machine already prepared in #295.

#295's current candidate explicitly closes this counterexample by requiring terminal `{kind: land}` exactly and by separating reserve/complete/release/ambiguous outcomes. Once #295 is integrated, the smallest repair for #276 is to consume that shared exact-program sequencing primitive rather than keep a second weaker parser/cursor. If #276 must remain independently testable before that integration, at minimum add the exact terminal-land rejection regression now and preserve the current no-caller-semantics boundary.

The validate→activation race and turn-only yaw dependency identified on `679d0eafa7a72264341e2179128df294246534a2` are resolved on this HEAD: the started activation is awaited outside `lifecycle_lock`, and #260 yaw is opened only when the next exact kind is `move`. No physical checkpoint is implicated.

djibian (commented):
GO 1da5355829312ba7525dff03a84652d1071a3c9d

Independent falsification review against current main@fb0ce1b5d139c7b343dc1e059b7084140af9bb6e, the integrated #295 sequencing primitive, and every still-applicable owner-authoritative #276 refutation/findings. This execution did not modify #276.

The original authority-by-callback/raw-command findings are resolved. The importable `TrustedSetpointHlTransport` now binds exact concrete #267/#266/#262/#257/#272/#271/#273 objects on one epoch, exposes only typed already-flying horizontal/turn methods whose geometry/timing comes from #256/#260/#268, and has no bridge/provenance constructor input or generic/raw command API. Its base current-program hook fails closed, so direct imported construction cannot establish the positive provenance path.

The #278/#249 provenance and production-reachability refutations are also resolved. Actual `serve_physical_host.py` owns the one live session/bridge, stages only a host-validated binding, and its trusted activation worker calls `activate_validated_run(...)` with that lexical bridge and distinct teacher channel. The returned process-local controller owns the co-located host subclass whose fresh current-program hook re-asserts the exact #267 profile/AST/rotated epoch through that host bridge. Ordinary IPC can request only parameter-free `execute-next-inflight`; extra caller motion/provenance fields are rejected. The actual runner path, not an AST-extracted synthetic wrapper, delegates that operation to the activated controller.

The later exact-program findings are resolved by consuming integrated `PhysicalProgramSequence` rather than keeping the weaker duplicate cursor. The complete canonical AST is validated before reset/takeoff, including exact terminal `{kind: land}`; the actual-host regression supplies the extra-field landing counterexample and proves no takeoff/effect occurs. After causal takeoff, reserve/complete/release/ambiguous state derives only the next exact move/turn from the immutable teacher-bound AST. Caller substitution cannot alter that order. The validate→activation scheduler race is fenced by awaiting the already-started activation before reacquiring the lifecycle lock, and #260 yaw is opened lazily only for horizontal moves, not turns.

The one-send transport still installs the acknowledgement callback before exactly one plain `send_packet(packet)` with no application retry, rechecks current binding/teacher/watchdog around fresh #257 supervisor evidence and live #272 SafeLink, and consumes the private #279 completion permit only with causal completion back to `FLYING`; uncertain emitted/completion state poisons sequencing instead of manufacturing a retry. The bounded slice still excludes landing, vertical motion, STOP, trajectories/spirals and motorized qualification.

I found no remaining applicable counterexample on this exact Ready HEAD. Fresh Ready CI is still a separate integration prerequisite; the prior Draft run cannot satisfy the required `CI Gate` decision context.

djibian (commented):
GO b44f1da22241f818da29b231202a579947ba2338

Independent exact-HEAD review against current main@4603b9809cc9e14486248101accd3bb1cf12bf1b and every still-applicable owner-authoritative #276 finding. This execution did not modify #276.

The only delta from the independently reviewed GO candidate 1da5355829312ba7525dff03a84652d1071a3c9d is the merge of current main's #299 result: three added X3 props-off acquisition-support files under `experiments/crazyflie-ukf-surface-range/`. They do not overlap any of #276's nine physical-host/transport/test files, and current `AGENTS.md`, `docs/PRODUCT_VISION.md`, and `docs/ROADMAP.md` are blob-identical to the contract/vision/roadmap reviewed for the prior candidate. I found no new interaction that reopens an earlier authority, provenance, production-reachability, sequencing, one-send/acknowledgement, completion, or caller-substitution counterexample.

The bounded #276 surface therefore remains as previously reviewed: host-owned fresh #278/#249 provenance, exact teacher-bound program sequencing after causal takeoff, parameter-free external in-flight execution request, typed horizontal/yaw SETPOINT_HL only, one no-retry send with fresh authority/supervisor/SafeLink checks, fail-closed ambiguous sequencing, and no landing/vertical/STOP/trajectory/spiral or motorized checkpoint. Fresh Ready `CI Gate` on this exact HEAD remains a separate integration prerequisite.

djibian (commented):
GO b44f1da22241f818da29b231202a579947ba2338

Independent falsification review against current main@4603b9809cc9e14486248101accd3bb1cf12bf1b, the prior exact GO on 1da5355829312ba7525dff03a84652d1071a3c9d, and every still-applicable owner-authoritative #276 finding. This execution did not modify #276.

The only candidate change after the previously reviewed 1da5355 physical-effect implementation is integration of current main's already-reviewed #299 result: three isolated X3 laboratory acquisition files under `experiments/crazyflie-ukf-surface-range/`. The #276 physical host, activation, transport, exact-program sequencing, safety/authority composition and deterministic effect tests are byte-for-byte unchanged by that merge-only refresh, and #299 introduces no dependency or alternate path into the physical-host trust domain.

Accordingly the prior closure of the callback/raw-command, current-program provenance, production reachability and exact-program sequencing refutations remains applicable on this exact Ready HEAD. The effect surface still accepts only parameter-free next-step IPC, derives move/turn semantics from the teacher-bound canonical AST, performs fresh host-local #278/#249 assertion inside the effect transaction, rechecks concrete #267/#266/#262/#257/#272/#271/#273 authority/lifecycle state, emits exactly one no-retry SETPOINT_HL packet, and uses the private #279 completion permit with causal fresh supervisor evidence. Landing, vertical motion, STOP, trajectories/spirals and motorized qualification remain outside this slice.

I found no new counterexample introduced by the current-main merge and no unresolved applicable refutation on b44f1da22241f818da29b231202a579947ba2338. Fresh exact-HEAD Ready CI remains a separate integration prerequisite.

PR #279, opened 2026-09-09
Bind physical completion to exact accepted effect

djibian:
Addresses the completion-boundary refutation recorded on #276 without adding any physical command surface.

Current #273 allows `complete_accepted_effect(next_phase, prove_completion)` to be called by any holder of a `PhysicalExecutionDomain`; the existing deterministic fixture demonstrates that `lambda: True` can manufacture completion after a positive acknowledgement. Once #276 adds ordinary GO_TO_2 emission, that would let a caller restore `FLYING` and overlap a still-running high-level trajectory.

This bounded prerequisite binds completion to the exact accepted effect:
- `mark_accepted()` mints and returns one opaque `AcceptedEffectCompletionPermit`;
- the process-wide execution state stores that exact object as the sole pending completion capability;
- `complete_accepted_effect(...)` now requires identity with that exact pending permit before any completion proof runs;
- the permit is consumed before the potentially blocking proof;
- reset, rejection, unresolved effects and neutral close paths clear any stale permit;
- a wrong/forged or consumed permit cannot alter the pending accepted effect.

The permit is deliberately **not** completion evidence by itself. The future #276 transport must keep the exact permit private and pair it with concrete fresh #257 trajectory-completion evidence (or #264 for landing). This PR only removes the cross-caller assertion-shaped bypass in #273.

Deterministic regressions prove that:
- `object() + lambda: True` cannot restore motion eligibility;
- the real pending effect remains `AWAITING_COMPLETION` after a forged permit attempt;
- a consumed permit cannot complete a later accepted effect;
- negative/ambiguous proof still forces recovery;
- existing reset/effect exclusion remains unchanged.

No cflib/CRTP/browser effect surface and no motorized checkpoint are introduced.

Refs #157. Prerequisite for #276.

### DIFF
--- a/tools/ci/test_physical_execution_domain.py
+++ b/tools/ci/test_physical_execution_domain.py
@@ -147,8 +147,12 @@ def test_definitive_rejection_restores_exact_prior_phase() -> None:
 
     with execution.effect_transaction(lambda: None) as effect:
         effect.mark_emitted()
-        effect.mark_accepted()
-    execution.complete_accepted_effect(domain.FLYING, lambda: True)
+        takeoff_permit = effect.mark_accepted()
+    execution.complete_accepted_effect(
+        takeoff_permit,
+        domain.FLYING,
+        lambda: True,
+    )
     require(
         execution.phase == domain.FLYING,
         "accepted takeoff completion establishes flying",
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO c22080d2e5f3e46841b679cc064da00ecf4f2cb1

Independent review against current main@e3658f5620115de6c6017ac6656ec2882744b472 and the bounded completion-provenance finding recorded on #276.

This candidate correctly narrows #273 without pretending to prove trajectory completion itself. A positive acknowledgement now creates one exact one-shot `AcceptedEffectCompletionPermit`; process-wide state retains that exact object; completion requires identity with it; stale/forged/consumed permits cannot alter the pending effect; and the permit is consumed before the potentially blocking completion proof so negative or ambiguous proof still forces recovery. Reset/rejection/unresolved paths clear stale permit state.

The important scope boundary is explicit and preserved: the permit is not #257/#264 evidence, and `lambda: True` remains acceptable only to the trusted consumer that already holds the exact permit. #276 must therefore keep the permit private and pair it with concrete fresh trajectory/landing completion evidence before restoring ordinary motion eligibility. This PR does not by itself close that later integration obligation.

The diff adds no cflib/CRTP/browser command surface and preserves the existing process-wide reset/effect exclusion. Deterministic coverage exercises the exact prior cross-caller counterexample, stale permit replay and failure-to-recovery semantics. No unresolved counterexample was found for this bounded prerequisite.

Issue #280, opened 2026-09-09
Establish non-forgeable production current-program provenance


djibian:
Parent: #157. Blocks #276.

## Problem

#278 established the fresh host-initiated #249 profile/AST/connection-epoch round trip, but #276 has now demonstrated a missing trust boundary between that non-authority browser evidence and the first flight-capable effect consumer.

The exact counterexample is durable on #276 (latest: NO_GO d57629e9aef79e93e7709c99baf364a0e84ec4d4): ordinary imported Python code can create or select a bridge/session whose profile/AST/epoch values are already known, answer its own current-program challenge, and reach an effect-eligible wrapper/binder without the response having come from the actual production browser `assertCurrentProgram()` path. Constructor-shape booleans, leading-underscore names and mutable session fields do not establish provenance.

A real Crazyradio connection is not enough. It proves the live epoch, not that the browser workspace/profile was freshly re-asserted.

## Bounded objective

Establish the smallest production composition boundary that lets a trusted physical effect consumer request fresh exact #249 evidence while preventing that same effect-caller authority domain from manufacturing or self-answering the evidence.

Preserve:
- exact profile + backend-neutral AST + reconnect-sensitive epoch binding;
- host-initiated one-shot challenge/freshness;
- timeout/mismatch/reconnect/replay fail-closed behavior;
- `executionAuthority:false` for the browser/preflight channel;
- the distinct #267 teacher authorization boundary;
- no motorized effect in this prerequisite.

## Acceptance

A deterministic counterexample must start from the exact information available to the physical effect caller (authorized profile/AST/epoch and ordinary imported host APIs) and prove it cannot create accepted current-program evidence by:
1. constructing/selecting its own bridge/session;
2. self-answering a challenge;
3. calling any module-reachable constructor/binder/mint path;
4. mutating ordinary reachable Python object state used as provenance.

The positive path must demonstrate that the actual production #249 responder can still satisfy a fresh host request and that the resulting provenance is consumable by #276 without exposing the responder/mint authority to the effect caller.

If same-process object separation cannot satisfy this invariant, use the smallest separate authority process/channel or cryptographic/opaque browser-origin evidence necessary; do not simulate privacy with naming conventions.

## Non-goals

- no new student/browser effect API;
- no arming or flight command;
- no change to #256/#257/#260/#262/#266/#267/#268/#271/#272/#273 semantics;
- no physical checkpoint;
- no broad security framework or general IPC abstraction.

Refs #249 #278 #276.

[...]

PR #278, opened 2026-09-09
Bridge current-program preflight into trusted physical host

djibian:
Implements the smallest production non-authority handoff required by #277 / #276.

The trusted physical host creates one-shot current-program challenges bound to the exact expected profile ID, semantic AST binding and live connection epoch. The ordinary browser capability bearer remains strictly read-only and still exposes only `readCapabilities()` / `readConnectionEpoch()`.

A distinct host-generated `preflightResponderToken` authenticates a closure-private responder inside the production `physical_preflight_runtime.js`. That responder cannot be invoked through the ordinary capability adapter and derives every positive response by calling the already-integrated exact-current #249 `assertCurrentProgram()` path. The ordinary capability bearer receives 401 on both responder endpoints.

The host mints one immutable `CurrentProgramPreflightEvidence` only after:
- a live epoch check before challenge creation;
- one responder-authenticated claim of the host-created nonce;
- one matching non-authority response derived from the real #249 runtime assertion;
- a second live epoch check after the response.

Timeout, failed #249 assertion, profile/AST/epoch mismatch, replay, late/unknown response and reconnect fail closed. Evidence construction requires the host-only mint key and the minted profile/AST/epoch/challenge fields are frozen after construction. The responder capability remains non-authority and exposes no arming, setpoint, movement, light or other physical effect.

Deterministic regressions cover exact success, ordinary-bearer rejection on responder endpoints, production responder use of the real #249 path, AST/profile/epoch invalidation, one-shot replay/late response, host timeout, fabricated evidence and post-mint mutation attempts.

No motorized checkpoint is requested or required.

Refs #277. Parent #157. Blocks #276.

PR #282, opened 2026-09-09
Record current-program provenance prerequisite

djibian:
Records the newly demonstrated #280 prerequisite in the product roadmap before later physical-effect work relies on the older #278 boundary.

#278 remains established as the fresh host-initiated #249 profile/AST/epoch round trip. The new durable conclusion is narrower: #276 has falsified same-process Python wrappers as **effect-eligible provenance**, because caller-reachable bridge/session state can be constructed, mutated or self-answered inside the same authority domain.

This roadmap-only update therefore:
- preserves #278 as the non-authority browser handoff;
- records #280 as the technical production-composition prerequisite before the first physical effect consumer;
- requires responder/mint authority to remain outside the effect caller's importable authority domain, with only a fresh exact request/reply capability crossing that boundary;
- records #279 as the integrated exact accepted-effect completion permit;
- keeps #276 non-integrable until #280 is established;
- preserves the existing #267 teacher split and all #257/#260/#262/#266/#268/#271/#272/#273 safety semantics.

No implementation, governance, CI, browser effect API or physical checkpoint is changed.

Refs #157 #280 #276.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -247,11 +247,17 @@ justified by the current evidence.
   during or after preflight invalidates it; #246 provides the concrete host-side
   `ReadOnlyCapabilitySession` that opens one explicit Crazyradio link, creates/
   rotates that epoch around live connections, invalidates it on cflib disconnect
-  and reconstructs descriptors from current connected evidence; and #249 binds
-  the production physical submission path to that live session, preserving the
-  exact current activity profile and semantic workspace AST across preflight and
-  re-asserting them with the same connection epoch immediately before any later
-  separately authorized effect. Integrated #256 then codifies the physical
+  and reconstructs descriptors from current connected evidence; #249 binds the
+  production physical submission path to that live session, preserving the exact
+  current activity profile and semantic workspace AST across preflight; and
+  integrated #278 adds a fresh host-initiated one-shot profile/AST/epoch
+  re-assertion channel from the production browser path. #276 subsequently
+  falsified same-process wrappers around that channel as effect provenance:
+  ordinary effect-side Python can still construct/mutate/self-answer reachable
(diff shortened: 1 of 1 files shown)


djibian (commented):
GO 21dcbeb64cb822f1230f53c0e53db24bedbc00c3

Independent review against current main@c5c3ab23fda657aef99e00bec44ece8a33387557, issue #280, and the durable #276 provenance refutations.

This roadmap-only candidate records the smallest conclusion actually established by current evidence without rewriting the validated #278 result. It preserves #278 as the fresh non-authority host↔browser #249 round trip, records #280 as the additional effect-provenance prerequisite exposed by the same-process counterexamples, and explicitly prevents #276 from becoming an integration candidate until responder/mint authority is outside the effect caller's importable authority domain.

The update also correctly incorporates integrated #279 while preserving the existing #257/#260/#262/#266/#267/#268/#271/#272/#273 boundaries and leaves safe real execution unproven. It adds no effect API, governance change or physical checkpoint.

No applicable refutation remains for this exact documentation candidate.

djibian (commented):
GO 21dcbeb64cb822f1230f53c0e53db24bedbc00c3

Independent review against current main@c5c3ab23fda657aef99e00bec44ece8a33387557 and the authoritative #280/#276 provenance findings.

This roadmap-only candidate accurately records the newly demonstrated technical prerequisite without claiming that #278 itself failed: the fresh host-initiated #249 round trip remains established, while effect-eligible provenance now explicitly requires #280's production composition boundary outside the effect caller's importable bridge/session authority domain. It also correctly records #279 as integrated and keeps #276 non-integrable until #280 is established.

No governance, CI, browser effect API, command surface or physical checkpoint is changed. The exact Ready HEAD has canonical CI Gate success.

PR #288, opened 2026-09-10
Add trusted teacher decision channel to physical host

djibian:
Implements #286 as the smallest distinct trusted teacher-decision path inside the #283/#280 physical host.

The trusted launcher may install an optional, distinct `--teacher-fd`. The teacher channel is independent of ordinary caller IPC and browser bootstrap. The teacher side names one exact profile/AST/connection-epoch binding; the host creates a fresh correlation challenge; only one exact boolean decision is accepted. Positive #267 authority is minted only after the teacher write side reaches EOF, so duplicate or late frames fail closed before authority publication. The live connection epoch is checked before, during and after minting. Denial, EOF, malformed data, wrong correlation/binding, duplicate/late data, reconnect/epoch change or channel ambiguity is terminal.

`serve_physical_host.py` starts this one-shot trusted path independently. Ordinary caller IPC remains `validate-run-context` only and cannot trigger, select, replace or write teacher decisions. The resulting `TeacherRunAuthorization` remains process-local and is never serialized to caller or browser channels. No student flight-request workflow or physical-effect operation is added.

Deterministic coverage exercises exact positive binding, denial/EOF/malformed data, correlation and binding failures, duplicate/late frames, epoch changes, permanent #267 invalidation, separation from ordinary/browser channels, and absence of effect surfaces. The core harness reaches this regression through the existing teacher-authorization test.

No motorized checkpoint is requested.

Refs #286 #157. This is the trusted teacher-source prerequisite for completing #276 production activation.

PR #283, opened 2026-09-09
Root current-program provenance in physical host process

djibian:
Implements the bounded #280 production-composition prerequisite exposed by the repeated #276 provenance counterexamples.

The candidate follows the latest authoritative #280 architecture correction 5609326083:
- one trusted physical-host authority process owns the exact live Crazyflie/session, integrated #278 bridge/responder, and the future co-located #276 transport plus its identity-sensitive safety/effect domains;
- the ordinary caller/UI remains outside that process, owns no Crazyflie object, bridge/responder credential or effect transport, and sends only bounded untrusted `validate-run-context` semantic/run requests over launcher-installed IPC;
- browser bootstrap remains a distinct one-way trusted-composition channel;
- the host performs the fresh #249 assertion internally and keeps `CurrentProgramPreflightEvidence` local so future #276 can consume it immediately on the same live session/connection epoch before composing #267/#266/#262/#257/#272/#271/#273;
- caller-visible replies remain explicitly `executionAuthority:false` and are never effect provenance.

The applicable e58 reachability counterexample is addressed by lexical authority encapsulation: session/bridge/responder imports and live objects exist only inside `_run_physical_host()` locals, not as attributes on the importable `__main__` module or another production mint/binder surface. Importing `serve_physical_host.py` creates or exposes no live session, bridge, responder, client, binder or effect object.

Deterministic regressions preserve the real production-shaped browser positive path, reproduce caller-side public-bridge self-answer and forged local IPC attacks, fail closed when the browser responder is absent or the AST binding mismatches, and inspect the host composition to prove `bridge`/`session` are runner locals with no global/nonlocal publication and no `__main__` authority registry.

The caller-selectable provenance client and preflight-only broker remain removed. This prerequisite emits no movement/arming/setpoint/reset command and requests no physical checkpoint.

Refs #280 #157. Blocks #276 until independently validated and integrated.

PR #289, opened 2026-09-10
Add exact-bound causal takeoff effect core

### LINKED ISSUE #287: Establish causal physical takeoff transition into the execution domain
## Product role

Prerequisite for #157 / #276 discovered from the integrated #273 execution lifecycle.

Current #276 ordinary `GO_TO_2` motion correctly requires `PhysicalExecutionDomain.phase == flying`, but current production state cannot legitimately reach that phase: #273 starts `recovery-required`, successful #266 reset establishment yields only `inactive`, and the only authorized transition to `flying` is an actually emitted physical effect followed by its exact positive acknowledgement permit and fresh completion proof.

A deterministic fixture may seed `flying`; production must not. This issue establishes the smallest causal initial-flight transition so later already-flying movement can become genuinely reachable without weakening #273.

## Established student semantic boundary

- The student AST already requires top-level `takeoff` first and `land` last.
- V2 `takeoff.height_m` is globally bounded `0.2..1.5 m`; the representative progression profiles currently narrow it to `0.5..1.0 m`.
- The physical path must consume that exact already profile-validated absolute target height. It must not add a second pupil-facing takeoff vocabulary or reinterpret student horizontal `set_speed` as vertical speed.

### Exact AST binding is the semantic source, not caller motion data

`VERIFIED_BY_CODE_INSPECTION` on current `main@ec0c98a68744dd945d8ea53dc6261ab478e715a7`: `physical_capability_contract.bindAst(ast)` returns the canonical JSON serialization of the complete backend-neutral AST; it is not an opaque hash. The #249/#278 current-program assertion already binds that exact canonical string as `astBinding` together with the profile and connection epoch.

Consequently the trusted physical host already has a sufficient immutable semantic source for takeoff once the #267 run binding is established: parse the exact `teacher_authorization.binding.ast_binding`, require `{version:1, semantics:"webeeblocks-ast-v1"}`, require the first top-level program statement to be exactly a valid `takeoff`, reject extra/unknown takeoff fields, and derive `height_m` from that statement. The effect API must **not** accept a caller-supplied takeoff height and merely compare it to bounds. This also avoids creating a second serialization/hash algorithm in Python: use the exact canonical binding already proven by the production browser and treat it as read-only semantic input.

For #287, only the first top-level `takeoff` is consumed; no general physical AST interpreter is introduced by this slice. Later in-flight execution still needs its own exact-program sequencing semantics and must not infer that an arbitrary bounded caller motion is authorized merely because #249 re-asserted the surrounding AST binding.

## Pinned firmware command selection

`VERIFIED_BY_CODE_INSPECTION` against Crazyflie firmware `54f31e243a0b28b67efef5ba20dbb6d9890a5478` and cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873`.

The first draft of this issue assumed cflib's public `TAKEOFF_2` duration API. The pinned firmware exposes a smaller and safer deterministic product boundary: `COMMAND_TAKEOFF_WITH_VELOCITY` (command 9).

- Firmware defines the packed request as `groupMask, height, heightIsRelative, yaw, useCurrentYaw, velocity`.
- With `heightIsRelative=false`, the student target remains the existing absolute `takeoff.height_m` semantic.
- With `useCurrentYaw=true`, takeoff preserves the current yaw and does not introduce an unrequested heading change.
- Firmware computes `duration = abs(height - current planner Z) / velocity` from its own current planner position immediately before `plan_takeoff(...)`; the host therefore does not need to invent a ground-Z assumption or add a fresh-Z observer merely to derive duration.
- The pinned firmware labels its `defaultTakeoffVelocity = 0.5 m/s` setting as a safe default. The product request should send an explicit positive `0.5 m/s`, rather than zero/negative velocity, so execution does not silently inherit a mutable `hlCommander.vtoff` parameter changed outside WebeeBlocks.
- `COMMAND_TAKEOFF_WITH_VELOCITY` is part of the pinned high-level CRTP protocol and firmware C API even though current cflib `HighLevelCommander` does not expose a Python convenience method for it. A narrow typed packet builder is therefore required; no generic raw-packet API is justified.
- The normal SETPOINT_HL four-byte reply path is common to command 9, so integrated #271 acknowledgement freshness remains applicable.

Exact typed request target for the first implementation is therefore equivalent to little-endian `<BBf?f?f>`:

`command=9, groupMask=0, absoluteHeight, heightIsRelative=false, yaw=0.0 (ignored), useCurrentYaw=true, velocity=0.5`.

This source selection supersedes the earlier draft requirement to invent a separate host takeoff-duration policy for `TAKEOFF_2`.

## Required trust/safety composition

The production takeoff consumer must stay inside the #283 trusted physical-host TCB and, for the exact run/connection epoch:

1. start from #273 `inactive`, never by fabricating/adopting `flying` state;
2. consume the exact #266 established powered session and exact #267 teacher authorization;
3. require the #262 emergency-watchdog liveness guard already active and keep it live throughout takeoff/completion;
4. derive the takeoff statement/height from the exact canonical AST string in the #267 binding, not from caller-supplied motion parameters;
5. re-assert that exact current #278/#249 profile + canonical AST binding + connection epoch inside the effect transaction immediately before emission;
6. require fresh #257 pre-effect supervisor evidence with no blocking fault, `can_fly=true`, `is_flying=false`, no conflicting active high-level trajectory, and the exact expected same-session identity/epoch;
7. construct only the typed command-9 request above from the AST-derived absolute takeoff height;
8. re-check mutable teacher/powered/watchdog authority after request construction and immediately before #272/#271/emission;
9. require live #272 SafeLink, register the exact SETPOINT_HL reply callback first, and perform exactly one plain no-retry physical send under the process-wide #271 acknowledgement domain;
10. on positive acknowledgement, use the exact #279 completion permit and causal fresh #257 evidence to move #273 to `flying`; ambiguity, timeout, disconnect, malformed reply or uncertain completion must remain fail-closed/recovery-required; definitive firmware rejection may restore the prior stable `inactive` phase;
11. expose no raw packet API and accept no caller-supplied takeoff height, provenance, bridge, teacher/reset authority, powered-session evidence or watchdog authority.

## Causal takeoff completion

Completion must not depend on an elapsed host sleep or a pre-existing `hl_traj_finished=true` bit.

The exact pre-effect #257 baseline is `not flying`. After a positive command acknowledgement, completion authority requires a fresh same-epoch post-effect observation that establishes physical flight under high-level control (`is_flying=true`, `hl_control_active=true`, no blocking fault). From that causal flight state, the takeoff may complete when fresh #257 evidence reports the high-level trajectory finished while flight/high-level control remain valid. If the first fresh post-ack sample already reports `is_flying=true`, `hl_control_active=true`, and `hl_traj_finished=true`, the pre-effect `not flying` baseline plus process-wide #273 effect exclusion still supplies the causal transition; a stale finished bit by itself never does.

Watchdog liveness and exact teacher/powered authority remain rechecked while completion is pending. Any ambiguous observation leaves the accepted effect unresolved/recovery-required rather than manufacturing `flying`.

## Deterministic acceptance

Machine-only tests must prove at least:

- no transition/effect from `recovery-required`, already-`flying`, wrong epoch/session/CF, stale/missing teacher authorization, dead/ambiguous watchdog, unsafe/faulted/not-flight-capable supervisor state or lost SafeLink;
- malformed/non-canonical/wrong-version AST binding, missing/non-first takeoff, unknown takeoff fields, non-finite/out-of-profile height, or any caller-supplied alternate height cannot reach packet construction;
- the exact height used by command 9 is extracted from the exact #267 canonical AST binding that #278/#249 freshly re-asserts;
- exact command-9 packed structure, absolute height, current-yaw preservation and explicit `0.5 m/s` velocity;
- no fallback to mutable firmware `hlCommander.vtoff`, no coupling to student `set_speed`, and no arbitrary raw request bytes;
- callback-before-send, one plain send, no application retry;
- definitive rejection vs ambiguous result semantics;
- causal post-ack completion: a pre-takeoff `inactive/not-flying` observation or stale `hl_traj_finished` alone cannot advance #273;
- positive completion advances the real shared #273 domain from `inactive` to `flying`, after which the separately bounded #276 in-flight transport becomes eligible;
- the production-shaped trusted-host control flow, not a source-extracted/test-only analogue, exercises this transition with deterministic fake hardware seams.

## Scope boundary

This issue does not authorize arming, real takeoff, flight, a motorized checkpoint, landing, vertical student motion, STOP/e-stop semantics, or a generic raw SETPOINT_HL surface. Real execution proof remains a later explicit checkpoint after deterministic composition is complete.

Refs #157 #276 #273 #266 #267 #262 #257 #249 #278 #271 #272 #279 #283.

djibian:
Adds the smallest machine-only #287 prerequisite that is independently complete without claiming production host reachability.

This slice provides two bounded pieces on top of current `main`:

- exact command-9 takeoff semantics derived solely from the canonical backend-neutral AST string already bound by #267/#249, with no caller-supplied height or raw-packet surface;
- a trusted causal takeoff effect core that composes the existing #266/#262/#257/#272/#271/#273/#279 domains, requires fresh not-flying/can-fly supervisor evidence, performs one no-retry SETPOINT_HL send, distinguishes definitive rejection from ambiguity, and advances to `flying` only after causal fresh post-ack supervisor completion evidence.

The importable core deliberately has no positive current-program provenance path: its current-binding hook fails closed until later #283 physical-host composition supplies the lexical #278/#249 assertion. Deterministic tests cover packet shape, AST-derived height, authority/session/epoch identity, fail-closed preconditions, one-shot acknowledgement behavior, rejection, ambiguity and causal completion.

This PR intentionally does **not** modify `serve_physical_host.py`, does **not** complete or close #287, and does **not** claim that production can yet reach takeoff or the already-flying #276 motion surface. The remaining #287 work is the actual trusted-host control flow: under #273 establish #266 from explicit fresh flight-inactive safety, obtain the distinct #288/#267 teacher receipt only on the post-reset epoch, compose/activate the same-session watchdog/supervisor/SafeLink/ack state, and invoke the host-bound takeoff consumer with fresh lexical #278/#249 provenance. That follow-up must exercise the real production runner with deterministic seams rather than AST-extracted definitions.

No motorized checkpoint, real reset/takeoff/flight, arming, landing, vertical student motion, STOP/e-stop, generic raw SETPOINT_HL surface, or caller-selectable reset/teacher/provenance authority is requested or authorized.

Refs #287 #157 #276 #273 #266 #267 #262 #257 #249 #278 #271 #272 #279 #283 #288.


djibian:
DRAFT_FINDING `5554ce72e18264969433821e22940c6fd7102d0e` — follow-up ordering requirement for the remaining #287 production-host composition, **not a blocker to the now-narrowed machine-only core slice**.

Re-evaluation after the PR claim was narrowed: `takeoff_command.derive_bound_takeoff_command(...)` is pure/non-authority, and the actual CRTP packet factory remains inside `PhysicalExecutionDomain.effect_transaction(self._assert_current_authority)`, after the fail-closed current-program hook has run. The importable base transport cannot emit at all because that hook has no positive implementation. Therefore this HEAD may legitimately establish the typed command-9 semantics and causal effect core without yet proving the production host/profile path.

For the follow-up that completes #287, preserve the stronger ordering: the host-local #278/#249 assertion must positively re-establish the exact profile/canonical AST/new connection epoch before `_default_packet_factory` or any physical effect is reachable, and must be repeated after request construction immediately before fresh #257/#272/#271/emission. A height that is globally Runtime-valid but outside the active activity profile, or a non-canonical binding not equal to the browser-produced binding, must fail through the existing profile-bound preflight/provenance path rather than by duplicating profile bounds in `takeoff_command.py`.

No motorized checkpoint is implicated.

djibian (commented):
NO_GO 5554ce72e18264969433821e22940c6fd7102d0e

Independent falsification review against current main@b4633d92f173aef168102da065c1cecff233aa7d, the current #287 production-ordering finding, and the current ROADMAP.

The new command/core work is useful and bounded: takeoff height is derived from the exact canonical #267 AST binding rather than caller motion data, command 9 is typed with absolute height/current-yaw preservation/explicit 0.5 m/s velocity, the transport binds the concrete #266/#262/#257/#272/#271/#273 authorities, sends once without application retry, and only advances through the private #279 completion permit after causal fresh supervisor flight evidence.

The integration-blocking gap is exactly the Draft boundary already stated by this PR and sharpened on #287: there is still no production-shaped trusted-host control flow. This candidate does not modify `serve_physical_host.py`. Its positive transport fixture directly seeds `execution._PROCESS_STATE.phase = INACTIVE`, fabricates a powered-session/watchdog lifecycle with the test-only factory token, self-authorizes the teacher binding with a local `lambda`, and supplies a test-only `HostBoundTakeoffTransport` provenance override. Those are appropriate isolated core fixtures, but they do not establish the required production transition.

The ordering matters now that #288 is integrated: #266 deliberately rotates to a new live connection epoch, so a #267 teacher receipt minted before reset is stale after successful establishment. The actual host path must first perform the #273-protected #266 reset/postcondition transaction, then obtain the distinct #288 teacher decision on the new epoch, then construct/activate the same-epoch #257/#262/#272/#271 takeoff consumer, and only then enter command 9 with a fresh lexical #278/#249 assertion immediately before emission.

Smallest repair boundary: complete that actual trusted-host runner/control-flow path (or an explicitly injectable runner that is itself the production path), consuming the post-reset process-local #288 receipt and exact same-session safety objects. Add the negative ordering witness that a pre-reset teacher receipt cannot survive epoch rotation, and a positive deterministic witness that the post-reset receipt reaches the real host-bound takeoff consumer. Keep ordinary caller IPC unable to supply reset safety, powered-session/watchdog authority, teacher authority, provenance or raw takeoff parameters.

No motorized checkpoint or real reset/takeoff is required.

djibian (commented):
NO_GO 5554ce72e18264969433821e22940c6fd7102d0e

Independent falsification review against current main@b4633d92f173aef168102da065c1cecff233aa7d and the PR's narrowed Ready claim (machine-only exact-bound command/core, explicitly not production host reachability).

The prior production-reachability NO_GO is no longer applicable to this deliberately narrowed slice: the Ready description now explicitly leaves the real #287 host/reset→post-reset-teacher→takeoff path for a follow-up. The bounded transport core itself still preserves the useful concrete #266/#262/#257/#272/#271/#273/#279 composition, one plain no-retry send, definitive-rejection vs ambiguity semantics, and causal fresh supervisor completion.

One exact-semantic counterexample still blocks this narrower claim. `physical_capability_contract.bindAst()` on current main defines the authoritative `astBinding` as recursively key-sorted, whitespace-free canonical JSON. `takeoff_command._parse_ast_binding()` parses JSON and validates structure, but never proves that the supplied string is that canonical serialization. Therefore a semantically equivalent but non-canonical binding such as `{"version":1,"semantics":"webeeblocks-ast-v1","program":[{"kind":"takeoff","height_m":0.8},{"kind":"land"}]}` is accepted and reaches command-9 packet construction even though current-main `bindAst()` would canonicalize the top-level keys as `program,semantics,version`. This directly misses #287's deterministic acceptance requirement that malformed/non-canonical AST binding cannot reach packet construction and weakens the claim that the command is derived from the exact canonical binding.

Smallest repair: reject any parsed binding whose exact text is not the current-main canonical-JSON representation (recursively sorted object keys, no insignificant whitespace, JSON number/string semantics compatible with the browser contract), and add a deterministic counterexample using a structurally valid but differently ordered/otherwise non-canonical JSON string. Do not introduce a second incompatible serialization algorithm: either faithfully mirror the established canonical contract for the supported AST value domain or factor/reuse an exact shared definition. No production-host work or motorized checkpoint is required for this scoped repair.

djibian (commented):
NO_GO f643fd77c90f1097713091e836526bfd0e825d3e

Independent falsification review against current main@b4633d92f173aef168102da065c1cecff233aa7d and the exact narrowed Ready claim. The prior non-canonical key-order/whitespace/number-spelling counterexample is materially repaired, but the new Python serializer still does not mirror the authoritative current-main `physical_capability_contract.canonicalJson()` exactly.

Current-main `canonicalJson(value, path, depth)` starts at depth 0 and fails when `depth > 100`. Candidate `takeoff_command._canonical_json(value)` has no depth argument or nesting bound. A canonical-looking JSON string with a valid first top-level takeoff followed by more than 100 nested arrays therefore parses successfully in Python, round-trips through this candidate's `_canonical_json`, and reaches `derive_bound_takeoff_command()`/command-9 construction. The browser-side `bindAst()` rejects that same value as `nesting is too deep`, so it cannot be the exact #249/#267 canonical AST binding this Ready slice claims to consume.

There is a second exact-serialization mismatch in the same helper: browser `Object.keys(value).sort()` orders JavaScript strings by UTF-16 code units, while Python `sorted(value)` orders Unicode code points. For keys around a supplementary-plane/BMP boundary (for example `😀` versus `\ue000`), the order differs. The current takeoff parser validates only the first statement and otherwise allows such nested JSON values in the bound program, so this is not excluded structurally before the canonicality check.

Smallest repair: mirror the current-main canonical contract rather than an approximation: carry the same depth counter/`> 100` rejection through arrays and objects, and sort object keys by UTF-16 code-unit order (or otherwise factor an exact shared definition). Add deterministic counterexamples for an over-depth later program value and the UTF-16/code-point ordering difference, proving neither can reach command construction. Preserve the already-correct JSON number/string handling and the machine-only/no-effect scope. No production-host work or motorized checkpoint is required for this repair.

djibian (commented):
GO 725821282d7e1fb32b6adf2266c808fb661d1475

Independent falsification review against current main@11c0ef67c93db61a88b66b6367a77a5927b52b33 and the exact narrowed Ready claim.

The prior production-host reachability finding is no longer applicable to this deliberately machine-only core slice: the PR explicitly leaves the #273-protected #266 -> post-reset teacher decision -> same-session safety composition -> host-bound takeoff path to the remaining #287 work. Current main additionally integrates #290, closing the previously identified post-reset teacher-binding bootstrap prerequisite without expanding this PR's claim.

The two still-applicable exact-canonical refutations are resolved on this HEAD. `takeoff_command` now requires byte-for-byte canonical JSON after parsing in the JavaScript Number domain, mirrors the browser `depth > 100` boundary recursively, sorts object keys by UTF-16 code-unit order, and preserves the established JSON number/string presentation rules. Deterministic regressions cover reordered/whitespace/numeric-spelling alternatives, excessive nesting, and the supplementary-plane/BMP key-order counterexample.

The bounded effect core also survives falsification of its claimed safety boundary: direct import has no positive current-program path; exact #266/#262/#257/#272/#271/#273/#279 objects and same Crazyflie/epoch identities are required; command 9 is derived only from the exact teacher-bound AST with absolute height/current-yaw/0.5 m/s semantics; fresh not-flying/can-fly supervisor evidence and SafeLink are checked before emission; callback registration precedes one plain send; send/ack ambiguity remains recovery-required and poisons acknowledgement freshness; definitive rejection restores inactive; and a positive acknowledgement reaches `flying` only through the private completion permit plus causal fresh flying/high-level/finished supervisor evidence.

The small `watchdog_liveness` accessor exposes only the exact object identity already held by the guard and is used to reject cross-Crazyflie composition; it adds no effect or minting surface.

No motorized checkpoint is required for this scoped candidate.

PR #290, opened 2026-09-10
Publish post-reset binding on trusted teacher channel

djibian:
Adds the smallest no-effect prerequisite identified on #287 after integrated #266/#288.

#266 rotates the connection epoch, while #288's existing teacher-first request requires the peer to already know that exact host-minted post-reset epoch. This slice keeps the same distinct launcher-installed teacher socket and adds a host-first `PostResetTeacherDecisionChannel`: after trusted host composition has a post-reset `PhysicalRunBinding`, it publishes the exact profile/canonical AST/new epoch as `executionAuthority:false`, correlates one explicit teacher decision, seals peer write EOF, rechecks the live epoch throughout, and mints the existing #267 process-local receipt.

Stale pre-reset epoch fails before publication. The teacher peer cannot substitute profile/AST/epoch/correlation. Reconnect, denial and late/duplicate data fail closed. No physical reset/effect API is added, and the existing #288 teacher-first protocol remains unchanged.

Core CI runs the deterministic regressions through the existing teacher-authorization harness.

This PR does not itself perform #266 reset, modify `serve_physical_host.py`, complete #287, or authorize takeoff/flight/testing. It only makes the required post-reset teacher decision protocol executable for the later production-host path.

Refs #287 #266 #288 #267 #157.

### DIFF
--- a/tools/ci/test_post_reset_teacher_decision.py
+++ b/tools/ci/test_post_reset_teacher_decision.py
@@ -1,0 +1,306 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+import json
+from pathlib import Path
+import socket
+from threading import Thread
+
+ROOT = Path(__file__).resolve().parents[2]
+PHYSICAL = ROOT / "tools" / "physical"
+
+import sys
+sys.path.insert(0, str(PHYSICAL))
+
+import post_reset_teacher_decision as post_reset  # noqa: E402
+import teacher_decision_channel as base_channel  # noqa: E402
+import teacher_run_authorization as auth  # noqa: E402
+
+
+def require(condition: bool, message: str) -> None:
+    if not condition:
+        raise AssertionError(message)
+
+
+def expect_error(callable_, pattern: str) -> None:
+    try:
+        callable_()
+    except base_channel.TeacherDecisionChannelError as exc:
+        require(pattern in str(exc), f"expected {pattern!r} in {exc!r}")
+        return
+    raise AssertionError("expected TeacherDecisionChannelError containing " + repr(pattern))
+
+
+class Epoch:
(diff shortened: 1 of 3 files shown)


djibian (commented):
GO 71b7fdd79c2e80c074a653c5b3cdf9e472f50403

Independent falsification review against current main@b4633d92f173aef168102da065c1cecff233aa7d, the #287 post-reset teacher-binding bootstrap finding, and integrated #266/#267/#288.

This bounded no-effect slice closes the identified protocol gap without weakening the existing authority boundary. `PostResetTeacherDecisionChannel` remains on the exact launcher-installed teacher socket, requires a host-owned `PhysicalRunBinding` whose epoch is currently live, publishes that exact profile/canonical-AST/new-epoch tuple as `executionAuthority:false`, and accepts only one exactly correlated boolean decision for the same tuple. It rechecks the live epoch before publication, after the decision, after peer write-side sealing, and after #267 minting; stale/pre-reset, substituted profile/AST/epoch/correlation, reconnect, denial, duplicate/late data and ambiguous transport all fail closed before a reusable receipt can escape.

The change does not add reset, watchdog, flight, packet, HighLevelCommander or ordinary caller/browser effect authority. It leaves the existing #288 teacher-first protocol intact and adds the host-first variant only as the prerequisite later production #287 composition can select after #266 has established the new epoch. Deterministic regressions exercise the exact positive post-reset proposal/receipt path plus stale epoch, substitution, reconnect, denial and duplicate/late counterexamples, and the existing core harness now invokes them through the teacher-authorization test.

The earlier #287 finding remains correctly scoped to the follow-up production host path: this PR does not claim to perform #266 reset, modify `serve_physical_host.py`, or reach takeoff. No motorized checkpoint is implicated.

PR #294, opened 2026-09-10
Preserve capability bridge across trusted reset

djibian:
Adds the smallest no-effect #287 substrate needed to keep the already-configured production #249 responder usable across the #266 reset/epoch rotation.

This slice introduces `PostResetCapabilityHttpBridge`, a read-only bridge that preserves its existing loopback address and bearer tokens while making the session handoff explicit and fail-closed:

- `begin_post_reset_replacement()` is intended for the trusted #266 prior-evidence invalidation edge and immediately makes the old bridge view unavailable;
- replacement refuses a stale/wrong current epoch, an active #249 assertion, duplicate replacement, shutdown, or a post-reset session that reuses the prior epoch;
- `install_post_reset_session()` reopens the same bridge identity only for a genuinely new live epoch;
- #249 assertion admission is serialized against the replacement edge so reset cannot slip between the assertion's first epoch read and publication of its pending challenge;
- deterministic coverage forces that exact race window and proves it fails closed, while normal HTTP reads after replacement see only the new read-only session and `executionAuthority:false` remains unchanged.

The regression is included in the existing Runtime core CI path through the teacher-authorization contract chain.

This PR deliberately does **not** modify `serve_physical_host.py`, expose the raw Crazyflie, perform or authorize a reset, compose #266 itself, mint teacher authority, emit command 9, or complete #287. The remaining production work still has to keep the post-reset raw Crazyflie lexical to the trusted host, compose #273-protected #266 -> #290 -> same-session #257/#262/#272/#271 -> #289, and prove the post-reset read-only view and effect object are the same live session/epoch.

No motorized checkpoint or real reset/takeoff is requested or authorized.

Refs #287 #266 #278 #249 #290 #289 #157.

### DIFF
--- a/tools/ci/test_post_reset_capability_bridge.py
+++ b/tools/ci/test_post_reset_capability_bridge.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import json
 from pathlib import Path
 import sys
-from threading import Thread
+from threading import Event, Thread
 from urllib.error import HTTPError
 from urllib.request import Request, urlopen
 
@@ -33,6 +33,22 @@ class FakeSession:
         }
 
 
+class BlockingFirstEpochSession(FakeSession):
+    def __init__(self, epoch: str, label: str) -> None:
+        super().__init__(epoch, label)
+        self.first_read_started = Event()
+        self.release_first_read = Event()
+        self._read_count = 0
+
+    def read_connection_epoch(self) -> str:
+        self._read_count += 1
+        if self._read_count == 1:
+            self.first_read_started.set()
+            if not self.release_first_read.wait(1.0):
+                raise RuntimeError("test did not release first epoch read")
+        return super().read_connection_epoch()
+
+
 def require(condition: bool, message: str) -> None:
     if not condition:
         raise AssertionError(message)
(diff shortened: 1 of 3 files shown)


djibian (commented):
NO_GO be8a4f0c3d77e20ad0e35fc7a824447340005705

Independent falsification review against current main@ac46378c61e3aab6ae53083c867bfbcafbb271a3 and the exact Ready claim.

The two-phase replacement invariant is bypassable through the subclass's public `session` property setter. `PostResetCapabilityHttpBridge.__init__()` needs the setter only because the base constructor assigns `self.session = session`, but the setter remains callable afterward and unconditionally writes `_session`. Thus, while no replacement is pending, code holding this bridge can execute `bridge.session = other_live_session` and rotate the read-only view without `begin_post_reset_replacement()`, without checking the previous epoch, and without requiring a genuinely new epoch through `install_post_reset_session()`. This contradicts the candidate's explicit claim that later rotation is permitted only through the two-phase methods and leaves the old view-to-new-view transition outside the fail-closed replacement boundary.

Smallest repair: make the property assignment construction-only (or avoid the base assignment path), and reject any later direct `session = ...` mutation. Add the deterministic counterexample: direct post-construction assignment of a different/same-epoch session must fail and must not alter the bridge's observable epoch/capability view. Preserve the existing two-phase replacement and assertion-admission race coverage. No physical checkpoint is required.

djibian (commented):
NO_GO 8d3ed403da0fdff191b14697a3fe713f16ca712d

Independent falsification review against current main@ac46378c61e3aab6ae53083c867bfbcafbb271a3 and the exact Ready claim.

The prior direct-`session`-assignment finding is resolved, but the claimed fail-closed reset cutover is still not linearizable for ordinary capability reads. The HTTP handlers evaluate `bridge.session.read_connection_epoch()` / `bridge.session.read_capabilities()` in two steps: the subclass property acquires `_session_lock`, checks `_replacement_previous_epoch`, returns the old session, and releases the lock *before* the actual session read runs. A request can therefore obtain the pre-reset session, block inside `read_capabilities()` (or the epoch read), then `begin_post_reset_replacement()` can acquire the locks, mark replacement pending, return to #266, and allow the reset to proceed while that already-issued request continues on the stale session and can later return pre-reset data. This contradicts the candidate's explicit boundary that beginning replacement immediately makes the old bridge view unavailable and leaves a read crossing the prior-evidence invalidation edge.

Smallest repair boundary: make the read-only capability/epoch operations participate in the same replacement cutover, so `begin_post_reset_replacement()` cannot return while an admitted old-session read can still complete as a successful old view, and no new read can be admitted once replacement starts. A bridge-owned read method / active-reader fence is preferable to relying on a property getter whose lock ends before the session method call. Add deterministic coverage with a blocking old-session capability (and ideally epoch) read: start the HTTP/read operation, begin replacement concurrently, prove the cutover either drains that read before returning or causes it to fail closed, and prove no successful stale response can cross the completed replacement-begin edge. Preserve the repaired construction-only setter and the existing #249 assertion-admission race coverage. No physical checkpoint is required.

djibian (commented):
GO 8d3ed403da0fdff191b14697a3fe713f16ca712d

Independent falsification review against current main@ac46378c61e3aab6ae53083c867bfbcafbb271a3 and the exact Ready claim.

The prior NO_GO on be8a4f0c3d77e20ad0e35fc7a824447340005705 is resolved on this HEAD: direct post-construction `session = ...` assignment is construction-gated and rejected, with deterministic same-epoch and different-epoch counterexamples proving the observable bridge view remains unchanged. The explicit replacement path still refuses stale/wrong old epochs, duplicate replacement, shutdown, active #249 assertions and same-epoch reinstallation; assertion admission is serialized across the pre-read/pending-registration race, while replacement remains read-only/non-authority and preserves the configured loopback/tokens only after a genuinely new epoch is installed.

I found no remaining applicable refutation within this bounded no-effect substrate. This GO does not claim production host composition, reset authority, teacher authority, command emission or real-flight acceptance.

djibian (commented):
GO 4891df8f80a25f1f3c8bb9cfd26b960d985aab3e

Independent falsification review against current main@ac46378c61e3aab6ae53083c867bfbcafbb271a3 and the exact Ready claim.

Both still-applicable prior findings are resolved on this exact HEAD. Direct post-construction `session = ...` mutation remains construction-gated and covered for same-epoch and different-epoch attempts. The later stale-read cutover finding is repaired by keeping ordinary epoch/capability reads inside the bridge-owned session lock, including the HTTP response publication boundary, so `begin_post_reset_replacement()` cannot complete while an admitted old-session response can still be published and requests admitted after the cutover fail closed. The exact counterexample is now deterministic: a blocked old `/v1/capabilities` read must drain before replacement begin returns, then a subsequent capability request receives replacement-pending rather than stale data.

The existing #249 assertion-admission fence, stale/wrong epoch rejection, same-epoch reinstall rejection, shutdown/duplicate checks, preserved loopback/token identity, and `executionAuthority:false` boundary remain intact. I found no remaining applicable refutation within this bounded no-effect substrate. This GO does not claim production host composition, reset/teacher authority, command emission, or real-flight acceptance.

djibian (commented):
GO 540b0a1d78b2a87f8cc50ea8f162cdfe85f40e1f

Independent falsification review against current main@ac46378c61e3aab6ae53083c867bfbcafbb271a3 and the exact Ready claim.

The exact current HEAD changes only deterministic test teardown relative to the previously reviewed 4891df8f80a25f1f3c8bb9cfd26b960d985aab3e: it starts the missing assertion-race HTTP worker and verifies every bridge worker actually stops after shutdown. It does not alter the production replacement substrate.

Both still-applicable prior findings remain resolved. Direct post-construction `session = ...` mutation is construction-gated and covered for same-epoch and different-epoch attempts. Ordinary epoch/capability HTTP reads stay inside the bridge-owned session lock through response publication, so `begin_post_reset_replacement()` drains an admitted old view before returning and no successful stale response can cross the completed cutover. #249 assertion admission remains serialized against replacement; wrong/stale epochs, duplicate replacement, shutdown, active assertions and same-epoch reinstall remain fail-closed; stable loopback/token identity is preserved only after a genuinely new epoch is installed; and the surface remains read-only with `executionAuthority:false`.

I found no remaining applicable refutation within this bounded no-effect substrate. This GO does not claim production host composition, reset authority, teacher authority, command emission, or real-flight acceptance.

Issue #296, opened 2026-09-11
Automate durable publication of future physical-test evidence


djibian:
## Goal

Make publication of future physical-test evidence a first-class, reusable project mechanism so raw test outputs do not depend on manual GitHub attachments and remain directly usable by stateless Controllers.

Owner instruction (2026-09-11): future test data should be published by the AI/tooling whenever it has access to the generated files, with a deterministic local bridge when the files originate on the test machine. Manual `github.com/user-attachments` uploads must not be the canonical evidence path.

This is deliberately separate from #292. PR #292 repaired historical accessibility for already-acquired #180/#236/#251 evidence; this issue defines the durable mechanism for future tests.

Refs #70 #292.

## Product / governance boundary

This work concerns the checkpoint/evidence mechanism under the current `AGENTS.md` Human Boundary and CI/evidence rules. It must preserve those authority boundaries:

- real physical action and human PASS/FAIL/NOT_NEEDED remain human decisions;
- publication of raw machine-generated evidence is not itself a human verdict;
- Controllers/AI may collect, verify, materialize and publish evidence, but must not fabricate or infer owner-authoritative human results;
- real flight/publication authority is not broadened by this mechanism;
- evidence must remain bound to the exact tested candidate/request.

The implementation may technically change checkpoint tooling, so it requires the governance scrutiny applicable to changes affecting checkpoint/notification mechanisms and evidence-producing scripts.

## Required outcome

For a future physical checkpoint that produces files, the normal path should be:

`TEST_REQUIRED / exact test request -> deterministic preparation -> physical execution -> local output files -> automated evidence publication -> durable Git/GitHub evidence -> human result`

The owner should not need to manually attach archives to an issue merely to make test data available to Controllers.

A local helper/agent command is an acceptable bridge where a remote Controller cannot directly access files on the physical test machine. The intended UX is that the AI/tooling invokes or guides that deterministic bridge; manual GitHub attachment remains optional convenience only.

## Evidence contract

The generic mechanism must, for every published evidence set:

1. bind the evidence to the exact checkpoint/request and exact tested Git SHA;
2. record the test profile/purpose and enough provenance to identify how the files were produced;
3. preserve raw bytes rather than silently normalizing measurement files (including line endings when relevant);
4. calculate and record cryptographic digests (SHA-256) and file sizes;
5. provide a machine-verifiable manifest and fail closed on missing/mismatched files;
[...]

PR #292, opened 2026-09-10
Import raw #70 checkpoint traces for machine replay

djibian:
Makes the already-acquired physical evidence from #180, #236 and #251 machine-readable to stateless Controllers.

This PR is intentionally opened Draft while a temporary branch-only importer retrieves the owner-attached archives from `github.com/user-attachments`, verifies the published SHA-256 digests where available, safely extracts UTF-8 evidence including all CSV traces, records source/archive/file provenance, commits those files, and removes the importer workflow from the branch.

Final intended diff: evidence/provenance files only under `experiments/crazyflie-ukf-surface-range/evidence/`; no workflow change, firmware change, Runtime change, retuning, physical test or flight authority.

Refs #70 #180 #236 #251.

PR #297, opened 2026-09-11
Establish the missing raw inputs for the X3 firmware proof

djibian:
The #70 architecture review requires archived IMU/barometer evidence to be inspected before another firmware candidate or physical test. #292 made the retained traces accessible, so the former access limitation no longer explains the missing proof.

This change verifies all 94 retained files against both import manifests, inventories all 76 CSVs while preserving exclusions, and records the reproducible derived result: continuous raw barometer, accelerometer and gyroscope samples are absent. All 4,448 #180 baroHeight samples are zero from an inactive log; S3 event-relative and held snapshots cannot substitute for a continuous independent signal. No independent metric vehicle-Z trajectory or measured mixed case is retained.

The scientific result is UNPROVEN, not a new physical FAIL/PASS. The report and smallest X3 roadmap update make the next boundary explicit: acquisition plus a frozen offline calculation/reference, using the existing #251 firmware and unchanged parameters. The pinned firmware already exposes the missing sensor logs; no estimator patch is justified merely to obtain them. #296 remains the separate generic evidence-publication workstream.

Validation: regenerated inventory is byte-identical; all retained hashes and sizes verified; changed, missing and unexpected raw-file controls are rejected. The audit emits no inferred physical verdict and executes no historical capture code. git diff --check passes. No firmware, Runtime, CI/checkpoint mechanism or motor behavior is changed; no checkpoint is requested.

Refs #70 #292. Independent review must verify the scientific scope and source/manifest provenance under current main AGENTS.md.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -338,11 +338,19 @@ justified by the current evidence.
   estimate from IMU/barometer evidence that excludes suspect ToF can bound
   `delta z` tightly enough that `delta h = delta z_ind - delta c` distinguishes
   terrain, true vertical motion and a mixed event
-- next proof: reanalyse the existing #180/#236/#251 archives before collecting
-  new physical data. Verify provenance/hashes and available columns, use the raw
-  barometer plus IMU/attitude when present, quantify uncertainty/latency, and
-  compare the same frozen calculation across stationary, terrain, vertical and
-  mixed controls with an independent metric reference where one exists
+- archived-input result: #292 made the retained #180/#236/#251 text accessible;
+  the [input audit](../experiments/crazyflie-ukf-surface-range/evidence/analysis-2026-09-11/README.md)
+  verifies all 94 retained files. No CSV contains continuous raw barometer or
+  accelerometer/gyroscope samples; no independent metric vehicle-Z trajectory
+  or measured mixed case is retained. Inactive/conditional diagnostics cannot
+  substitute for these inputs. The independent-displacement proof is UNPROVEN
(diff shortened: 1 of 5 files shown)


djibian (commented):
GO bf2f75ffb5583e77bf32c9e4029cbf2b7db19b51

Independent falsification review against current main@c65c5561690c5c1703082ae9367e43961d032320 and #70 architecture conclusion 5606273253.

The candidate stays inside the authorized Lab-only evidence boundary. The audit reads retained #292 bytes only, checks the SOURCE.json and MANIFEST.sha256 inventories for exact path/hash/size agreement, preserves explicit excluded CSVs/omissions, and does not reinterpret imported hashes as independent capture authentication. The generated inventory is consistent with the stated scope: 50+16+28 = 94 retained files, 36+14+26 = 76 CSVs, raw barometer/accelerometer/gyroscope columns are absent from non-excluded CSVs, and the four #180 baroHeight streams contain 1115+1000+1582+751 = 4,448 finite samples with one distinct value, 0.0.

The scientific conclusion is correctly bounded to UNPROVEN because the pre-registered independent-displacement calculation cannot be reconstructed without the required continuous independent inputs/reference. Conditional/held S3 diagnostics and UKF/ToF-derived state are explicitly not promoted to substitutes. The next-step guidance preserves the frozen 0.5 s before/after statistic, independent metric reference, unchanged #251 firmware/parameters, and no-motorized/no-checkpoint boundary. I found no candidate-local path that manufactures a physical PASS/FAIL or authorizes a new estimator patch/test.

No applicable refutation found for this exact HEAD.

PR #298, opened 2026-09-11
Automate durable physical-test evidence publication

### LINKED ISSUE #296: Automate durable publication of future physical-test evidence

djibian:
Implements the generic future evidence path requested by #296 without changing physical-test authority or synthesizing human results.

The new `tools/evidence/publish_test_evidence.py` local bridge consumes a registered output profile and exact local test-output directory, preserves raw bytes, binds the evidence to the exact checkpoint/request and tested Git SHA, records file sizes and SHA-256 digests, bundles the exact profile contract, writes a machine-verifiable `MANIFEST.sha256`, and verifies the materialized copy before publication. The initial `physical-csv-text-v1` profile covers bounded CSV/text/JSON/log evidence and requires at least one CSV; future bounded output shapes use a new repository-controlled versioned profile rather than silently widening an existing evidence contract. Fresh verification requires the bundled profile bytes/digest to match that registered profile exactly.

The verifier is fail-closed not only on paths, digests and profile bounds but also on storage provenance: the evidence directory and required top-level contract artifacts must be real in-tree objects, `raw/` must be a real non-symlink directory, and symlinks anywhere under `raw/` are rejected. Decision-relevant bytes therefore cannot be supplied through checkout-dependent external indirection.

The `publish` path is deliberately optimistic and fail-closed: it requires a clean checkout on an exact base SHA, reconstructs remote `main`, refuses an already-existing evidence branch, commits only the evidence destination, and rechecks remote base/branch state immediately before publication. The actual destination-ref creation is an atomic create-if-absent CAS using an explicit empty-expected `--force-with-lease=refs/heads/<branch>:`; a branch created in the final check→push gap is rejected even when an ordinary push would have been a valid fast-forward. Exact post-push ref verification remains mandatory, and failed or ambiguous publication keeps the local source/evidence without overwriting concurrent work.

`tools/ci/test_evidence_publication.py` provides deterministic coverage for CRLF/raw-byte preservation through Git staging, exact SHA/request/provenance binding, missing and unexpected evidence rejection, post-copy mutation/digest detection, manifest tampering, profile file/byte bounds, rejection of a self-consistent but more permissive bundled profile, rejection of raw-directory and top-level contract-file symlink indirection, successful remote publication, ordinary pre-existing-branch refusal, and an injected check→push race that creates the destination ref after the final absence read and proves that ref is preserved. The existing CI Gate select contract runs this test; no new workflow, trigger, permission, notification or human-test queue is introduced.

`docs/TEST_EVIDENCE.md` makes Git-backed raw evidence the canonical future path and manual `github.com/user-attachments` non-canonical. `docs/CONTROLLER_NOTIFICATIONS.md` documents the publication bridge while preserving the existing single `TEST_REQUIRED` boundary. Published `EVIDENCE.json` fixes `human_verdict` to `null`; PASS/FAIL/NOT_NEEDED remain separate owner-authoritative actions bound to the applicable request/candidate.

This is governance/evidence-mechanism work explicitly requested by #296, so the `.github/workflows/ci.yml` trust-boundary change is intentionally narrow and requires the stricter independent review applicable to checkpoint/evidence mechanisms before integration.

Closes #296. Refs #70 #292.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,7 @@ jobs:
           python3 tools/ci/test_repository_hygiene.py
           python3 tools/ci/test_windows_release_contract.py
           python3 tools/ci/test_controller_contract.py
+          python3 tools/ci/test_evidence_publication.py
       - name: Select required suites
         id: scope
         run: >-
--- a/docs/CONTROLLER_NOTIFICATIONS.md
+++ b/docs/CONTROLLER_NOTIFICATIONS.md
@@ -29,6 +29,23 @@ provenance/digest and procedure are ready. Enabled profiles are:
 
 Any unknown profile is rejected until its preparation is implemented explicitly.
 
+## Raw output publication
+
+When an enabled or future checkpoint produces bounded CSV/text/log output, the
+canonical data path is Git-backed evidence, not a browser-only issue attachment.
+Use the repository-controlled profile plus
+`tools/evidence/publish_test_evidence.py` as described in
+[`TEST_EVIDENCE.md`](TEST_EVIDENCE.md). AI/tooling should invoke that bridge when
+it can access the files; on a physical test machine it may guide the same
+deterministic local command.
+
+The helper binds the exact request and tested SHA, preserves raw bytes, records
+sizes/SHA-256/provenance, verifies a manifest, and race-safely pushes a new
+evidence branch without force or overwrite. A Controller then opens the short
+Draft PR. Publication leaves `human_verdict` null: PASS/FAIL/NOT_NEEDED remains a
+separate owner-authoritative action. Manual `github.com/user-attachments` uploads
+may remain optional convenience copies but are non-canonical.
+
 ## Human concurrency and idempotence
 
 There may be only one unresolved TEST_REQUIRED globally. There is no queue. If a
--- a/docs/TEST_EVIDENCE.md
+++ b/docs/TEST_EVIDENCE.md
@@ -68,7 +68,12 @@ python3 tools/evidence/publish_test_evidence.py \
 
 Verification fails closed on a missing, unexpected, modified, size-mismatched or
 digest-mismatched raw file, malformed metadata/profile, changed manifest, invalid
-tested commit, or byte-preservation attributes.
+tested commit, or byte-preservation attributes. It also requires the bundled
+`PROFILE.json` bytes and digest to match the repository-controlled registered
+profile for that versioned id; a self-consistent evidence directory cannot widen
+its own file/size contract. `raw/`, `EVIDENCE.json`, `PROFILE.json`,
+`MANIFEST.sha256` and `.gitattributes` must be real in-tree files/directories,
+not symlink indirections to checkout-dependent external content.
 
 ## Publish without a manual attachment
 
--- a/docs/TEST_EVIDENCE.md
+++ b/docs/TEST_EVIDENCE.md
@@ -1,0 +1,119 @@
+# Durable physical-test evidence
+
+Issue #296 defines the canonical path for future bounded CSV/text/log evidence.
+Raw files produced by a physical checkpoint should become ordinary Git content;
+a browser-only `github.com/user-attachments` upload may be convenient transport,
+but it is not canonical project evidence.
+
+The publication path is:
+
+`exact request -> local output directory -> materialize/verify -> new evidence branch -> Draft PR -> human verdict`
+
+`tools/evidence/publish_test_evidence.py` is the deterministic local bridge. It
+does not run a physical test and it never records `PASS`, `FAIL` or `NOT_NEEDED`.
+
+## Registering a test-output profile
+
+Each reusable output shape is a repository-controlled JSON file under
+`tools/evidence/profiles/`. A profile declares:
+
+- a stable versioned profile id;
+- `required_globs`, each of which must match at least one produced file;
+- `allowed_globs`, outside of which publication fails closed;
+- `max_files` and `max_total_bytes` bounds.
+
+Adding a profile is sufficient to describe a new bounded output shape; no
+experiment-specific importer workflow is required. Version an incompatible
+profile rather than silently changing the meaning of evidence already published
+under an older id.
+
+The initial `physical-csv-text-v1` profile requires at least one CSV and allows
+bounded CSV, text, JSON and log files.
+
+## Materialize and verify locally
+
+The source directory is never modified or removed. From a checkout that contains
+the exact tested commit:
+
+```sh
+python3 tools/evidence/publish_test_evidence.py \
+  --repo-root . materialize \
+  --source /path/to/test-output \
+  --destination experiments/<experiment>/evidence/checkpoint-<N>-<sha12> \
+  --profile physical-csv-text-v1 \
+  --target-sha <exact-tested-40-char-sha> \
+  --checkpoint-ref issue-<N> \
+  --purpose checkpoint \
+  --request https://github.com/djibian/webeeblocks/issues/<N>#issuecomment-<ID> \
+  --provenance '<one-line capture command or equally precise production description>'
+```
+
+The resulting directory contains:
+
+- `raw/` — byte-for-byte copies of the registered source files;
+- `PROFILE.json` — the exact profile contract used for this capture;
+- `EVIDENCE.json` — exact SHA/request/profile/purpose/capture provenance, sizes,
+  digests, and an explicit `human_verdict: null`;
+- `MANIFEST.sha256` — an independently checkable raw-file SHA-256 manifest;
+- `.gitattributes` — `raw/** -text`, so Git does not normalize CRLF measurement
+  bytes or turn them into whitespace failures.
+
+A fresh Controller or operator can verify an evidence directory with:
+
+```sh
+python3 tools/evidence/publish_test_evidence.py \
+  --repo-root . verify \
+  --evidence-dir experiments/<experiment>/evidence/checkpoint-<N>-<sha12>
+```
+
+Verification fails closed on a missing, unexpected, modified, size-mismatched or
+digest-mismatched raw file, malformed metadata/profile, changed manifest, invalid
+tested commit, or byte-preservation attributes.
+
+## Publish without a manual attachment
+
+When tooling/AI has access to the local files and authenticated Git push is
+available, use `publish` instead of manually staging an archive:
+
+```sh
+python3 tools/evidence/publish_test_evidence.py \
+  --repo-root . publish \
+  --source /path/to/test-output \
+  --destination experiments/<experiment>/evidence/checkpoint-<N>-<sha12> \
+  --profile physical-csv-text-v1 \
+  --target-sha <exact-tested-40-char-sha> \
+  --checkpoint-ref issue-<N> \
+  --purpose checkpoint \
+  --request https://github.com/djibian/webeeblocks/issues/<N>#issuecomment-<ID> \
+  --provenance '<one-line capture command or equally precise production description>' \
+  --base-sha <exact-current-main-sha> \
+  --base-ref main \
+  --branch evidence/issue-<N>-<sha12>
+```
+
+Publication requires a clean checkout exactly at `--base-sha`, reconstructs the
+remote `main` SHA, refuses an already-existing evidence branch, creates the local
+branch, materializes and re-verifies the evidence, commits only that destination,
+rechecks remote `main` and the destination branch immediately before push, and
+uses a normal non-force push. It then verifies that the remote branch resolves to
+the exact created commit.
+
+If the remote base moves, the evidence branch appears concurrently, push is
+rejected, or remote verification is ambiguous, the command reports failure and
+keeps the local source/evidence for recovery. It never deletes raw source data.
+
+After a successful push, the Controller/AI opens a short Draft PR from the
+reported branch to `main`, links the exact checkpoint/request, and lets normal
+governance/independent review validate the evidence mechanism and content. The
+branch plus PR are project state; no Controller ownership marker is needed.
+
+## Authority boundary
+
+Raw publication is machine evidence, not a real-world acceptance verdict.
+`human_verdict` is deliberately fixed to `null`; this helper has no command-line
+surface that can publish `PASS`, `FAIL` or `NOT_NEEDED`. Those remain separate,
+owner-authoritative actions bound to the applicable TEST_REQUIRED request.
+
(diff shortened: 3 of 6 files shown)


djibian (commented):
NO_GO 166057c681fbf8425c0d54f1bc70d880bb4fa889

The publication core still has a check→push race that can move a concurrently-created evidence branch. `publish()` reconstructs absence with `ls-remote` immediately before `git push`, but the push is an ordinary non-force update. If another actor creates the same branch in the gap pointing to `base_sha` (or another ancestor of our evidence commit), our push is a valid fast-forward and therefore moves that concurrent ref instead of refusing it. The existing “race” test only covers a branch that already exists before `publish()` begins, so it cannot falsify this TOCTOU.

Smallest repair boundary: make the ref creation itself an atomic create-if-absent CAS, for example an empty expected `--force-with-lease=refs/heads/<branch>:` lease (or an equivalent exact Git primitive), retain post-push exact-ref verification, and add a deterministic test that creates the destination branch after the final absence reconstruction but before push and proves the concurrent ref is preserved. Update the documentation/body that promises a normal non-force push, since ordinary fast-forward semantics cannot establish the claimed non-overwrite guarantee.

No human checkpoint is implicated.

djibian (commented):
DRAFT_FINDING ceca6c006891da049fe5a28e5a02f841d1b72926

The existing atomic-create NO_GO remains applicable on this merge-only HEAD. A second fail-closed gap is visible in the evidence verifier: the bundled profile's `max_files` and `max_total_bytes` constraints are validated as positive integers but are never enforced against the materialized `raw/` set in `verify()`.

`materialize()` does enforce both bounds on the source, so the normal producer path is bounded. But the acceptance contract explicitly requires a fresh stateless Controller to retrieve and machine-verify durable evidence. A self-consistent committed evidence directory can currently contain more than the profile's `max_files` or `max_total_bytes` if its `EVIDENCE.json` records, digests and manifest are updated consistently; `verify()` checks paths/globs/digests but will accept that profile violation. That makes verification weaker than the bundled profile contract it claims to verify.

Smallest repair: after reconstructing the exact raw file set in `verify()`, enforce `len(actual_paths) <= max_files` and the sum of verified record/file sizes `<= max_total_bytes`, and add deterministic self-consistent tampering regressions for both limits (not just unexpected path/digest mismatch). Preserve the bundled immutable profile approach and the current byte-preservation semantics.

No human checkpoint is implicated.

djibian (commented):
GO 56322e0ab4677d6e95574e159bbae64d2a818b1b

Independent governance/evidence review against current main@4603b9809cc9e14486248101accd3bb1cf12bf1b, owner-authorized issue #296, and both prior authoritative findings on #298. This execution did not modify #298.

The atomic-publication NO_GO is resolved. The destination ref is still reconstructed immediately before publication, but the write authority no longer relies on that read: `git push --force-with-lease=refs/heads/<branch>:` supplies an empty expected value, making destination creation an atomic create-if-absent CAS. The deterministic regression injects a ref creation after the final absence read and immediately before push, requires publication to fail, and proves the concurrently created ref remains at its original SHA. Exact post-push ref verification and fail-closed retention of local source/evidence are preserved. The detailed publication documentation accurately explains that the empty lease is not authority to overwrite an existing branch.

The later verifier-bound finding is also resolved. Fresh `verify()` now enforces both bundled-profile limits against the reconstructed durable raw set: exact file count and summed verified record sizes. The regression rewrites a bundled profile plus its bound digest into a self-consistent smaller `max_files` or `max_total_bytes` contract and proves verification rejects each violation, so a fresh stateless Controller cannot accept evidence that exceeds the profile merely because manifest/digests were updated consistently.

The remaining mechanism matches #296's authority boundary: registered profiles bound allowed/required files and bounded size; raw bytes and CRLF survive Git staging; exact request/tested SHA, profile digest, sizes, SHA-256 records and manifest are retained; missing/unexpected/modified evidence fails closed; the evidence destination is create-only; publication never synthesizes PASS/FAIL/NOT_NEEDED (`human_verdict` is fixed null); no new notification class, human-test queue, workflow trigger or permission is introduced. The `.github/workflows/ci.yml` trust-boundary delta only runs the deterministic evidence-publication contract test inside the existing selector verification step.

The merge of current main/#299 adds only the already-integrated independent-input acquisition files and does not alter the evidence mechanism. I found no remaining applicable refutation on this exact Ready HEAD. Fresh exact-HEAD `CI Gate` remains a separate integration prerequisite.

djibian (commented):
NO_GO 56322e0ab4677d6e95574e159bbae64d2a818b1b

Independent falsification review against current main@4603b9809cc9e14486248101accd3bb1cf12bf1b and the stricter evidence/checkpoint trust boundary. The prior atomic-create and bundled-profile-limit findings are repaired on this exact Ready HEAD, but fresh verification still accepts decision-relevant evidence whose apparent Git content is actually supplied through working-tree symlinks.

`verify()` resolves and reads the top-level `EVIDENCE.json`, `PROFILE.json`, `MANIFEST.sha256` and `.gitattributes` without rejecting symlinks. More critically, it checks only `raw_root.is_dir()`; a committed `raw` symlink to another directory passes that test, `raw_root.rglob('*')` traverses the target, and each resulting `raw/<file>` is not itself a symlink. A self-consistent metadata/manifest set can therefore verify successfully while the raw bytes are not stored under the evidence directory in Git at all and can vary with external checkout state. That contradicts the PR's canonical Git-backed, statelessly machine-verifiable evidence claim.

Smallest repair boundary: fail closed unless `raw` is a real non-symlink directory and every required top-level evidence artifact is a real non-symlink regular file; keep rejecting symlinks anywhere under `raw`. Add a deterministic tampering regression that replaces the materialized `raw/` directory with a symlink to matching external files (and ideally top-level contract-file symlinks) and proves fresh `verify()` rejects it. No physical checkpoint is implicated.

djibian:
NO_GO 56322e0ab4677d6e95574e159bbae64d2a818b1b

Fresh independent evidence-contract review found one remaining fail-closed gap not covered by the existing GO. `verify()` treats the bundled `PROFILE.json` as authoritative after checking only that its SHA-256 matches `EVIDENCE.json`. It never compares those bytes/digest to the repository-controlled registered profile at `tools/evidence/profiles/<id>.json` (nor otherwise binds the accepted profile to trusted repository state).

That means a self-consistent durable evidence directory can widen its own contract: change `PROFILE.json` to increase `max_files`/`max_total_bytes` or broaden `allowed_globs`, update `EVIDENCE.json.profile.sha256`, update raw-file records/manifest consistently, and fresh `verify()` will accept the widened profile. The new limit regressions actually demonstrate the underlying self-contained-profile behavior by rewriting the bundled profile and its digest; they only test *smaller* limits. With larger limits or broader globs, the same self-consistent rewrite bypasses the repository-registered contract. This contradicts #296's requirement that expected evidence be validated according to the registered test profile and the documentation's statement that profiles are repository-controlled/versioned.

Smallest repair boundary: in fresh verification, resolve the registered profile id from the repository and require the bundled profile bytes (or at minimum its digest) to equal that registered version before applying its constraints. Because incompatible profiles are already required to get a new versioned id, exact-byte binding is compatible with the documented immutability model. Add a deterministic counterexample that rewrites the bundled profile plus metadata digest to a *more permissive* contract and proves verification rejects it. Preserve the existing raw-byte, manifest, create-only publication and atomic destination-ref semantics.

No human checkpoint is implicated.

djibian (commented):
GO d711e61cede02e985e222b663d47ed3490c194e1

Independent falsification review against current main@83b205b8b1b403fe243b6b16fbd79da13efb5bfd and the stricter governance/evidence trust boundary. This execution has not modified #298.

The exact-HEAD repair resolves the remaining symlink-provenance NO_GO. Fresh `verify()` rejects a symlinked evidence directory, requires each top-level contract artifact (`EVIDENCE.json`, `PROFILE.json`, `MANIFEST.sha256`, `.gitattributes`) to be a real regular file, requires `raw/` to be a real non-symlink directory, and rejects symlinks encountered anywhere beneath it before accepting the exact file set. The deterministic regressions replace `raw/` with an external-directory symlink and replace `EVIDENCE.json` with an external-file symlink and require fail-closed rejection.

The earlier atomic-create and profile-bound findings remain repaired: publication uses an empty-expected `--force-with-lease=refs/heads/<branch>:` CAS with an injected final check→push race regression, while verification enforces registered-profile identity, `max_files`, `max_total_bytes`, allowed/required globs, exact file sizes/digests and manifest. The materialization/publication path still preserves raw bytes, stages only the evidence destination, reconstructs the exact remote base immediately before effect, verifies the resulting remote ref, fixes `human_verdict` to null, and introduces no new notification or human-verdict authority.

The intervening #300 integration is outside this evidence mechanism and this Ready head is rebased on current main. I found no remaining applicable refutation on this exact candidate. Fresh exact-HEAD `CI Gate` remains a separate integration prerequisite.

PR #305, opened 2026-09-11
Prepare exact terminal physical landing semantics

djibian:
Establishes the no-effect prerequisite for the remaining controlled terminal landing slice of #157, on top of the integrated exact-program physical host path.

`PhysicalProgramSequence` now validates the complete currently supported post-takeoff envelope (horizontal move / turn only) at construction time, before reset or takeoff can be attempted. It exposes one opaque terminal-landing claim only after every prior exact motion has causally completed. Definitive no-effect/rejection keeps the same terminal statement pending; an ambiguous emitted landing makes sequencing terminal rather than manufacturing completion; only causal landing completion advances past the final statement.

The new pure `landing_command.py` derives pinned firmware `COMMAND_LAND_WITH_VELOCITY` (10) solely from the teacher-bound canonical AST. Because this bounded physical slice admits no intervening vertical effect, it uses the exact bound absolute takeoff height as a relative downward landing distance, preserves current yaw and fixes the explicit firmware safe-default landing velocity at 0.5 m/s. There is no caller-supplied height, velocity, yaw, raw packet, teacher/session object or CRTP send surface.

The existing `test_physical_program_sequence.py` gate now covers exact terminal reservation/completion identity, caller-substitution rejection, motion/landing ambiguity, eager rejection of unsupported middle statements, and the exact command-10 packet/provenance semantics. It remains on the existing Runtime core CI path; no workflow or authority boundary is added.

This PR deliberately does **not** emit a landing command and does not claim an executable normal-completion path yet. A later trusted host consumer must still compose fresh #278/#249 provenance, #267 teacher authority, #266/#262 session/watchdog state, #257 supervisor evidence, #272 SafeLink, #271 acknowledgement freshness, #273 exclusion, and #264 controlled-landing completion around the exact terminal claim. No motorized checkpoint is requested or authorized.

Refs #157 #264 #276 #289 #295.

Issue #304, opened 2026-09-11
Compose trusted controlled landing for physical program completion


djibian:
## Goal

Extend the integrated #276 physical-host effect path with the separately bounded controlled high-level landing needed to complete the exact teacher-bound physical program safely. This is the next #157 physical-backend slice after the first ordinary move/turn SETPOINT_HL consumer integrated at `main@9b21eb5a7083b3524e47e962864cde9cab333016`.

Refs #157 #276 #264 #295.

## Current boundary

#276 deliberately excludes terminal landing even though #295 requires the canonical physical AST to end with exact `{kind: land}`. The host can currently establish causal takeoff and execute parameter-free exact-sequence move/turn effects, but it cannot consume that validated terminal land as a normal completion effect.

Integrated #264 already provides non-authority landing-completion evidence: one fresh pre-land high-level-flight baseline followed by later same-epoch finished, non-flying, high-level-inactive #257 evidence. The missing work is the trusted effect composition and program-sequence transition around that evidence.

## Required authority topology

Preserve the integrated physical-host trust root rather than creating a parallel landing API:

- terminal landing is derived only from the next exact #295 `PhysicalProgramSequence` element; ordinary caller IPC must not supply altitude, duration, raw packet bytes, bridge/provenance data or landing authority;
- fresh host-owned #278/#249 current-program provenance and the exact #267 teacher binding remain mandatory at the landing effect boundary;
- the same exact #266 powered session, #262 watchdog, #257 supervisor reader, #272 SafeLink, #271 acknowledgement domain and #273 process-wide effect/reset exclusion are reused;
- emit the controlled high-level landing primitive, never high-level `stop()`/motor cut as normal completion;
- install the acknowledgement callback before exactly one no-retry send and preserve the existing ambiguity/recovery rules;
- after accepted emission, consume only the private #279 completion authority together with #264 fresh landing-completion evidence before transitioning the execution domain to inactive/completed state;
- definitive pre-effect/rejected outcomes must not advance the exact program sequence; ambiguous emitted/completion state must poison completion rather than manufacture a retry;
- watchdog/session handling remains active until trusted landing completion is established.

## Deterministic acceptance

- [ ] The actual production physical-host control flow can consume the exact terminal land from a validated #295 program after causal takeoff and zero or more integrated move/turn steps.
- [ ] External IPC remains parameter-free and cannot substitute landing semantics, raw bytes, provenance or authority.
[...]

PR #308, opened 2026-09-11
Compose trusted terminal controlled landing

djibian:
Implements the remaining deterministic #304 trusted-host terminal landing slice on top of integrated #305 semantics.

- adds one trusted command-10 terminal landing transport that reuses the #276 current-program/teacher/session/watchdog/supervisor/SafeLink/acknowledgement/exclusion trust root;
- derives the landing request only from the exact teacher-bound canonical AST; no caller height, velocity, yaw, raw packet or retry surface exists;
- installs the SETPOINT_HL acknowledgement callback before exactly one plain send;
- requires the private accepted-effect completion permit plus #264 same-epoch landing-completion evidence before transitioning #273 to inactive;
- extends the production activation controller so the existing parameter-free `execute-next-inflight` operation consumes the exact terminal land only after all prior exact motions causally complete;
- extends the actual host sequencing regression through terminal landing, including the boundary-only takeoff/land program and malformed-terminal preflight rejection.

This is deterministic fake/injected-hardware coverage only. It authorizes no motorized action and requests no `TEST_REQUIRED` checkpoint.

Refs #304 #157 #305 #276 #264.

### DIFF
--- a/tools/ci/test_physical_controlled_landing_transport.py
+++ b/tools/ci/test_physical_controlled_landing_transport.py
@@ -1,0 +1,300 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+import json
+from pathlib import Path
+import struct
+import sys
+
+ROOT = Path(__file__).resolve().parents[2]
+PHYSICAL = ROOT / "tools" / "physical"
+sys.path.insert(0, str(PHYSICAL))
+
+import controlled_landing_transport as landing_transport  # noqa: E402
+import physical_execution_domain as execution  # noqa: E402
+import serve_reference_capabilities as capability_bridge  # noqa: E402
+import teacher_run_authorization as teacher  # noqa: E402
+import test_physical_setpoint_hl_transport as base  # noqa: E402
+
+
+def require(condition: bool, message: str) -> None:
+    if not condition:
+        raise AssertionError(message)
+
+
+def expect_error(callable_, error_type, pattern: str) -> None:
+    try:
+        callable_()
+    except error_type as exc:
+        require(pattern in str(exc), f"expected {pattern!r} in {exc!r}")
+        return
+    raise AssertionError(f"expected {error_type.__name__} containing {pattern!r}")
+
+
+def canonical_ast() -> str:
+    return json.dumps(
+        {
+            "program": [
+                {"height_m": 0.8, "kind": "takeoff"},
+                {"kind": "land"},
+            ],
+            "semantics": "webeeblocks-ast-v1",
+            "version": 1,
+        },
+        separators=(",", ":"),
+        sort_keys=True,
+    )
+
+
+class TestHostBoundLandingTransport(landing_transport.TrustedControlledLandingTransport):
+    def __init__(self, *, bridge, **kwargs) -> None:
+        self._test_bridge = bridge
+        super().__init__(**kwargs)
+
+    def _read_current_binding(self):
+        binding = self.teacher_binding
+        try:
+            evidence = self._test_bridge.assert_current_program(
+                profile_id=binding.profile_id,
+                ast_binding=binding.ast_binding,
+                connection_epoch=binding.connection_epoch,
+                timeout_seconds=0.25,
+            )
+        except Exception as exc:
+            raise landing_transport.ControlledLandingTransportError(
+                "integrated #278/#249 current-program re-assertion failed"
+            ) from exc
+        if type(evidence) is not capability_bridge.CurrentProgramPreflightEvidence:
+            raise landing_transport.ControlledLandingTransportError(
+                "invalid current-program evidence"
+            )
+        if evidence.execution_authority is not False:
+            raise landing_transport.ControlledLandingTransportError(
+                "current-program evidence became authority"
+            )
+        if (
+            evidence.profile_id != binding.profile_id
+            or evidence.ast_binding != binding.ast_binding
+            or evidence.connection_epoch != binding.connection_epoch
+        ):
+            raise landing_transport.ControlledLandingTransportError(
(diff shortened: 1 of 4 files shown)


djibian:
DRAFT_FINDING 7a8096dae334a1029d6c5070963197ff747e5a05

The production-host sequencing regression now exercises the terminal branch, but the new 216-line trusted `TrustedControlledLandingTransport` itself is still replaced wholesale by `FakePhysicalTransportBase` in that test. I do not see a deterministic test that executes the real landing transport with injected/fake Crazyflie seams.

That leaves the new effect-critical mechanics unfalsified by the candidate tests: exact command-10 bytes from `landing_command`, callback-before-send ordering, exactly one no-retry `send_packet`, definitive rejection restoring `FLYING`, timeout/malformed/disconnect poisoning after emission, positive acknowledgement being insufficient without #264 completion, watchdog liveness through completion, and successful #279 completion to `INACTIVE`. A regression inside any of those new paths could pass the current host test because its fake landing method simply appends an event and sets the domain inactive.

Smallest repair boundary: add a focused deterministic test of the real `TrustedControlledLandingTransport`, reusing the existing `test_physical_setpoint_hl_transport.py` fake Crazyflie/session/ack/supervisor seams where practical, and wire it into the existing Runtime core harness. Cover at least success with exact command-10 request + callback-before-one-send + #264 completion to inactive, definitive rejection/no advance, one ambiguous post-send outcome with no retry, and accepted acknowledgement followed by failed landing-completion evidence forcing recovery. No physical checkpoint is implicated.

djibian:
DRAFT_FINDING 3dc43f4a4dba07dc8dd95bb4f020d71ee3766860

The focused real-transport regression now repairs the earlier test-coverage finding, but the landing effect boundary still has a current-program provenance TOCTOU window.

`send_controlled_landing()` obtains `final_binding = self._read_current_binding()` and validates it, then performs two potentially blocking fresh supervisor operations (`_read_fresh_finished_flying()` and `capture_pre_land_flight()`). After those reads it only re-checks the already-captured teacher binding before SafeLink/ack/send; it does not obtain a fresh #278/#249 current-program assertion again. Therefore the browser/current program can change during either blocking observation and the old teacher-bound command-10 request can still be emitted. That conflicts with #304's requirement that fresh host-owned current-program provenance remain mandatory at the landing effect boundary and that stale current-program state cannot emit.

Smallest repair: after the blocking #257/#264 pre-land observations and immediately before SafeLink/ack/emission, call the host-bound `_read_current_binding()` again, re-assert its authority, and require it to equal the initial binding used to derive the request. Add a deterministic regression that changes the current-program fixture during a pre-land supervisor read and proves no packet is emitted. No physical checkpoint is implicated.

djibian:
NO_GO 266049d1e6aa3eb668e8a2a660d0c7bf684a6b12

The implementation now closes the current-program provenance TOCTOU window: it performs a second host-owned `_read_current_binding()` after the blocking #257/#264 pre-land observations and requires exact equality with the binding used to derive command 10.

The candidate still lacks the deterministic oracle requested by the applicable finding. `test_changed_current_program_blocks_before_landing_effect()` changes the program before the first effect-boundary provenance read; it does not exercise a change *during* `_read_fresh_finished_flying()` / `capture_pre_land_flight()`, which is the race the new second read exists to close. A regression that accidentally removes or moves that second read could therefore remain green.

Smallest repair boundary: in the focused real-transport test, inject one current-program change from a pre-land supervisor read/capture seam after the initial provenance read, then assert that the final provenance reconstruction rejects and `send_calls` stays empty. Keep the execution domain definitively `FLYING`. No physical checkpoint is implicated.

djibian (commented):
GO d10b826c532dd79a456d568598058f4612b5ee27

Independent falsification review against current main@06102e51564036542a3962bf3b96c1b989688a2a, integrated #276/#305, issue #304, and the prior findings on this PR. This execution did not modify #308.

The earlier real-transport coverage gap is resolved: the actual `TrustedControlledLandingTransport` is exercised with the existing trusted-domain fake seams for exact command-10 bytes, one-shot emission, definitive rejection, ambiguous acknowledgement, accepted acknowledgement followed by failed #264 completion, stale current-program rejection, and absence of an importable positive provenance path. That focused regression is executed through the existing physical-host Runtime core test path.

The later current-program TOCTOU finding is also resolved on this exact HEAD. The transport derives the command from an initial fresh host-owned current-program assertion, performs the potentially blocking #257/#264 pre-land observations, then reconstructs `_read_current_binding()` again immediately before SafeLink/ack/emission, re-validates the mutable authority and requires exact equality with the initial binding. The new deterministic race regression changes the current program only after the initial provenance read crosses fresh supervisor preparation and proves the final reconstruction rejects with no packet sent while the execution domain remains `FLYING`.

The production composition preserves the integrated trust root rather than adding a parallel API. The existing parameter-free `execute-next-inflight` operation reserves the exact #305 terminal landing only when it is the next canonical statement; move/turn remains ordered through the same sequence, definitive pre-emission/rejected outcomes do not advance it, ambiguous outcomes poison it, and accepted landing completes the sequence only after the transport has causally established #273 `INACTIVE`. Boundary-only takeoff/land, malformed terminal preflight rejection, and normal move/turn/land ordering are covered through the actual physical-host control path.

No caller-selected landing height, velocity, yaw, raw bytes, provenance or retry surface is introduced; normal completion remains controlled high-level landing rather than STOP/motor cut; watchdog liveness is retained through fresh #264 completion. I found no remaining applicable refutation on this exact Ready HEAD. Canonical exact-candidate `CI Gate` remains a separate integration prerequisite, and this GO authorizes no motorized checkpoint.

PR #309, opened 2026-09-11
Add trusted controlled terminal landing transport

djibian:
Builds the trusted terminal landing effect core required by #304 on top of integrated #305 semantics.

`TrustedControlledLandingTransport` reuses the existing #276 authority topology and exposes one parameter-free `send_controlled_landing()` operation. Inside the process-wide physical effect exclusion it reasserts the host-local current profile/AST/epoch binding, derives the exact command-10 request only through `landing_command.derive_bound_landing_command`, requires fresh finished-flight and #264 pre-land evidence, rechecks teacher/session/watchdog authority and SafeLink, installs the acknowledgement callback before exactly one plain no-retry SETPOINT_HL send, and treats positive acknowledgement only as the private #279 completion permit. Completion reaches `inactive` only after later fresh same-epoch #264 evidence establishes finished, non-flying and high-level-inactive state while the watchdog remains live. Ambiguous acknowledgement/completion stays fail-closed.

The deterministic regression covers exact command bytes/provenance, fresh landing completion, changed-current-program rejection before effect, definitive rejection without retry, ambiguous acknowledgement recovery, and absence of caller-selected landing parameters. It is wired into the existing Runtime core harness.

This Draft is deliberately a bounded effect-core slice: #304 remains open until the production physical host consumes the exact terminal sequence claim through this transport and completes the parameter-free program lifecycle. No motorized checkpoint or real-flight qualification is requested.

Refs #304 #157 #305 #276 #264 #279.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -133,6 +133,10 @@ def main():
     if physical_setpoint_transport_test.returncode:
         print('FAIL trusted one-shot SETPOINT_HL effect transport',file=sys.stderr);print(physical_setpoint_transport_test.stdout,file=sys.stderr);print(physical_setpoint_transport_test.stderr,file=sys.stderr);return physical_setpoint_transport_test.returncode
     print(physical_setpoint_transport_test.stdout.strip())
+    controlled_landing_transport_test=subprocess.run([sys.executable,'tools/ci/test_controlled_landing_transport.py'],text=True,capture_output=True)
+    if controlled_landing_transport_test.returncode:
+        print('FAIL trusted one-shot controlled landing transport',file=sys.stderr);print(controlled_landing_transport_test.stdout,file=sys.stderr);print(controlled_landing_transport_test.stderr,file=sys.stderr);return controlled_landing_transport_test.returncode
+    print(controlled_landing_transport_test.stdout.strip())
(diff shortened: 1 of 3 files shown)


djibian (commented):
NO_GO 2c6cfdc7c28c6a51beb2e835f2626f2a60e15996

The terminal effect still has a current-program provenance TOCTOU after its last host-local #278/#249 assertion. `send_controlled_landing()` reads and validates `final_binding`, derives command 10 from that binding, then performs `_read_fresh_finished_flying()` and `capture_pre_land_flight()`, both potentially blocking fresh supervisor operations. After those reads it only re-validates the already-captured teacher/session/watchdog authority before SafeLink and emission; it never reconstructs the current program again.

A browser/current-program change during either blocking observation can therefore leave the old teacher-bound command bytes eligible for the one-shot send. The existing changed-current-program regression changes the binding before the first assertion and does not falsify this interleaving. That conflicts with #304's requirement that fresh host-owned current-program provenance remain mandatory at the landing effect boundary and that stale current-program state cannot emit.

Smallest repair boundary: after the two pre-land supervisor reads and immediately before SafeLink/ack/emission, obtain `_read_current_binding()` again, re-assert it, and require exact equality with the binding from which the request was derived. Add a deterministic regression that changes the current-program fixture during a pre-land supervisor read and proves `send_packet` is never reached. No physical checkpoint is implicated.

Issue #320, opened 2026-09-12
Apply exact physical set_speed as trusted no-effect run state


djibian:
## Goal

Advance #157 after integrated #316 by consuming the existing backend-neutral `set_speed(speed_m_s)` statement on the trusted physical path as **host-local no-effect run state** for subsequent horizontal moves.

The pure physical timing policy from #268 already models the established Runtime/Webots meaning: `set_speed` changes later horizontal translation only, defaults to 0.35 m/s, and is physically bounded to the currently proven 0.10–0.35 m/s envelope. The physical program sequencer still rejects `set_speed` before takeoff, so that semantic continuity is not yet reachable.

Refs #157 #183 #268 #295 #316 #276.

## Required boundary

- Derive every speed value only from the exact teacher-bound canonical AST and exact #295-style cursor. Ordinary physical-host IPC stays parameter-free and cannot supply speed, index, timing policy, completion or authority.
- Treat `set_speed` as a **no-effect semantic step**. It must emit no cflib/CRTP/SETPOINT_HL, commander, reset or acknowledgement transaction and must not enter the physical-effect ambiguity/retry APIs.
- Preserve the already-established physical timing envelope from #268: default horizontal speed 0.35 m/s and accepted physical `set_speed` values 0.10–0.35 m/s. The backend-neutral AST still admits up to 0.6 m/s; values above the proven physical envelope must therefore fail complete physical-program validation **before reset/takeoff**, not silently clamp or widen the physical command envelope.
- `set_speed` affects subsequent horizontal `move` duration only. It must not alter yaw/turn, takeoff, landing, vertical policy or watchdog timing.
- Keep state per exact activated run; no global/persistent speed survives a new activation.
- Applying the exact state step requires the same causally established `FLYING` run, live #262/#266 watchdog/session/epoch and current #278/#249 program provenance used by the no-effect wait boundary. If application cannot complete under that exact binding, fail the sequence terminal without advancing into a later physical effect.
- Advance the exact cursor once only after the host-local timing policy has accepted the AST-derived value. A consumed speed statement cannot be replayed by duplicate execution.
- Later horizontal effects must continue to reconstruct their own existing provenance/effect/completion boundaries; a completed speed-state step is not physical authority for the later send.

## Deterministic acceptance

- [ ] Canonical `takeoff -> set_speed(0.10) -> move -> land` is admitted before takeoff and consumed in exact order by the actual parameter-free production host path.
[...]

PR #316, opened 2026-09-11
Support exact no-effect physical wait pacing

djibian:
Implements #312 as the next bounded #157 physical-continuity slice after integrated #308.

The exact physical-program sequence now admits the existing Runtime v2 `wait(seconds)` statement inside the top-level takeoff/land envelope. Wait duration is derived only from the teacher-bound canonical AST, preserves the established 0.1–5.0 s semantic envelope, and has a distinct claim/completion path so it cannot be released or classified as an emitted physical effect.

The trusted physical run controller consumes that wait through the existing parameter-free `execute-next-inflight` operation. It keeps the same powered session/connection epoch and #262 watchdog live, reasserts the host-owned current-program binding before and after pacing, uses bounded monotonic host-time slices, emits no cflib/CRTP flight command, and advances the exact cursor only after the full duration completes. Any pacing/watchdog/epoch/provenance failure makes the sequence terminal without skipping into a later effect.

Deterministic regressions cover exact wait bounds and claim identity, interruption without advancement, monotonic/liveness/epoch pacing, caller semantic substitution rejection, and the actual production-host `takeoff -> wait -> land` composition while preserving existing move/turn/landing behavior.

No motorized checkpoint or `TEST_REQUIRED` is requested or authorized.

Closes #312. Refs #157 #295 #308 #262 #267.

PR #295, opened 2026-09-10
Bind physical motion to exact program sequence

djibian:
Introduces the smallest no-effect sequencing primitive needed to reconcile #276 with the completed #287/#293 production takeoff path.

After the trusted takeoff consumer has established the first exact statement and causally advanced the shared physical execution domain to `flying`, `PhysicalProgramSequence` starts at canonical AST index 1 and derives only the exact next top-level horizontal move or turn from the immutable teacher-bound `astBinding`. Its API accepts no caller-selected motion or program index. One opaque pending claim prevents concurrent advancement; causal completion advances exactly one statement, definitive no-effect/rejection leaves the same statement next, and an ambiguous emitted result makes sequencing terminal instead of manufacturing progress.

The primitive reuses the integrated canonical physical AST/takeoff parser and #256 motion bounds, emits no CRTP command, and fails closed on unsupported statements rather than skipping them. Runtime CI now executes deterministic regressions proving exact order, claim identity, caller-substitution rejection, ambiguity handling, malformed/unsupported statement rejection, and canonical takeoff-boundary requirements.

This is intentionally a narrow prerequisite for the existing #276 SETPOINT_HL effect core; it does not add landing, vertical motion, control-flow interpretation, raw packet APIs, or any real/motorized test.

Refs #276 #287 #157.

PR #322, opened 2026-09-12
Apply exact physical set_speed as trusted run state

djibian:
Advances #157 by extending the exact teacher-bound physical sequence with backend-neutral `set_speed(speed_m_s)` as host-local no-effect run state.

The physical sequencer now validates `set_speed` before reset/takeoff against the already-proven #268 physical envelope (0.10–0.35 m/s) while leaving the generic semantic AST ceiling at 0.6 m/s unchanged. Exact speed statements have their own no-effect claim lifecycle and remain outside the physical-effect release/ambiguity APIs.

The production activation path consumes the exact next speed statement with no caller semantic input, requires the established flying/watchdog/epoch/current-program binding, mutates only the per-run `HighLevelTimingPolicy`, revalidates the live binding, then advances the exact cursor. Failure is terminal. Later horizontal moves keep their existing provenance/effect/completion path and receive the same run-local timing policy; turn timing remains independent.

Production-shaped deterministic coverage exercises exact ordering, strict physical bounds, no-effect behavior, selected-speed move duration, turn independence, pre-takeoff rejection, and terminal watchdog-loss behavior.

Refs #320 #157 #268 #295 #316 #276.

### DIFF
--- a/tools/ci/test_physical_high_level_timing.py
+++ b/tools/ci/test_physical_high_level_timing.py
@@ -64,6 +64,11 @@ assert "send_packet" not in source
 assert "takeoff_duration" not in source
 assert "landing_duration" not in source
 
+# Keep the new exact-AST set_speed production-shaped regression on the same
+# already-selected physical timing CI path without broadening workflow authority.
+import test_physical_set_speed_state as speed_state  # noqa: E402
+assert speed_state.main() == 0
+
 print(
     "PASS pure physical timing preserves horizontal speed/yaw-rate ceilings "
     "for serialized rest-to-rest HighLevelCommander trajectories"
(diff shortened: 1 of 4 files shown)


djibian (commented):
Controller Review: a7f79d548505cb6a667e1966100f51b8f7c95828

Verdict: GO
Reviewed HEAD: a7f79d548505cb6a667e1966100f51b8f7c95828
Findings: No P0/P1 finding. The candidate keeps set_speed as exact-AST-derived host-local no-effect state, rejects values outside the established 0.10–0.35 m/s physical timing envelope before reset/takeoff without narrowing the generic AST, preserves caller parameter-freedom and the existing live watchdog/epoch/current-program gates, advances the sequence only after successful local state application, and reuses the per-run timing policy only for subsequent horizontal move timing while turn/takeoff/landing semantics remain uncoupled. The dedicated production-shaped regression is wired into the already-selected physical timing CI path. Exact Ready CI Gate remains separately required before integration.

PR #325, opened 2026-09-12
Make physical host authority regression causal

### LINKED ISSUE #323: Remove caller-teardown race from physical host authority regression
## Problem

Canonical PR #319 run `34676465040` exposed a nondeterministic failure in `tools/ci/test_physical_host_authority_path.py` that is already present on current `main@5c0c4fe42ac4935ab73a75544e012b61a95272ec`.

The `student-ui` job (`103507106577`) ran the shared `run_runtime_v2_core_harness.py` and failed with:

`AssertionError: trusted teacher peer received no post-reset proposal`

In the **same workflow run on the same exact candidate**, `runtime-v2-core` (`103507106549`) ran the same core harness and the same authority-path test successfully. This is therefore a deterministic-test race, not evidence that #319's file-broker changes alter physical authority semantics.

## Causal race

`run_host()` currently receives the successful `validate-run-context` reply and immediately executes `caller_peer.shutdown(socket.SHUT_WR)`. In `serve_physical_host.py`, EOF on that caller socket enters `finally`, sets `host_stopping`, and joins the activation thread.

The activation thread waits for `staged_ready`, then checks `host_stopping` **before** sending the post-reset teacher proposal. Depending on scheduling, caller EOF can therefore set `host_stopping` after staging but before the activation worker runs, causing a normal host teardown and a teacher-side EOF with no proposal.

This explains both outcomes observed on the same SHA: if the activation thread wins the race, the test passes; if caller teardown wins, the teacher receives no proposal.

## Smallest repair

Keep the ordinary caller channel alive until the one-shot teacher peer has completed its proposal/decision exchange in the teacher-enabled test cases, then close the caller write side and let host teardown join the activation worker. The no-teacher case can retain its current immediate caller teardown.

Do **not** weaken the authority oracle, add sleeps, inflate host timeouts, or change production host lifecycle merely to make CI green.

## Acceptance

- [ ] teacher-enabled positive and mismatched-decision cases cannot close caller IPC before the teacher exchange has completed;
- [ ] the test still fails if no teacher proposal is produced by the host;
- [ ] mismatch remains fail-closed and positive authority still reaches the causal effect chain;
- [ ] no-teacher caller-only validation remains effect-free;
- [ ] repeated/shared Runtime v2 core execution is deterministic without timing sleeps;
- [ ] no product/authority semantics change.

Refs #319.

djibian:
Repairs #323 without changing production physical-host semantics.

The shared authority-path regression previously closed ordinary caller IPC immediately after `validate-run-context`. That caller EOF could set `host_stopping` before the already-staged activation worker published its host-first teacher proposal, so identical core harness executions could either pass or fail by scheduling order.

Teacher-enabled cases now wait on the existing one-shot teacher peer's bounded completion before caller teardown. This is causal synchronization, not a delay: no sleep is added, the existing 1 s peer bound is unchanged, and a missing proposal still fails. Caller-only/no-teacher validation retains immediate teardown and remains effect-free.

CI must re-run the shared Runtime v2 core path and preserve mismatch/positive/no-teacher authority assertions.

Fixes #323.

### DIFF
--- a/tools/ci/test_physical_host_authority_path.py
+++ b/tools/ci/test_physical_host_authority_path.py
@@ -230,6 +230,19 @@ def run_host(*, teacher_mode: str | None) -> dict[str, object]:
         },
         "caller reply must remain diagnostic/non-authority",
     )
+
+    # Teacher-enabled cases must let the host's activation worker consume the
+    # staged binding and complete the one-shot proposal/decision exchange before
+    # caller EOF can trigger host teardown. This is a causal test synchronization
+    # point, not a delay: a missing proposal still fails the existing bounded join.
+    if teacher_worker is not None:
+        teacher_worker.join(timeout=1.0)
+        require(not teacher_worker.is_alive(), "teacher peer did not finish one-shot exchange")
+        require(
+            "error" not in teacher_result,
+            "teacher peer failed: " + repr(teacher_result.get("error")),
+        )
+
     caller_peer.shutdown(socket.SHUT_WR)
 
     worker.join(timeout=3.0)
@@ -238,14 +251,6 @@ def run_host(*, teacher_mode: str | None) -> dict[str, object]:
         "error" not in outcome,
         "actual production host runner failed: " + repr(outcome.get("error")),
     )
-
-    if teacher_worker is not None:
-        teacher_worker.join(timeout=1.0)
-        require(not teacher_worker.is_alive(), "teacher peer did not finish one-shot exchange")
-        require(
-            "error" not in teacher_result,
-            "teacher peer failed: " + repr(teacher_result.get("error")),
-        )
 
     caller_peer.close()
     if teacher_peer is not None:


djibian (commented):
GO a2ac0356a9000a41e3a67050d4e85ff3273e0d93

Independent falsification review against current main@7c39df019948ffda93224db3702578f79613f0a8 and issue #323's test-only repair boundary. This execution did not mutate the candidate.

The exact diff changes only `tools/ci/test_physical_host_authority_path.py`: teacher-enabled cases now wait for the existing bounded one-shot teacher peer to consume the host-first proposal and finish its decision exchange before ordinary caller EOF is injected. This removes the scheduling race without a sleep, timeout inflation, weakened assertion, or production-host lifecycle change. A missing proposal still leaves the teacher thread alive at the existing 1 s join and fails; a malformed/mismatched exchange still surfaces through `teacher_result`; only after that causal checkpoint does caller shutdown exercise host teardown. The no-teacher case still reaches caller EOF immediately and remains effect-free.

I tried to falsify the change against the positive, mismatched-teacher, and caller-only paths. The positive path still requires the ordered real-domain events through fresh post-reset provenance, teacher proposal, watchdog activation and transport emission; mismatch still cannot reach watchdog/effect; caller-only still cannot trigger reset or effect. The repair matches #323's acceptance boundary and does not alter product/authority semantics.

I found no remaining applicable refutation on this exact Ready HEAD. Fresh canonical exact-HEAD Ready `CI Gate` remains the separate integration prerequisite; the newest run must settle successfully.

PR #311, opened 2026-09-11
Preserve conditional X3 metric-reference envelopes

djibian:
The integrated X3 collector and descriptive pressure probe still lack a machine-readable external reference for vehicle displacement, surface change and event timing. A point timestamp or a descriptive test label must not erase that reference's uncertainty.

This adds an offline reference processor with hash-bound witness files, explicit metric/event intervals and independently annotated synchronization anchors. Exact JSON decimals feed rational affine-clock constraints without an initial binary-float rounding. The correlated rate/offset polygon rejects inconsistent anchors and extrapolation; one label-independent calculation preserves signed Δz/Δh/Δclearance bounds and uncertain pressure-window starts.

The result remains COMPUTED_CONDITIONAL: physical reference and affine-clock validation false, independent displacement UNPROVEN and no physical verdict. Source separation uses the opened file's device/inode identity, including hard-link rejection. No predictor, new firmware, checkpoint request or publication mechanism is introduced.

Exact candidate: `0eb17c24e8473c4e4295944c1c326c57b6d12877`, containing current `main@2bacdd340e668c606e10df334698db0ba26793ec`. This is a merge-only base refresh of repaired candidate `c8c009a473f2a5c0eea3cc5dd217a00286220a33`; the substantive five-file PR diff is unchanged.

The repair addresses the exact `dfb6ec856487d051cee7223ed11f28a58712771f` finding at https://github.com/djibian/webeeblocks/pull/311#issuecomment-5643238736: file-entry tests now reject the rounded-away interior sync contradiction and preserve a nonzero sub-float displacement. The added CI script wiring has been restored to current main as required by #70's investigation scope; the hard-link repair is retained. This repair account is not an independent GO.

Validation on the repaired substantive candidate covers 14 reference controls via `python3 -m unittest discover -s experiments/crazyflie-ukf-surface-range -p 'test_metric_reference*.py'`; all six existing frozen-pressure controls pass. Exact refreshed-head Draft CI Gate run `34674572292` is green. There is no diff to the main X3 build oracle, .github or tools/ci. Ready CI Gate and independent exact-HEAD review remain separate integration requirements; a Draft run does not authorize integration.

Remaining scientific boundaries include real reference/synchronization validity, sensor producer timing, an independent predictor with uncertainty/latency limits and exact support preparation. No physical result is derived.

Refs #70 #299 #300 #296.

PR #326, opened 2026-09-12
Freeze independent X3 vertical replay predictor

djibian:
Refs #70

## Result

This adds the next bounded #70/X3 offline component after the integrated collector, pressure probe and conditional metric-reference processor.

- `frozen_vertical_predictor.py` consumes only the exact continuous barometer/IMU capture plus the hash-bound conditional metric-reference timing result; pose/range, `stateEstimate.z`, `stateEstimate.vz`, S3 diagnostics and estimator attitude are forbidden predictor inputs.
- The barometer arm preserves the pre-registered B−/B+ windows and all admissible event/timing intervals without midpoint selection, with drift fitted only on guaranteed prior stationary calibration.
- The inertial arm uses raw body-frame accelerometer samples and pre-declared deterministic specific-force, tilt, initial-velocity, intersample-acceleration and timing bounds. The explicit intersample bound covers continuous acceleration deviation that retained rows and a gap limit alone cannot prove.
- Timing sensitivity includes IMU samples that bracket off-grid admissible boundaries, so interpolation at an admissible boundary cannot depend on an acceleration excluded from the bound.
- Barometer and inertial intervals are combined only by intersection; disagreement stays `DISJOINT` rather than being tuned away.
- Cross-stream device-time alignment uses a signed modular offset around the first barometer timestamp, so the collector-valid case where the first 10 ms IMU callback precedes the first 20 ms barometer callback is retained while exact half-period ambiguity and non-increasing aligned time remain rejected.
- Metric-reference validation flags must be actual JSON booleans; malformed truthy values fail closed rather than being coerced to validation.
- The latency upper bound includes the retained event-end interval width as well as the fixed B+ delay/window and declared timing/delivery allowances.
- The output reports the pre-registered 5 cm half-width / 1 s latency budgets only as conditional checks. `declared_bounds_validated:false`, `sensor_producer_timing_validated:false`, `independent_displacement_verdict:UNPROVEN` and `physical_verdict:null` remain explicit.
- Standalone synthetic controls cover signed replay, bound widening, sensor disagreement, negative-prefix IMU/barometer alignment, calibration/input rejection, interval timing without midpointing, off-grid timing-bracket coverage, intersample displacement allowance, widened-end latency, strict validation-flag typing, exact source/digest binding and rejection of pose/ToF/estimator sources.
- `run_s3_build_oracle.sh` remains exactly current `main`; these offline predictor component tests do not expand the firmware/CI evidence path. Canonical Ready `CI Gate` remains the separate integration prerequisite.

Exact candidate: `c7144876d0a2c7a32f17abb451185cfa91e26866`, based on `main@cf1a430af4d75e28e64b41f4b940e80abbba042a`.

The earlier implementation's local component tests and `py_compile` passed before its original publication. No standalone component execution is claimed for this exact candidate by this description; Draft run #1186 completed only `CI Gate (Draft)` with product suites skipped, which is not integration evidence.

## Boundary

This freezes replay mechanics only. It does not validate the declared error/tilt/intersample/timing bounds, physical-reference annotations, affine-clock assumption or sensor-producer timing. It does not estimate terrain truth, flash firmware, change #251 parameters, tune S3/ToF/barometer, modify Runtime/controller behavior, request `TEST_REQUIRED`, or authorize any motorized test. A later checkpoint is eligible only after the complete exact support/bound provenance/reference/publication path is prepared under `AGENTS.md`.

### DIFF
--- a/experiments/crazyflie-ukf-surface-range/FROZEN_VERTICAL_PREDICTOR.md
+++ b/experiments/crazyflie-ukf-surface-range/FROZEN_VERTICAL_PREDICTOR.md
@@ -38,6 +38,7 @@ The input schema is `webeeblocks.x3.vertical-predictor-input.v1`:
     "specific_force_error_g": 0.0,
     "max_body_z_tilt_deg": 0.0,
     "initial_velocity_error_m_s": 0.0,
+    "intersample_acceleration_error_m_s2": 0.0,
     "barometer_displacement_error_m": 0.0,
     "sensor_time_error_s": 0.0,
     "delivery_latency_error_s": 0.0
@@ -50,6 +51,8 @@ physical bounds**. Every bound must be justified and frozen before a confirmatio
 capture; event outcomes must never be used to narrow it. The result always keeps
 `declared_bounds_validated:false` and `sensor_producer_timing_validated:false`.
 A controller or human verdict must validate the bound provenance separately.
+The metric-reference validation flags are accepted only as actual JSON booleans;
+malformed truthy values fail closed rather than being coerced into validation.
 
 The supplied metric-reference result must be
 `webeeblocks.x3.metric-reference-result.v1 / COMPUTED_CONDITIONAL`, and its
@@ -87,11 +90,22 @@ event interval, then the resulting acceleration envelope is integrated from an
 initial velocity interval
 `[-initial_velocity_error_m_s, +initial_velocity_error_m_s]`.
 
+Retained rows alone do not prove the continuous acceleration between samples.
+`intersample_acceleration_error_m_s2` is therefore an external deterministic
+bound on the absolute difference, at every instant in the admissible replay span,
+between the actual world-vertical acceleration and the piecewise-linear bounded
+acceleration reconstructed from retained rows. For an admissible span of duration
+`T`, the inertial displacement interval is widened by
+`0.5 * intersample_acceleration_error_m_s2 * T²` on both sides. This is deliberately
+conservative and remains conditional on the declared intersample bound being
+valid; the capture gap limit alone is not treated as such a bound.
+
 For uncertain event/log timing, the integration anchor is the **lower supplied
 endpoint**, never a hidden midpoint. A conservative timing-sensitivity allowance
 covers the remaining start/end interval and declared sensor-time error using the
-maximum acceleration envelope over the admissible span. This is conditional
-arithmetic, not proof that Crazyflie log timestamps are sensor-producer times.
+maximum acceleration envelope over the admissible span, including rows needed to
+bracket off-grid admissible boundaries. This is conditional arithmetic, not proof
+that Crazyflie log timestamps are sensor-producer times.
 
 This deliberately does not integrate estimator attitude or `stateEstimate.vz`.
 A later candidate may use a more informative raw-gyro/attitude model only if its
@@ -109,8 +123,10 @@ The result reports two preregistered budget checks only:
 
 - `conditional_half_width_within_5cm`: the overlapping interval has half-width at
   most 5 cm, conditional on all declared bounds being valid;
-- `conditional_latency_within_1s`: the fixed B+ endpoint (0.75 s after the
-  transition) plus declared timing/delivery allowances is at most 1 s.
+- `conditional_latency_within_1s`: relative to the earliest admissible event end,
+  the upper latency includes the complete retained event-end interval width, the
+  fixed B+ acquisition endpoint (`POST_DELAY_S + WINDOW_S = 0.75 s`), both declared
+  sensor-time allowances and the declared delivery allowance, and is at most 1 s.
 
 These are **not** physical PASS/FAIL. In particular, interval half-width is not an
 error measurement against the external metric reference, and the latency is not
@@ -129,18 +145,20 @@ The file entry point hash-binds all three inputs, refuses path escape and existi
 output overwrite, aligns IMU to the first barometer Crazyflie log timestamp, and
 fails closed on malformed/non-finite data, clock reversal, barometer gaps over
 40 ms or IMU gaps over 50 ms. The synthetic controls verify signed replay,
-conditional-bound widening, sensor disagreement, interval timing without
-midpoint selection, source/digest binding and rejection of optimistic/invalid
-inputs. They are component tests, not physical evidence.
+conditional-bound widening, sensor disagreement, negative-prefix clock alignment,
+interval timing without midpoint selection, off-grid timing brackets, malformed
+validation flags, intersample displacement allowance, widened-end latency,
+source/digest binding and rejection of optimistic/invalid inputs. They are
+component tests, not physical evidence.
 
 ## Remaining boundary
 
 This component freezes the sensor predictor mechanics but does **not** validate
-its declared bounds, physical-reference annotations, affine clock assumption or
-sensor-producer timing. It also does not estimate terrain Δh, validate the mixed
-case against a physical reference, publish raw physical evidence, prepare a human
-checkpoint, flash firmware, retune #251, alter Runtime/controller behavior or
-authorize motorized testing.
+its declared error, tilt, intersample or timing bounds, physical-reference
+annotations, affine clock assumption or sensor-producer timing. It also does not
+estimate terrain Δh, validate the mixed case against a physical reference, publish
+raw physical evidence, prepare a human checkpoint, flash firmware, retune #251,
+alter Runtime/controller behavior or authorize motorized testing.
 
 Before a new `TEST_REQUIRED` can be justified, the exact props-off support must
 bind this predictor to the retained acquisition/reference inputs, pre-register
--- a/experiments/crazyflie-ukf-surface-range/FROZEN_VERTICAL_PREDICTOR.md
+++ b/experiments/crazyflie-ukf-surface-range/FROZEN_VERTICAL_PREDICTOR.md
@@ -1,0 +1,148 @@
+# Frozen conditional X3 vertical replay predictor
+
+This is the next bounded offline component for #70 after the integrated input
+collector, frozen pressure probe and conditional metric-reference processor. It
+answers a narrower question than the firmware problem: **given pre-declared
+error/timing bounds, do the raw barometer and raw accelerometer admit one
+consistent vehicle vertical-displacement interval?** It does not classify
+terrain, change the Crazyflie estimator, or establish physical flight capability.
+
+The predictor is intentionally independent of the suspect ToF path. Its file
+entry point accepts exactly three retained sources: the continuous barometer CSV,
+the continuous IMU CSV and the exact `metric_reference.py` result. Pose, range,
+`stateEstimate.z`, `stateEstimate.vz`, S3 diagnostics and estimator attitude are
+not predictor inputs. The metric-reference result is used for calibration/event
+time intervals only; the predictor does not use its reference Δz/Δh values to
+construct its sensor prediction.
+
+The pinned #251 capture format comes from Crazyflie firmware 2026.08
+`54f31e243a0b28b67efef5ba20dbb6d9890a5478`. In that source `acc.x/y/z` are
+body-frame acceleration in G and `baro.asl` is altitude in metres. The collector
+still retains gyroscope and pose diagnostics, but this smallest replay does not
+silently turn estimator attitude into independent evidence. Instead it carries a
+pre-declared bound on the angle between body Z and world vertical.
+
+## Frozen model
+
+The input schema is `webeeblocks.x3.vertical-predictor-input.v1`:
+
+```json
+{
+  "schema": "webeeblocks.x3.vertical-predictor-input.v1",
+  "sources": {
+    "barometer": {"path": "barometer.csv", "sha256": "<64 hex>"},
+    "imu": {"path": "imu.csv", "sha256": "<64 hex>"},
+    "metric_reference": {"path": "metric-reference.json", "sha256": "<64 hex>"}
+  },
+  "declared_bounds": {
+    "specific_force_error_g": 0.0,
+    "max_body_z_tilt_deg": 0.0,
+    "initial_velocity_error_m_s": 0.0,
+    "barometer_displacement_error_m": 0.0,
+    "sensor_time_error_s": 0.0,
+    "delivery_latency_error_s": 0.0
+  }
+}
+```
+
+The zero values above are syntax examples, **not recommended or validated
+physical bounds**. Every bound must be justified and frozen before a confirmation
+capture; event outcomes must never be used to narrow it. The result always keeps
+`declared_bounds_validated:false` and `sensor_producer_timing_validated:false`.
+A controller or human verdict must validate the bound provenance separately.
+
+The supplied metric-reference result must be
+`webeeblocks.x3.metric-reference-result.v1 / COMPUTED_CONDITIONAL`, and its
+recorded barometer-capture digest must match the exact supplied barometer CSV.
+This prevents combining event timing from one capture with sensor data from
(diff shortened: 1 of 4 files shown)


djibian (commented):
DRAFT_FINDING 99f8e74bd2f43277061a2c273e02a1d8b4568b8c

Two bounded findings remain before this Draft can be offered as a stable candidate.

1. `run_s3_build_oracle.sh` now invokes `test_frozen_vertical_predictor.py`. That is the same CI/evidence-path expansion already rejected during #311: the current #70 architecture boundary explicitly keeps this offline X3 work out of CI/governance changes. A path-scoped experiment selection does not make the new invocation absent. Smallest repair: restore `run_s3_build_oracle.sh` exactly to current main and keep the predictor/component tests standalone; the normal required canonical `CI Gate` remains the separate integration prerequisite.

2. `unwrap_rows()` aligns the IMU stream to the first barometer timestamp by unsigned modulo and rejects any resulting offset >= half the 24-bit period. The integrated collector starts separate log blocks (`barometer` at 20 ms and `imu` at 10 ms); a valid first IMU callback can therefore precede the first barometer callback by a few milliseconds. Such a retained capture is currently rejected as `source cannot be aligned to the barometer clock origin`. Smallest repair: use a signed modular offset around the barometer origin (while still rejecting genuinely ambiguous/backward spans) and add a deterministic regression with, for example, IMU first at origin−10 ms and barometer first at origin.

Neither finding implies a physical verdict or checkpoint, and neither justifies firmware retuning or motorized testing.

djibian (commented):
NO_GO 99f8e74bd2f43277061a2c273e02a1d8b4568b8c

The two exact-candidate findings recorded while this HEAD was Draft remain unresolved after it returned Ready:

1. Remove the `test_frozen_vertical_predictor.py` invocation from `run_s3_build_oracle.sh` and restore that oracle to current `main`. The #70/#311 boundary does not authorize expanding this firmware/CI evidence path for the offline X3 predictor; canonical `CI Gate` remains required independently.
2. Make cross-stream timestamp alignment accept the collector-valid case where the first 10 ms IMU callback precedes the first 20 ms barometer callback by a small amount. Use signed modular offset around the barometer origin and add a deterministic negative-prefix regression, while preserving backward/ambiguous-span rejection.

Smallest repair is limited to those two points. No firmware retuning, physical test, Runtime/controller change or checkpoint follows from this verdict.

djibian (commented):
GO b7e74bb1856378d78ad8fc4dceb056a28eb895bd

Independent falsification review against current main@fe625d093a0f5000945475a68f6a196985e26333. This execution did not mutate the candidate branch or PR state.

The two exact-head refutations on 99f8e74 are repaired without broadening scope. `run_s3_build_oracle.sh` is no longer in the PR diff and is therefore restored to current main; the offline predictor remains a standalone experiment component rather than being inserted into the firmware/CI evidence path. Cross-stream alignment now maps device timestamps to the signed modular interval around the barometer origin, explicitly rejects the exact half-period ambiguity, preserves the existing per-stream duplicate/backward/gap checks, and additionally rejects non-increasing aligned time. The deterministic regression shifts the IMU stream by exactly -10 ms and verifies the retained negative prefix and monotone crossing through the barometer origin.

I also checked the remaining changed-file boundary: only the frozen predictor documentation, implementation, and standalone synthetic tests differ from main. The implementation continues to forbid pose/ToF/estimator sources at the file entry point, hash-binds the barometer/IMU/reference inputs, keeps declared physical/timing bounds explicitly unvalidated, reports disagreement as `DISJOINT`, and leaves `independent_displacement_verdict` as `UNPROVEN` with `physical_verdict:null`. I found no path from these offline calculations to firmware, Runtime/controller authority, tuning, or motorized testing.

No standalone component execution is claimed by this review; the candidate deliberately keeps that component test out of the firmware CI oracle. Fresh canonical exact-HEAD Ready `CI Gate` remains the separate integration prerequisite. I found no remaining applicable refutation on this exact Ready HEAD.

djibian (commented):
NO_GO b7e74bb1856378d78ad8fc4dceb056a28eb895bd

The two earlier findings on `99f8e74` are resolved on this exact candidate: the S3 firmware oracle is back out of the PR diff, and signed modular IMU/barometer alignment now has the negative-prefix regression.

One bounded correctness finding remains in the new predictor. `imu_segment()` says its `timing_sensitivity_error_m` conservatively covers every admissible start/end shift, but `max_acc` is computed only from sample rows whose timestamps lie inside `[outer_start, outer_end]`. `integrate_projected()`/`interpolate_projected()` can use the immediately outside bracketing row to define acceleration at an off-grid admissible boundary. A large acceleration in that outside bracket can therefore affect the displacement added/removed by a sensor-time shift while being excluded from `max_acc`, making the timing expansion too small and allowing `delta_z_interval_m` to fail to enclose an admissible replay.

Smallest repair: compute the timing acceleration bound over the actual admissible interpolated span (for example boundary-interpolated points at `outer_start`/`outer_end` plus all interior rows), or conservatively include the bracketing rows; add a deterministic off-grid regression where the immediately outside bracket has the dominating acceleration and verify the timing envelope widens enough. Keep the standalone component-test / no firmware-CI-path boundary unchanged.

This is only a conditional-envelope arithmetic finding. It does not establish any physical verdict, justify tuning, or authorize a test/checkpoint.

djibian (commented):
DRAFT_FINDING b059e69ff4fde56256dbc868233e82c178d4c09c

The arithmetic repair for the exact-head `NO_GO b7e74bb...` is directionally conservative: `max_acc` now ranges over all bracketed `samples` returned for the admissible outer interval, so an immediately outside row used by boundary interpolation can no longer dominate acceleration while being excluded from the timing-sensitivity bound.

One requested falsification control is still absent on this exact Draft HEAD. `test_frozen_vertical_predictor.py` is unchanged from `b7e74bb...`; there is no deterministic off-grid event/timing case where the immediately outside IMU bracket has the dominating acceleration and the assertion proves the resulting timing envelope encloses/widens for that case. Add that bounded regression before offering a new Ready candidate. Keep the existing standalone-test / no firmware-CI-path boundary unchanged.

The PR body also still names `b7e74bb...` as the exact candidate; refresh exact-candidate prose when the Draft stabilizes. Neither item implies a physical verdict, tuning, checkpoint, or motorized test.

djibian (commented):
NO_GO e35faec3803080989d8af136c8a417adba8875db

The exact-head timing-envelope refutation from `b7e74bb...` is repaired on this candidate: the acceleration bound now conservatively includes retained bracketing IMU samples, the deterministic off-grid regression exercises a dominating outside bracket, the PR body names this exact candidate, and the change set remains limited to the standalone X3 predictor/docs/tests.

One fail-closed provenance defect remains. `analyze()` emits `physical_reference_validated` and `affine_clock_validated` as `bool(reference.get(..., False))`. That coerces malformed non-empty values such as the JSON string `"false"` to Python `True`, so a hash-bound but malformed metric-reference result can make the predictor output claim a validation that the input did not actually express. This conflicts with the component's explicit boundary that it never certifies physical reference or clock validation.

Smallest repair: require both metric-reference validation fields to be actual JSON booleans (or otherwise preserve only an exact boolean value without truthiness coercion), fail closed on malformed types, and add a deterministic regression proving e.g. string `"false"` cannot become output `true`. Keep the current standalone/no-firmware-CI-path boundary unchanged.

This is provenance/type validation only. It does not establish a physical verdict, justify retuning, request `TEST_REQUIRED`, or authorize motorized testing.

djibian (commented):
GO e35faec3803080989d8af136c8a417adba8875db

Independent falsification review against current main@fe625d093a0f5000945475a68f6a196985e26333. This execution did not mutate the candidate branch or PR state.

The exact-head timing-envelope refutation on `b7e74bb...` and the follow-up Draft finding on `b059e69...` are repaired on this Ready candidate. `imu_segment()` now derives `max_abs_vertical_accel_bound_m_s2` from the complete bracketed sample set returned for the admissible outer interval, so acceleration used by off-grid boundary interpolation cannot be omitted merely because its source row lies immediately outside `[outer_start, outer_end]`. The new deterministic regression puts the dominating acceleration in exactly such an outside bracket (`t=0.99 s` for admissible `outer_start=1.0 s`), verifies that the bound observes `10*G`, and verifies that the resulting timing-expanded displacement interval encloses replay over the admissible outer endpoints.

The earlier exact-head findings remain repaired: the S3 firmware/CI oracle is absent from the diff; signed modular IMU/barometer alignment retains a valid negative IMU prefix while rejecting ambiguous/non-increasing alignment. The current diff remains limited to the frozen predictor documentation, implementation, and standalone synthetic controls. Inputs remain hash-bound and exclude pose/ToF/estimator state; declared physical/timing bounds remain explicitly unvalidated; disagreement remains `DISJOINT`; `independent_displacement_verdict` remains `UNPROVEN` and `physical_verdict:null`. I found no path to firmware, Runtime/controller authority, tuning, checkpoint creation, or motorized testing.

No standalone component execution is claimed by this review. Fresh canonical exact-HEAD Ready `CI Gate` remains the separate integration prerequisite. I found no remaining applicable refutation on this exact Ready HEAD.

djibian (commented):
DRAFT_FINDING e35faec3803080989d8af136c8a417adba8875db

A second bounded inertial-model issue is independent of the validation-flag type defect already recorded on this exact head. The documentation calls the inertial result a deterministic vehicle-displacement interval and applies `specific_force_error_g` only at retained raw accelerometer samples. `integrate_projected()` then trapezoid-integrates those sample envelopes, while the only between-row constraint is a maximum timestamp gap. A gap bound does not bound acceleration between samples: two captures can have identical sampled accelerations (and satisfy every current declared bound) while an unsampled acceleration excursion between rows changes velocity/displacement outside the computed interval. There is currently no jerk, intersample acceleration, quadrature, or integrated-displacement allowance that closes that gap.

Smallest repair for a physical conditional interval: freeze an explicit deterministic intersample/integration-error model or allowance, propagate it into `delta_z_interval_m`, and add a synthetic falsification case with identical retained rows but admissible between-sample motion that would escape the old interval. Alternatively, if the intended object is only a piecewise-linear log replay, narrow the claims and budget checks so it is not treated as a conditional physical vehicle-displacement bound.

This does not validate any physical bound or justify a checkpoint/test; it is exactly the missing condition required before the inertial arm can support the later 5 cm physical argument.

djibian (commented):
DRAFT_FINDING e35faec3803080989d8af136c8a417adba8875db

A third bounded arithmetic issue affects the advertised conditional latency budget. The barometer B+ calculation correctly preserves the full admissible event-end interval: its latest required window ends at `end_device_s[1] + POST_DELAY_S + WINDOW_S + sensor_time_error_s`. But `conditional_latency_upper_s` is reported as only `POST_DELAY_S + WINDOW_S + 2*sensor_time_error_s + delivery_latency_error_s`; it omits the width `end_device_s[1] - end_device_s[0]` of the retained end-time interval.

Therefore, when the metric-reference clock mapping leaves a non-zero end interval, the offline replay may need samples later than the reported upper latency relative to the earliest admissible transition. The existing interval-timing regression widens event endpoints but does not assert latency, so this under-reporting is currently untested.

Smallest repair: include the event-end interval width in the per-event worst-case latency calculation (along with the existing sensor/delivery allowances) and add a regression showing a widened `end_device_s` widens `conditional_latency_upper_s` and can flip the 1 s budget check. If a different latency origin is intended, define it explicitly and make the calculation/proof match that origin.

This remains a conditional offline budget issue only; it does not create a physical verdict or authorize `TEST_REQUIRED`.

djibian (commented):
DRAFT_FINDING 85ca0485608886e886829c7a38ebfcc78d8d30a6

This head only merges current `main@cf1a430af4d75e28e64b41f4b940e80abbba042a`; the predictor/test files are unchanged from `e35faec...`. Therefore the three unresolved exact-source findings remain applicable before a new Ready candidate:

1. Fail closed on `physical_reference_validated` / `affine_clock_validated` types instead of `bool(...)` truthiness, with a malformed-string regression.
2. Either add and propagate a deterministic intersample/integration-error bound for the inertial arm (with a falsification control), or explicitly narrow the result to a piecewise-linear logged-sample replay rather than a physical displacement interval.
3. Make `conditional_latency_upper_s` cover the retained `end_device_s` interval width (or define/prove another latency origin), with a widened-end regression.

The earlier signed alignment, S3-oracle boundary and off-grid timing-bracket repairs remain intact. The base is now current main, but the PR body still names the previous exact candidate/base and should be refreshed only after the Draft stabilizes. None of these findings authorizes physical testing or `TEST_REQUIRED`.

djibian (commented):
DRAFT_FINDING 907114b5e48c1b66f655be28f2253c82b026a289

The three earlier unresolved findings are partly repaired in the implementation on this Draft HEAD, but the candidate is not yet internally coherent or reviewable as complete.

1. `physical_reference_validated` / `affine_clock_validated` now fail closed on non-boolean input and a malformed-value regression exists; that finding is resolved.
2. `intersample_acceleration_error_m_s2` is now required and propagated into the inertial displacement interval, and the end-time interval width is included in `conditional_latency_upper_s`. However, the standalone tests and documentation have not been updated to match those new semantics. The shared `fixture()` still omits `intersample_acceleration_error_m_s2`, and the direct `imu_segment()` regression prep omits it too, so current component tests will raise before exercising their intended assertions. There is also no dedicated regression yet proving the new intersample allowance encloses the falsifying between-sample case, nor one proving widened `end_device_s` widens/flips the latency budget as previously requested.
3. `FROZEN_VERTICAL_PREDICTOR.md` still documents the old input schema without the intersample bound and still describes latency as a fixed 0.75 s endpoint plus allowances, omitting the retained end-time interval width. That documentation no longer matches the code.

Smallest repair: add the new bound to the documented schema and every test fixture/prep, add one deterministic intersample-envelope regression and one widened-end latency regression, refresh the explanatory text and exact-candidate/body prose, then offer a new Ready HEAD. Keep the standalone/no-firmware-CI-path boundary unchanged. No physical verdict or TEST_REQUIRED follows.

djibian (commented):
GO c7144876d0a2c7a32f17abb451185cfa91e26866

Independent falsification review against current `main@cf1a430af4d75e28e64b41f4b940e80abbba042a`. This execution did not mutate the candidate branch or PR state.

All still-applicable findings from the earlier heads are resolved on this exact Ready candidate. The firmware/S3 oracle is absent from the three-file diff; signed modular IMU/barometer alignment retains the valid negative-prefix case; off-grid timing sensitivity includes the bracketing acceleration and has its deterministic regression; metric-reference validation flags now require actual JSON booleans. The later inertial-model finding is addressed by an explicit externally declared `intersample_acceleration_error_m_s2` whose semantics are now documented as a bound on actual world-vertical acceleration versus the reconstructed piecewise-linear bounded trace over the whole admissible span, propagated conservatively as `0.5*A*T^2` on both sides. The standalone control with identical retained rows verifies the displacement allowance. The latency bound now includes the full retained `end_device_s` interval width and the widened-end regression crosses the 1 s budget as intended.

The documentation, input schema, implementation, tests and PR body are coherent on this head. The scope remains fail-closed and offline: pose/ToF/estimator state are forbidden predictor inputs, exact sources are hash-bound, sensor disagreement remains `DISJOINT`, declared physical/timing/intersample bounds remain unvalidated, and outputs remain `independent_displacement_verdict:UNPROVEN` / `physical_verdict:null`. I found no path to firmware mutation, Runtime/controller authority, tuning, checkpoint creation or motorized testing.

No standalone component execution is claimed by this review. The newest canonical Ready workflow run #1188 for this exact HEAD is still in progress at review time, so this GO is not CI evidence and does not authorize integration without the required successful final `CI Gate`. I found no remaining applicable refutation on this exact Ready HEAD.

PR #319, opened 2026-09-12
Implement Firefox direct project files through local WWI broker

djibian:
Advances #87 from the established #317 Qt/controller continuity prerequisite to the product file path.

This slice keeps the existing project manager semantics and adds a local WWI transport for browsers without File System Access. Chrome keeps its existing direct native-picker transport. On the Linux desktop Webots path, Firefox uses a Qt6 native file broker co-located with the Runtime v2 controller.

The broker exposes only session-scoped opaque references to the browser: JavaScript never supplies an arbitrary filesystem path. Open returns bytes plus an opaque reference, but the existing manager adopts that target only after full `.wbb` validation. Save As adopts a new opaque reference only after the native atomic write commits. Save reuses the exact existing reference with no dialog or generated copy; a missing or unavailable target fails explicitly. Cancellation remains neutral and errors preserve the current project/target.

The runtime integration consumes only `WEBEEBLOCKS_FILE_BROKER_V1` WWI traffic and forwards Runtime v2 traffic unchanged. The Qt provider is lazy, so the established Chrome Runtime path does not create a file dialog provider unless broker traffic is requested. The implementation uses the QFileDialog/QApplication continuity established by #317 rather than restarting the historical research harness.

Windows Firefox qualification remains a separate later boundary; the current Windows classroom release path is unchanged.

Local evidence before publication:
- native broker fake-provider C++ contract: PASS for capabilities, Open, Save As, same-reference Save, cancellation, explicit errors and no browser-supplied path;
- WWI JavaScript transport request/response contract: PASS;
- syntax checks for the transport, project UI and extended project-file test: PASS.

Canonical CI must still compile/link the Qt integration against official Webots R2025a and exercise the unchanged Runtime/Chrome path before this candidate is stable.

Refs #87 #317.

PR #327, opened 2026-09-12
Refresh Firefox project UI cache key

djibian:
Fix a deployment boundary left by #319: `project_ui.js` gained the Firefox WWI/native-broker path, but `blockly_v2.html` retained the pre-existing `rev=20260830-1` cache key. A browser profile that had already cached the old project UI could therefore keep executing the pre-#319 unsupported-Firefox code even with the new broker present.

This slice changes only the project UI cache key so the integrated #319 logic is fetched as a distinct resource. It does not change project semantics, the broker protocol, Chromium behavior, Windows qualification, or the human-checkpoint mechanism.

Refs #87 #319.

### DIFF
--- a/plugins/robot_windows/blockly_v2/blockly_v2.html
+++ b/plugins/robot_windows/blockly_v2/blockly_v2.html
@@ -66,6 +66,6 @@
   <script src="main.js"></script>
   <script src="physical_preflight_runtime.js"></script>
   <script src="classroom_fixes.js"></script>
-  <script src="project_ui.js?rev=20260830-1"></script>
+  <script src="project_ui.js?rev=20260912-1"></script>
 </body>
 </html>


djibian (commented):
GO 39f3a93c8905b811198bdbe9f35d967e6f5b824e

Independent falsification review against current main@fe625d093a0f5000945475a68f6a196985e26333. This execution did not mutate PR #327 or its candidate.

The exact Ready diff is one deployment-only change: the sole `project_ui.js?rev=` load in `blockly_v2.html` moves from `20260830-1` to `20260912-1`. That gives the already-integrated #319 project UI a distinct resource URL for profiles that may retain the old script, without changing the script itself, broker protocol, project semantics, Chromium path, Windows qualification, or the #87 human-evidence boundary. I found no second project-UI revision reference that would leave this page loading the stale URL.

I tried to falsify the scope against #87: this change does not claim or manufacture real Linux Firefox end-to-end acceptance, does not close #87, and does not infer Windows Firefox support. It only removes a plausible stale-resource deployment path for the integrated broker code.

Canonical `.github/workflows/ci.yml` run #1177 is associated with PR #327 at exact HEAD `39f3a93c8905b811198bdbe9f35d967e6f5b824e`, completed successfully, and its final job is exactly `CI Gate` with success. I found no applicable refutation on this exact Ready candidate.

PR #329, opened 2026-09-12
Bound Firefox broker reference lifetime

djibian:
## Why

The Linux Firefox Qt/WWI broker keeps every opaque Open/Save As reference in a session map capped at 128. The project manager previously never released references when an opened project was rejected by validation or when a later Open/Save As replaced the current target, so a long-lived session could eventually fail with `REFERENCE_LIMIT` despite retaining only one logical project target.

## Change

- add a protocol-v1 `RELEASE <opaque-ref>` session operation and advertise `referenceRelease:true`;
- remove released tokens from the Qt provider map without touching file contents;
- make the WWI transport use release only when the broker advertises the capability, preserving compatibility with older brokers;
- release rejected candidate references and superseded project targets in the manager while keeping browser-native handles a no-op cleanup path;
- preserve same-file `Save` semantics and neutral cancellation/errors;
- refresh the cache keys for the changed project-file modules;
- extend the broker protocol regression for release and malformed tokens.

Local bounded checks were run on product patch commit `a3e73c33c9f92f829621c59aec6a06442b9ba6a8`:

- `g++ -std=c++17 -Wall -Wextra -pedantic ... file_broker.cpp test_firefox_file_broker.cpp` + executable PASS;
- `node --check` on `project_file_wwi_transport.js`, `project_files.js`, and `project_ui.js`.

Current candidate `c795fdcbce7b492a3e8d08df929ae95f4a7740e3` adds only the merge of current `main@f33cebc17ad4ab0e382caede2a51b1bdb167752f`; the PR diff remains the same eight Firefox/reference-lifecycle files.

This does not claim real Linux Firefox end-to-end acceptance or Windows Firefox qualification.

Refs #87 #319 #327.

### DIFF
--- a/controllers/crazyflie_runtime_v2/file_broker.cpp
+++ b/controllers/crazyflie_runtime_v2/file_broker.cpp
@@ -169,7 +169,7 @@ bool FileBroker::handleMessage(const char *message, std::string *response) {
     const std::string provider = safeProviderName(mProvider->name());
     *response = responsePrefix(id) +
       "CAPABILITIES {\"protocol\":1,\"provider\":\"" + provider +
-      "\",\"providerInjectable\":true,\"operationsReady\":true,\"sameFileSave\":true,\"canonicalExtension\":\".wbb\"}";
+      "\",\"providerInjectable\":true,\"operationsReady\":true,\"sameFileSave\":true,\"referenceRelease\":true,\"canonicalExtension\":\".wbb\"}";
     return true;
   }
   if (operation == "OPEN" && parts.size() == 2) {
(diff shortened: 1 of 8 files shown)


djibian (commented):
GO c795fdcbce7b492a3e8d08df929ae95f4a7740e3

Independent falsification review against current `main@f33cebc17ad4ab0e382caede2a51b1bdb167752f`. This execution did not mutate PR #329.

The bounded claim holds on this exact Ready head. `RELEASE` is a non-file-effect protocol operation, validates the opaque token, is advertised explicitly through `referenceRelease`, and removes only the provider's session reference. The browser transport gates cleanup on the advertised capability for compatibility with older brokers and bounds only the cleanup acknowledgement. Project-manager cleanup releases rejected candidate references and superseded Open/Save As targets while leaving browser-native handles a no-op and preserving same-file Save semantics. Cache keys cover the changed browser modules. The broker regression exercises capability advertisement, valid release, and malformed-token rejection.

The newest canonical pull-request `CI Gate` run for this exact head is completed successful and its project-file/runtime coverage is green. I found no applicable unresolved refutation on this candidate. This GO is limited to the reference-lifetime repair; it does not claim Linux Firefox end-to-end acceptance or Windows Firefox qualification.

PR #333, opened 2026-09-12
Extend exact physical sequence toward vertical motion

djibian:
Refs #157 #321.

## Scope

This candidate extends the established deterministic physical program envelope with the already-generic Runtime v2 vertical motion semantic, without adding caller-selected altitude/state authority.

- `PhysicalProgramSequence` admits exact teacher-bound `vertical(up/down, distance)` statements, starts nominal altitude from the exact bound takeoff height, advances it only after definitive vertical completion, keeps horizontal/turn/wait/set_speed altitude-neutral, and eagerly rejects the whole program before flight if any intermediate nominal altitude leaves 0.2–1.5 m.
- The trusted SETPOINT_HL consumer emits exact bounded relative-world-Z vertical GO_TO effects through the existing teacher/session/watchdog/supervisor/SafeLink/acknowledgement/completion authority chain. Vertical duration is a fixed host policy with a 0.5 m/s peak ceiling and is independent of student horizontal `set_speed`.
- Terminal command-10 landing derives its relative descent from the exact final nominal altitude of the teacher-bound program; runtime sequencing permits landing only after every preceding exact effect has definitively completed.
- The parameter-free production-host regression consumes `takeoff -> vertical(up) -> vertical(down) -> land` in exact order, rejects caller-supplied direction/distance/height/index fields, proves no yaw dependency for world-Z-only motion, proves a definitive vertical rejection leaves the same exact step next, and proves an ambiguous vertical effect makes the remaining sequence fail closed.
- Pure timing regression establishes the fixed vertical ceiling at the inclusive 0.1 m and 0.8 m distance boundaries (and an interior point), and proves the policy is independent of run-local horizontal speed, alongside the one-shot GO_TO geometry/completion regression and cumulative altitude/landing-command tests.

The earlier Draft findings are resolved on this head: stale unsupported-vertical expectations were replaced with exact sequence/bounds/landing regressions, and the vertical timing regression exercises both admitted distance endpoints.

No real-flight qualification, motorized test, #70 estimator claim, checkpoint or `TEST_REQUIRED` is introduced or authorized by this candidate.

[...]

PR #332, opened 2026-09-12
Prepare X3 independent checkpoint support contract

djibian:
Refs #70 #326 #298 #331.

## Scope

This is support-only preparation after the integrated independent vertical replay predictor and integrated X3 capture-provenance correction. It does **not** enable a human-checkpoint profile and does not request a physical test.

- `tools/physical/run_x3_independent_capture.sh` freezes the deterministic local bundle runner expected by a later trusted checkpoint package: exact #251 `cf2.bin`, exact pinned cflib source/tree/subtree, the existing four-wheel Linux/Python 3.10 runtime closure, bundle manifest verification, isolated imports, exact request-SHA/bundle-target binding, and explicit operator confirmations for props removal and installation of the exact bundled binary before acquisition through the integrated collector.
- The runner fail-closes on the exact integrated collector provenance `x3-independent-props-off`: it parses `capture_independent_inputs.py --describe` and rejects any missing/different profile before either environment verification or recording can succeed. This addresses the prior exact-head `NO_GO 7510faed...` without enabling that profile in `human-checkpoint.yml`.
- Relative output paths remain relative to the operator's invoking directory rather than the bundle directory; the capture duration does not use Bash's special `SECONDS` variable.
- `X3_CHECKPOINT_SUPPORT.md` records the remaining eligibility boundary: exact unchanged firmware/parameters, numeric predictor-bound provenance frozen without target-outcome tuning, independently retained metric-reference/clock witnesses, and canonical Git-backed raw publication before a checkpoint may be enabled.
- The runner never flashes firmware, writes parameters, resets the estimator, invokes commander motion, publishes evidence, or manufactures a physical verdict. `--verify-environment` is hardware-free.

No Runtime/controller behavior, firmware source, estimator tuning, CI workflow, checkpoint registry, notification mechanism or governance file changes here. No `CHECKPOINT_REQUEST`, `TEST_REQUIRED`, or motorized action follows from this candidate.

Exact candidate: `0dcd25d13771e42efd8133a778b230114aacb1aa`, containing current `main@7f58f447daa36053bc4947a623e37fbe532a58d3`. The resulting PR diff remains bounded to the two support files; the new HEAD only merges current main, whose intervening #328 Firefox changes are outside this X3 slice. Fresh exact-head Ready CI and independent review are required. This execution mutated the candidate and does not claim independent GO.

## Remaining boundary

[...]

PR #335, opened 2026-09-12
Retain X3 sensor-to-output latency diagnostic

djibian:
Refs #70 #332.

## Scope

This bounded follow-up keeps the existing #251 firmware and X3 capture profile unchanged while retaining one already-exposed firmware diagnostic in the future raw capture:

- adds `stabilizer.intToOut` to the existing 20 ms `pose` log block; the six-float block remains 24/26 payload bytes;
- documents the pinned firmware semantics: microseconds from the IMU interrupt timestamp carried by the stabilizer's current `sensorData` to the end of that stabilizer iteration;
- explicitly keeps the value diagnostic-only. The low-priority log worker reads shared state later, so this observation is not atomically paired with an IMU row, is not a sensor producer timestamp and cannot validate `sensor_time_error_s` by itself;
- preserves all four existing CSV blocks, unchanged #251 firmware/parameters, X3 provenance identity, props-off boundary and the existing fail-closed acquisition lifecycle.

This directly records one existing internal timing witness without changing firmware or manufacturing a deterministic timing bound. It is intended to falsify overly optimistic sensor-to-output assumptions, not to turn observed latency into physical proof.

No estimator/controller/firmware source, Runtime, CI workflow, checkpoint registry, notification mechanism or governance change is included. No physical verdict, `CHECKPOINT_REQUEST`, `TEST_REQUIRED`, parameter write, estimator reset, commander motion or motorized action follows.

Exact candidate: `a97f906ceb9a1b55ab3a819e6c68fabb7ed599e9`, containing current `main@0a87debefe9a8dafc82ea77f37301e30867c5813`. The head-only transition from the previously reviewed `f0441cad8ca9ab831e8bca9d22d726e6455bc713` is a clean merge of current main after #333; the PR diff against current main remains limited to the collector and its acquisition documentation.

Any Draft `CI Gate (Draft)` result is non-decision evidence. A fresh canonical Ready `CI Gate` and independent exact-head review are required before integration.

### DIFF
--- a/experiments/crazyflie-ukf-surface-range/INDEPENDENT_INPUT_CAPTURE.md
+++ b/experiments/crazyflie-ukf-surface-range/INDEPENDENT_INPUT_CAPTURE.md
@@ -23,20 +23,27 @@ This is explicitly not remote firmware attestation. No flash is performed.
 | --- | ---: | --- | ---: |
 | `barometer` | 20 ms | `baro.asl`, `pressure`, `temp` | 12 bytes |
 | `imu` | 10 ms | `acc.x/y/z`, `gyro.x/y/z` | 24 bytes |
-| `pose` | 20 ms | downward range, roll/pitch, UKF Z/VZ | 20 bytes |
+| `pose` | 20 ms | downward range, roll/pitch, UKF Z/VZ, `stabilizer.intToOut` | 24 bytes |
 | `detector` | 20 ms | S3 state/reason/offset/barometer delta, local Flow flag, late eligibility | 24 bytes |
 
 Every field is fetched as a float; each block fits cflib's 26-byte log payload.
 Units remain the pinned firmware units: barometer altitude in m, pressure in
 mbar, temperature in Celsius; accelerometer in g; filtered gyro in degrees/s;
-downward range in mm; roll/pitch in degrees; UKF Z in m and VZ in m/s. Diagnostic
-UKF outputs must not become independent displacement inputs.
+downward range in mm; roll/pitch in degrees; UKF Z in m and VZ in m/s.
+`stabilizer.intToOut` is the pinned firmware's microsecond latency from the IMU
+interrupt timestamp carried by the stabilizer's current `sensorData` to the end
+of that stabilizer iteration. Diagnostic UKF outputs and this latency diagnostic
+must not become independent displacement inputs.
 
 Both the 24-bit device log timestamp and host monotonic receipt time are retained
 for every row. Log time does not identify a sensor's producer time, and fetching
 acceleration/gyro in one block does not prove simultaneous sensor acquisition.
-The chosen cadence is an acquisition configuration, not proof that inertial
-integration or a 5 cm / 1 s bound is achievable. No missing row is interpolated.
+The `intToOut` value is read later by the log worker from shared state and is not
+atomically paired with the pose or IMU row; it characterizes the firmware's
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO f0441cad8ca9ab831e8bca9d22d726e6455bc713

Independent falsification review against exact `main@f06c10f64f0d348ef0f06f1cc493d92fb480202f`. This execution did not mutate this candidate branch or PR state.

The two-file change is bounded to retaining one already-existing #251 firmware diagnostic in the raw X3 acquisition path. On pinned Crazyflie firmware `2026.08@54f31e243a0b28b67efef5ba20dbb6d9890a5478`, `stabilizer.intToOut` is a `LOG_UINT32` computed as `usecTimestamp() - sensorData.interruptTimestamp` at the end of the stabilizer iteration; the BMI088 path supplies that interrupt timestamp from its data-ready ISR. Pinned cflib explicitly permits a TOC variable to be fetched as a different wire type, so requesting this value as `float` keeps the six-value pose block at 24/26 bytes while firmware still uses the TOC-declared storage type.

The candidate correctly does not overclaim the observation: the log worker samples shared `intToOut` state later, so the retained value is neither atomically associated with the logged IMU row nor a barometer producer timestamp. It therefore cannot by itself validate `sensor_time_error_s`, eliminate the separate BMP3xx conversion/IIR dynamics bound, or establish physical success. Existing generic acquisition tests cover every declared BLOCKS column, live-variable presence, payload-size invariant, CSV retention and fail-closed capture lifecycle, so the added signal follows the already-tested path without widening checkpoint authority.

I found no applicable refutation of this bounded diagnostic claim. Ready integration evidence remains separate from this review. No checkpoint, `TEST_REQUIRED`, parameter/firmware change or motorized action follows from this GO.

djibian (commented):
GO a97f906ceb9a1b55ab3a819e6c68fabb7ed599e9

Independent exact-head revalidation against current `main@0a87debefe9a8dafc82ea77f37301e30867c5813`. This execution has not mutated PR #335 or its candidate branch.

The effective PR diff against current main remains exactly the same two-file X3 acquisition slice previously reviewed on `f0441cad8ca9ab831e8bca9d22d726e6455bc713`: one additional already-exposed `stabilizer.intToOut` value in the existing 20 ms pose block plus documentation of its diagnostic-only semantics. The head transition to `a97f906...` carries the integrated #333 physical vertical work from current main; none of those nine physical-host files is part of this PR diff.

The bounded claim remains correct: the six requested float values still fit the 26-byte cflib log payload, the retained firmware value is explicitly not treated as an atomic IMU-row timestamp or independent displacement input, and the documentation preserves the separate sensor-time/latency uncertainty rather than turning this observation into a physical acceptance bound. The change neither widens checkpoint authority nor modifies firmware/parameters, Runtime, CI, notification or governance mechanisms.

I found no applicable refutation on this exact Ready head. Canonical Ready `CI Gate` remains separate integration evidence. No checkpoint, `TEST_REQUIRED`, parameter write, estimator reset, commander motion or motorized action follows from this GO.

djibian (commented):
Post-integration note only: candidate `a97f906ceb9a1b55ab3a819e6c68fabb7ed599e9` is already merged as `82e03011d554a648d1bca10a00489503c1433e9a`; no new integration verdict is issued by this execution.

djibian (commented):
No further review action required: #335 is integrated as `main@82e03011d554a648d1bca10a00489503c1433e9a`.

Issue #340, opened 2026-09-12
Extend trusted physical sequence with bottom Color LED effect


djibian:
## Goal

Advance #157 from `main@0593ed990f495b5c6bd5f08945010748c0d34b7e` by consuming the already-generic `set_light(color)` statement on the trusted physical run for the **bottom Color LED Deck**, without widening browser/caller authority or treating deterministic host proof as real-device qualification.

Current integrated product evidence already covers the generic Blockly/AST palette (`off`, `red`, `green`, `blue`, `yellow`, `white`), simulation observability and exact-AST physical capability preflight. The remaining gap is the physical effect consumer itself.

## Primary-source boundary

The current Bitcraze Color LED firmware exposes the bottom deck value as writable `PARAM_UINT32 colorLedBot.wrgb8888`, formatted `0xWWRRGGBB`; the deck task transmits changed values to the deck as `[W,R,G,B]`. The official Color LED client writes that exact parameter and maps RGB to WRGB by extracting `white=min(R,G,B)` before packing.

Do **not** implement the trusted effect by blindly calling normal cflib `Param.set_value()`: cflib queues that write through its parameter updater and may use `send_packet(..., expected_reply=...)`, which can automatically retransmit on links that require resending. The trusted WebeeBlocks physical path requires an explicit one-effect ambiguity boundary rather than silently inheriting generic retry behavior.

For CRTP protocol v2 parameter writes, the firmware write response echoes the exact two-byte parameter id plus written value for a valid writable parameter. This can support a bounded one-shot acknowledgement/readback design when paired with the exact connection epoch, SafeLink, effect exclusion and fail-closed ambiguity semantics already integrated for physical effects.

## Required boundary

- Derive `set_light` color only from the exact teacher-bound canonical AST and exact `PhysicalProgramSequence` cursor. Ordinary host IPC remains parameter-free: callers cannot supply color, WRGB value, parameter id/name, cursor, completion or authority.
- Support only the existing bounded semantic palette. Use the official RGB→WRGB semantics, yielding exactly: `off=0x00000000`, `red=0x00ff0000`, `green=0x0000ff00`, `blue=0x000000ff`, `yellow=0x00ffff00`, `white=0xff000000`.
- Target only `colorLedBot.wrgb8888`. Resolve it from the exact live cflib parameter TOC and fail closed unless it is the expected writable `uint32_t` entry. Do not expose a generic parameter-write API.
[...]

PR #341, opened 2026-09-12
Extend exact physical sequence toward bottom Color LED

djibian:
Refs #157 #340.

## Scope

This candidate closes the bounded deterministic trusted-host slice for the already-generic bottom Color LED semantic.

- adds a narrow one-shot `colorLedBot.wrgb8888` PARAM transport surface for the existing Runtime v2 palette only;
- maps exactly `off`, `red`, `green`, `blue`, `yellow`, `white` to the official `0xWWRRGGBB` bottom-deck representation;
- resolves only the exact live writable `uint32_t` parameter from the cflib TOC and requires parameter protocol v2;
- avoids normal cflib `Param.set_value()` and its generic `expected_reply` resend path;
- serializes exact same-epoch Color LED PARAM transactions and requires live SafeLink, current teacher run, powered session, watchdog and process-wide effect exclusion;
- establishes a causal receive-freshness fence before each effect using a never-before-requested absent PARAM id on the pinned `radio://` path; only the exact firmware `ENOENT` response permits installing the target-write listener, while fence/write/readback uncertainty poisons the connection epoch;
- removes fallible write-listener teardown before minting the accepted-effect completion permit and treats later teardown/readback uncertainty as recovery-required rather than silently advancing;
- after exact write echo, requires a direct same-id PARAM readback of the exact value before restoring `FLYING` completion state;
- extends `PhysicalProgramSequence` with exact teacher-bound `set_light` claims; rejected/unemitted effects do not advance, ambiguity is terminal, and light is altitude/motion/speed neutral;
- composes the effect through the production host with ordinary `execute-next-inflight` IPC still parameter-free; caller-selected color/value/parameter/id/index/completion data remains rejected;
- adds deterministic transport, sequence and production-host regressions and wires them into the Runtime gate.

The causal-fence design specifically avoids treating cflib `Crazyflie.packet_sent` as an on-air boundary: on the pinned Crazyradio path that callback is only downstream of local enqueue. The unique invalid-id response is instead used as the receive-side ordering fence before any target reply can be accepted.

The production-host composition harness intentionally installs `FakeExecutionDomain`. Its ambiguity regression uses an explicit faithful fake transaction that starts in `FLYING`, records the one emitted effect as `EFFECT_UNRESOLVED`, and closes exceptional unresolved emission as `RECOVERY_REQUIRED`; the black-box assertions require the ambiguous request and every later request to fail closed, exactly one transport attempt, and no terminal landing. The production execution-domain lifecycle itself remains covered independently by its dedicated deterministic contract tests.

No generic physical parameter API, new Blockly vocabulary, top-deck light support, real-device qualification, checkpoint request or `TEST_REQUIRED` is introduced by this deterministic integration slice.

Exact Ready candidate: `7b9ec2e405161cfcac27317f6c7a3cd476ce3063`, containing current `main@b910d8edd9d922b93fd0600c8f933772faad89a9`. A fresh canonical Ready `CI Gate` on this exact head and an independent exact-head review resolving every still-applicable prior finding are required before integration.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -117,6 +117,10 @@ def main():
     if physical_program_sequence_test.returncode:
         print('FAIL exact physical program sequencing',file=sys.stderr);print(physical_program_sequence_test.stdout,file=sys.stderr);print(physical_program_sequence_test.stderr,file=sys.stderr);return physical_program_sequence_test.returncode
     print(physical_program_sequence_test.stdout.strip())
+    physical_color_led_transport_test=subprocess.run([sys.executable,'tools/ci/test_physical_color_led_transport.py'],text=True,capture_output=True)
+    if physical_color_led_transport_test.returncode:
+        print('FAIL trusted one-shot bottom Color LED transport',file=sys.stderr);print(physical_color_led_transport_test.stdout,file=sys.stderr);print(physical_color_led_transport_test.stderr,file=sys.stderr);return physical_color_led_transport_test.returncode
+    print(physical_color_led_transport_test.stdout.strip())
     physical_takeoff_command_test=subprocess.run([sys.executable,'tools/ci/test_physical_takeoff_command.py'],text=True,capture_output=True)
     if physical_takeoff_command_test.returncode:
         print('FAIL exact-bound physical takeoff command semantics',file=sys.stderr);print(physical_takeoff_command_test.stdout,file=sys.stderr);print(physical_takeoff_command_test.stderr,file=sys.stderr);return physical_takeoff_command_test.returncode
@@ -137,6 +141,10 @@ def main():
     if physical_host_inflight_test.returncode:
         print('FAIL actual physical-host exact-program in-flight composition',file=sys.stderr);print(physical_host_inflight_test.stdout,file=sys.stderr);print(physical_host_inflight_test.stderr,file=sys.stderr);return physical_host_inflight_test.returncode
     print(physical_host_inflight_test.stdout.strip())
+    physical_host_color_led_test=subprocess.run([sys.executable,'tools/ci/test_physical_host_color_led_sequence.py'],text=True,capture_output=True)
+    if physical_host_color_led_test.returncode:
+        print('FAIL actual physical-host exact bottom Color LED composition',file=sys.stderr);print(physical_host_color_led_test.stdout,file=sys.stderr);print(physical_host_color_led_test.stderr,file=sys.stderr);return physical_host_color_led_test.returncode
+    print(physical_host_color_led_test.stdout.strip())
     browser=browser_binary()
     with socketserver.TCPServer(('127.0.0.1',0),QuietHandler) as server:
         port=server.server_address[1]; threading.Thread(target=server.serve_forever,daemon=True).start(); time.sleep(.05); url=f'http://127.0.0.1:{port}/{HARNESS.as_posix()}'
--- a/tools/ci/test_physical_color_led_transport.py
+++ b/tools/ci/test_physical_color_led_transport.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 from pathlib import Path
+from threading import Event, Thread
 from types import SimpleNamespace
 import struct
 import sys
@@ -51,6 +52,42 @@ class Packet:
         self.port = port
         self.channel = channel
         self.data = bytearray(data)
+
+
+class CallbackBus:
+    def __init__(self) -> None:
+        self.callbacks: list = []
+
+    def add_callback(self, callback) -> None:
+        if callback not in self.callbacks:
+            self.callbacks.append(callback)
+
+    def remove_callback(self, callback) -> None:
+        if callback in self.callbacks:
+            self.callbacks.remove(callback)
+
+    def call(self, value) -> None:
+        for callback in tuple(self.callbacks):
+            callback(value)
+
+
+class GateLock:
+    """Deterministically hold callback processing after entry freshness capture."""
+
+    def __init__(self) -> None:
+        self.entered = Event()
+        self.release = Event()
+
+    def __enter__(self):
+        self.entered.set()
+        require(
+            self.release.wait(1.0),
+            "threaded stale callback was not released",
+        )
+        return self
+
+    def __exit__(self, exc_type, exc, traceback) -> bool:
+        return False
 
 
 class Platform:
@@ -115,6 +152,7 @@ class FakeCrazyflie:
         self.param = SimpleNamespace(
             toc=ParamToc(ParamElement() if element is None else element)
         )
+        self.packet_sent = CallbackBus()
         self.write_reply = write_reply
         self.read_reply = read_reply
         self.stale_write_value = stale_write_value
@@ -150,7 +188,7 @@ class FakeCrazyflie:
         elif channel == 1 and self.stale_read_value is not None:
             callback(
                 Packet(
-                    struct.pack("<HB L", ident, 0, self.stale_read_value),
+                    struct.pack("<HBL", ident, 0, self.stale_read_value),
                     port=port,
                     channel=channel,
                 )
@@ -179,17 +217,21 @@ class FakeCrazyflie:
         if packet.channel == 2:
             require(len(data) == 6, "write request must be exact uint32 PARAM write")
             self.current_value = struct.unpack("<L", data[2:6])[0]
-            if self.write_reply:
-                self._emit_header(Packet(data, port=packet.port, channel=packet.channel))
         elif packet.channel == 1:
             require(len(data) == 2, "read request must carry only exact parameter id")
-            if self.read_reply:
-                reply = data + bytes((0,)) + struct.pack("<L", self.current_value)
-                self._emit_header(
-                    Packet(reply, port=packet.port, channel=packet.channel)
-                )
         else:
             raise AssertionError("unexpected PARAM channel")
+
+        # Match pinned cflib: packet_sent is called only after link.send_packet()
+        # returns and before an independently delivered incoming PARAM callback is
+        # accepted by this deterministic fake.
+        self.packet_sent.call(packet)
+
+        if packet.channel == 2 and self.write_reply:
+            self._emit_header(Packet(data, port=packet.port, channel=packet.channel))
+        elif packet.channel == 1 and self.read_reply:
+            reply = data + bytes((0,)) + struct.pack("<L", self.current_value)
+            self._emit_header(Packet(reply, port=packet.port, channel=packet.channel))
 
 
 def packet_factory(channel: int, data: bytes):
@@ -205,7 +247,13 @@ def unique(label: str) -> str:
     return f"color-{label}-{_counter}"
 
 
-def state(*, blocking_fault=False, is_flying=True, hl_control_active=True, hl_traj_finished=True):
+def state(
+    *,
+    blocking_fault=False,
+    is_flying=True,
+    hl_control_active=True,
+    hl_traj_finished=True,
+):
     return SimpleNamespace(
         bitfield=0,
         blocking_fault=blocking_fault,
@@ -368,6 +416,64 @@ def test_success_is_one_write_then_fresh_readback_without_retry() -> None:
         fixture.close()
 
 
+def _prove_callback_started_before_send_stays_stale(*, channel: int) -> None:
+    is_write = channel == 2
+    cf = FakeCrazyflie(write_reply=not is_write, read_reply=is_write)
+    fixture = Fixture("threaded-write" if is_write else "threaded-read", cf)
+    try:
+        if is_write:
+            request = struct.pack("<HL", 0x1234, 0x00FF0000)
+            stale_reply = request
+        else:
+            request = struct.pack("<H", 0x1234)
+            stale_reply = request + bytes((0,)) + struct.pack("<L", 0x00FF0000)
+            cf.current_value = 0x00FF0000
+        packet = packet_factory(channel, request)
+        event = Event()
+        reply_box: list[bytes | None] = [None]
+        gate = GateLock()
+        listener = fixture.transport._listen_for_param(
+            channel,
+            request[:2],
+            packet,
+            event,
+            reply_box,
+            gate,
+        )
+        worker = Thread(
+            target=listener.header_callback,
+            args=(Packet(stale_reply, port=0x02, channel=channel),),
+            daemon=True,
+        )
+        worker.start()
+        require(
+            gate.entered.wait(1.0),
+            "stale callback must enter before the current send",
+        )
+
+        # This exact send advances packet_sent only after the stale callback has
+        # already snapshotted freshness=False. The callback resumes afterwards.
+        cf.send_packet(packet)
+        gate.release.set()
+        worker.join(timeout=1.0)
+        require(not worker.is_alive(), "stale callback must finish deterministically")
+        require(
+            not event.is_set() and reply_box[0] is None,
+            "pre-send callback must not become fresh after packet_sent",
+        )
+        fixture.transport._remove_listener(listener)
+    finally:
+        fixture.close()
+
+
+def test_threaded_pre_emission_write_reply_cannot_authorize_effect() -> None:
+    _prove_callback_started_before_send_stays_stale(channel=2)
+
+
+def test_threaded_pre_emission_read_reply_cannot_prove_completion() -> None:
+    _prove_callback_started_before_send_stays_stale(channel=1)
+
+
 def test_pre_emission_matching_write_reply_cannot_authorize_effect() -> None:
     cf = FakeCrazyflie(
         write_reply=False,
@@ -406,6 +512,28 @@ def test_pre_emission_matching_read_reply_cannot_prove_completion() -> None:
         require(
             fixture.domain.phase == execution.RECOVERY_REQUIRED,
             "missing fresh readback after accepted write requires recovery",
+        )
+    finally:
+        fixture.close()
+
+
+def test_ambiguous_readback_poison_blocks_same_epoch_reuse() -> None:
+    cf = FakeCrazyflie(write_reply=True, read_reply=False)
+    fixture = Fixture("read-poison", cf)
+    try:
+        expect_error(
+            lambda: fixture.transport.send_color(color="red"),
+            execution.PhysicalExecutionDomainError,
+            "fresh effect-completion proof failed",
+        )
+        require(
+            fixture.domain.phase == execution.RECOVERY_REQUIRED,
+            "ambiguous readback must require physical recovery",
+        )
+        expect_error(
+            lambda: HostBoundColorTransport(**fixture.kwargs),
+            color.ColorLedTransportError,
+            "poisoned",
         )
     finally:
         fixture.close()
@@ -472,15 +600,18 @@ def test_toc_identity_and_type_fail_before_effect() -> None:
 def main() -> int:
     test_exact_palette_mapping()
     test_success_is_one_write_then_fresh_readback_without_retry()
+    test_threaded_pre_emission_write_reply_cannot_authorize_effect()
+    test_threaded_pre_emission_read_reply_cannot_prove_completion()
     test_pre_emission_matching_write_reply_cannot_authorize_effect()
     test_pre_emission_matching_read_reply_cannot_prove_completion()
+    test_ambiguous_readback_poison_blocks_same_epoch_reuse()
     test_write_listener_cleanup_failure_cannot_strand_completion_permit()
     test_direct_importable_core_has_no_positive_provenance_path()
     test_toc_identity_and_type_fail_before_effect()
     print(
         "PASS trusted bottom Color LED transport keeps exact palette/TOC authority, "
-        "uses one no-retry write plus causal readback, rejects pre-emission stale replies, "
-        "and cannot strand an accepted-effect completion permit"
+        "uses one no-retry write plus causal readback, classifies callback freshness at exact packet_sent, "
+        "poisons ambiguous same-epoch readback, and cannot strand an accepted-effect completion permit"
     )
     return 0
 
--- a/tools/ci/test_physical_color_led_transport.py
+++ b/tools/ci/test_physical_color_led_transport.py
@@ -1,0 +1,489 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+from pathlib import Path
+from types import SimpleNamespace
+import struct
+import sys
+
+ROOT = Path(__file__).resolve().parents[2]
+PHYSICAL = ROOT / "tools" / "physical"
+sys.path.insert(0, str(PHYSICAL))
+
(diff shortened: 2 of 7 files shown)


djibian (commented):
DRAFT_FINDING 507139fadc6e9a578a0ea4c2a6a841aac9163a5e

The initial one-shot PARAM transport has a fail-closed lifecycle gap after positive write acknowledgement. In `send_color()`, `effect.mark_accepted()` moves the process-wide execution domain to `AWAITING_COMPLETION` and returns the only completion permit, but the write-listener cleanup happens afterward in the `finally` block. If `_remove_listener()` raises there, the method exits before `complete_accepted_effect(...)`; the permit is lost while the execution domain remains `AWAITING_COMPLETION`. That state is neither ordinary effect-eligible nor accepted by `run_reset_establishment()`, so recovery is blocked rather than required/possible.

Smallest repair boundary: ensure every exception after the accepted-effect transition but before the readback completion path causally drives the execution domain to `RECOVERY_REQUIRED`, or perform all potentially failing write-listener cleanup before `mark_accepted()`. Add a deterministic regression that forces write-listener removal failure after an exact echoed write and proves the run does not remain stranded in `AWAITING_COMPLETION` and cannot silently advance/retry the light step.

djibian:
DRAFT_FINDING 507139fadc6e9a578a0ea4c2a6a841aac9163a5e

The PARAM reply listeners are installed before the corresponding packet is emitted, but the callbacks are immediately active and correlate only by port/channel plus the two-byte parameter id (the write path later checks full request bytes). A delayed same-id reply from an earlier transaction in the same connection epoch can therefore set `event` before the current `send_packet(...)`; after emission `_wait_for_packet()` can consume that pre-emission packet. For consecutive identical color writes, an old exact write echo could satisfy the new write acknowledgement; likewise an old same-value read reply could satisfy `_prove_readback()`. That does not establish the required causal/fresh reply/readback for the current effect.

Smallest repair boundary: keep the callback registered early enough not to miss an immediate response, but gate acceptance on a transaction-local post-emission arm/freshness boundary (a reply observed before the current request is marked emitted must be ignored/fail closed). Add deterministic regressions that inject matching same-id write/read replies before the current send and prove they cannot authorize acknowledgement/completion. Preserve the no-resend boundary.

djibian (commented):
DRAFT_FINDING 4b4cb3b1403356bf5951472df9a8b7b6d1e5c02b

The previous pre-send stale-reply finding is only partially repaired. `_listen_for_param()` now serializes callback processing with `freshness_lock`, and `send_color()` / `_prove_readback()` set `armed` while holding that lock immediately before invoking `send_packet()`. This prevents a callback from *processing* before the send, but it does not prove the matching packet arrived after the current request: a stale same-id callback from another thread can begin before the send, block on `freshness_lock`, then acquire it after `send_packet()` returns, observe `armed`, and populate the current transaction's reply box. An old exact echo/readback can therefore still satisfy the new request.

Smallest repair boundary: the freshness boundary must classify callback arrival/entry before the current emission as stale independently of when it later acquires a processing lock. For example, capture a transaction-local generation/armed state at callback entry under a non-blocking/atomic boundary that cannot turn a pre-emission callback into a post-emission one, while still registering before send so synchronous immediate replies are not missed. Add deterministic regressions where a matching callback begins before emission, is deliberately blocked until after `send_packet()`, and prove it cannot authorize either the write acknowledgement or readback completion. Preserve the one-shot/no-resend semantics.

djibian:
DRAFT_FINDING 4b4cb3b1403356bf5951472df9a8b7b6d1e5c02b

The cleanup repair correctly moves the fallible write-listener removal before `mark_accepted()`, but the reply-freshness repair does not yet establish causality. `armed` is set **before** `_send_packet(packet)` while `freshness_lock` is held. A delayed matching callback that is invoked in that arm/send window (or is blocked on `freshness_lock` while the current send runs) executes after the lock is released, sees `armed`, and is accepted even though its packet need not be a response to the current request. The lock orders callback execution; it does not bind packet arrival/provenance to the current send.

This is especially consequential for readback: a timed-out/ambiguous `_prove_readback()` drives the execution domain to recovery, but it does not poison the Color-LED PARAM epoch. A later trusted reset/new run on the same connection epoch can therefore install a new same-id/same-value read listener, and a delayed reply from the prior failed read can satisfy the new completion proof.

Smallest repair boundary: make readback uncertainty invalidate whatever freshness authority is needed to prevent same-epoch reuse (or otherwise establish a transaction-local causal receive boundary that cannot accept a callback originating before the current request), and keep the one-shot/no-resend rule. Add deterministic regressions that (1) schedule a matching callback across the arm/send critical section and (2) time out one readback, recover/reset, then deliver that old reply during a later same-value run; neither may authorize acknowledgement/completion.

djibian:
DRAFT_FINDING 3b39b5f8593e95ae52fc5be928db0eeab9cda64e

The new deterministic transport regression file is not yet part of decision CI. `tools/ci/test_physical_color_led_transport.py` is added on this head, but `tools/ci/run_runtime_v2_core_harness.py` at the same exact head still enumerates the physical Python tests explicitly and never invokes it. `ci-runtime.yml` runs that harness rather than test discovery. Draft CI skips Runtime by contract, and merely making the PR Ready would therefore not execute these Color-LED transport tests unless the harness is wired.

Smallest repair: add the new transport test to `run_runtime_v2_core_harness.py` (with a fail-closed labelled failure like the adjacent physical transport tests), then ensure the Ready selector requires Runtime for the final changed paths and the exact-head canonical `CI Gate` includes the resulting execution. This finding is independent of the outstanding reply-causality race.

djibian (commented):
DRAFT_FINDING 23592ec4f4bef844a4f97f5024942e8b179b94d3

The new production-host ambiguity regression does not actually prove terminal ambiguity. In `test_ambiguous_light_effect_makes_host_sequence_terminal()`, the fake executes `self.execution_domain.phase = RECOVERY_REQUIRED`, but `PhysicalExecutionDomain.phase` is a read-only property with no setter. That assignment raises `AttributeError` while the real process-wide domain is still `FLYING`; `_execute_light()` therefore takes its `phase == FLYING` branch and `release_unemitted(claim)` instead of `mark_ambiguous(...)`. The second parameter-free step can retry the same light. The test only asserts that both replies are failures and that landing did not occur, so this opposite/retryable behavior can still satisfy the test name and PASS message.

Smallest repair: inject ambiguity through the real execution-domain lifecycle (or a faithful fake that leaves the domain non-FLYING/recovery-required after an emitted uncertain effect), then assert the sequence itself is terminal and the second request cannot re-attempt the light. Also assert exactly one ambiguous-light attempt, so a retry cannot masquerade as terminal fail-closed behavior.

djibian (commented):
DRAFT_FINDING 23592ec4f4bef844a4f97f5024942e8b179b94d3

The exact-head candidate now wires the Color LED transport/host regressions into the Runtime gate and repairs the earlier listener-cleanup lifecycle defect, but the applicable freshness refutation from `DRAFT_FINDING 4b4cb3b1403356bf5951472df9a8b7b6d1e5c02b` is still unresolved despite the PR body claiming a post-emission freshness boundary.

`tools/physical/color_led_transport.py` is still blob `443a7a0756b52edd0537ada7bbde914259839287`, unchanged from the refuted `4b4cb3b...` design. `_listen_for_param()` first blocks on `freshness_lock` and only then reads `armed`; `send_color()` / `_prove_readback()` set `armed` while holding that same re-entrant lock immediately before `send_packet()`. Therefore a matching callback on another thread may enter before the current emission, block on the lock, resume after `send_packet()`, observe `armed`, and be accepted as the current reply. The newly wired regressions only inject matching stale replies synchronously during callback registration, before that blocking race exists; they do not exercise the cross-thread arrival-before-emission/processing-after-emission case required by the prior finding.

Smallest repair remains to classify freshness at callback entry/arrival independently of later lock acquisition (for example with a transaction-local generation/emission token captured without waiting on the processing lock), while preserving listener-before-send registration and synchronous immediate current replies. Add deterministic write-ack and readback regressions that start a matching callback before emission, hold it until after the current send, and prove it cannot authorize the effect or completion. Preserve the one-shot/no-resend and fail-closed recovery semantics.

No real-device qualification, checkpoint, `TEST_REQUIRED`, parameter write or motorized action follows from this Draft finding.

djibian (commented):
DRAFT_FINDING 23592ec4f4bef844a4f97f5024942e8b179b94d3

Independent exact-head falsification review against current `main@4e63f955ec6db9829c0a7ce373115fc1aace1e73`; this execution has not mutated PR #341 or its candidate branch.

The lifecycle finding from `507139f...` is repaired on this head: the fallible write-listener removal occurs before `mark_accepted()` mints the unique completion permit, and the deterministic cleanup-failure regression requires `RECOVERY_REQUIRED` rather than a stranded `AWAITING_COMPLETION` state. The sequence, production-host dispatch, Runtime-gate wiring and parameter-free caller boundary are also now present.

However, the still-applicable freshness finding from `4b4cb3b...` is not repaired despite the PR body claiming it is. `_listen_for_param()` still checks `armed.is_set()` only after entering `freshness_lock`. A matching callback can enter on another thread before the current send, block on that lock while `send_color()` / `_prove_readback()` arm and invoke `send_packet()`, then acquire the lock afterward and observe `armed=True`; that pre-emission packet can still populate the current reply box. The new stale-reply tests inject their stale callback synchronously during callback registration, so they do not exercise this blocked-callback race.

Repair the freshness boundary by classifying callback entry/arrival before it can block on the emission-processing lock (for example, snapshot the transaction-local armed/generation state at callback entry and reject an entry observed before emission even if processing occurs later). Add deterministic threaded regressions for both write acknowledgement and readback where a matching callback begins before emission, is held until after `send_packet()`, and must not authorize the current transaction. Preserve registration-before-send for synchronous immediate replies and the one-shot/no-resend effect semantics.

No real-device qualification, checkpoint or `TEST_REQUIRED` follows from this deterministic slice.

djibian:
DRAFT_FINDING 2e3846e4c6fcf9fe26dd6976fc6a3f98ed4a44c6

The new `packet_sent` freshness boundary still does not establish **physical emission** on the exact pinned Crazyradio path. In pinned cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873`, `Crazyflie.send_packet()` calls `self.link.send_packet(pk)` and then immediately invokes `self.packet_sent.call(pk)`. But `RadioDriver.send_packet()` only enqueues `pk` into `out_queue`; the `_RadioDriverThread` transmits the previously prepared `dataOut`, processes/enqueues that exchange's downlink, and only afterward dequeues `out_queue` to prepare the packet for the **next** radio exchange. Therefore `packet_sent` can become true before the current PARAM request has gone over the air.

A delayed same-id reply from an earlier PARAM transaction can consequently arrive after this head's `sent` event is set but before the current request is physically transmitted. `_listen_for_param()` will classify that old callback as fresh and it can authorize either the write acknowledgement or the same-value readback. This is the same causal class as the earlier stale-reply finding, just one layer lower than the callback-entry race.

Smallest repair boundary: do not treat `Crazyflie.packet_sent` as an on-air emission oracle for `radio://`. Establish a receive/transaction boundary that is causally downstream of the current radio exchange (or otherwise make the response uniquely attributable to the current transaction) while preserving the one-shot/no-resend requirement. Add a deterministic regression that models the pinned RadioDriver queue ordering: set the enqueue-level `packet_sent` marker, deliver an old matching reply before the queued current request is actually transmitted, and prove it cannot authorize write acknowledgement or readback completion. No real-device qualification or `TEST_REQUIRED` follows.

djibian (commented):
DRAFT_FINDING 176dc6766222d6cc659cfed2c29960082265da12

The production transport now binds PARAM reply freshness to the pinned cflib `packet_sent` callback, but the exact-head deterministic transport harness was not updated to that contract. `TrustedBottomColorLedTransport.__init__()` requires `crazyflie.packet_sent.add_callback/remove_callback`; `tools/ci/test_physical_color_led_transport.py` still gives `FakeCrazyflie` no `packet_sent` object at all. Consequently `Fixture(...)` cannot construct the transport once this test is actually executed. The current exact-head run 1288 is Draft-only and selected no Runtime suite, so its green `CI Gate (Draft)` does not exercise the newly wired transport test.

The applicable freshness refutation also still lacks the required deterministic cross-thread regressions. The existing stale-write/read tests invoke a matching callback synchronously during `add_header_callback()`. They do not start a matching callback before emission, block its processing until after the current `send_packet()`, and prove that callback-entry freshness rejects it. That blocked-callback race was the specific counterexample in the prior finding.

The production mechanism is directionally consistent with exact pinned cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873`: its `send_packet()` calls `link.send_packet(pk)` and then `packet_sent.call(pk)`, so the exact-packet marker is post-link-send. Repair the harness to model that callback contract, add threaded write-ack and readback regressions for arrival-before-emission/processing-after-emission, and ensure the transport test actually passes before offering a Ready head. Preserve the one-shot/no-resend and fail-closed ambiguity semantics.

No real-device qualification, checkpoint or `TEST_REQUIRED` follows from this finding.

djibian (commented):
DRAFT_FINDING d403b04a59afc6d23c7c861ede594212669d4478

Independent exact-head re-review against current `main@be74e3c0a177c91fa4acca05590f3ede9a28bab0`; this execution has not mutated PR #341 or its candidate branch.

The new threaded regressions correctly repair the *callback-entry/processing-lock* test gap, and the host ambiguity regression now uses the real execution-domain lifecycle. They do not resolve the deeper exact-pinned-radio causality finding already recorded on `2e3846e...`.

The new fake explicitly says it “Match[es] pinned cflib” by calling `packet_sent` after its own `send_packet()` has updated state and before delivering an incoming PARAM callback. That models only `Crazyflie.send_packet()`, not the pinned `radio://` link beneath it. At cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873`, `Crazyflie.send_packet()` calls `RadioDriver.send_packet(pk)` and then immediately `packet_sent.call(pk)`, while `RadioDriver.send_packet()` only enqueues `pk` into the size-1 `out_queue`. `_RadioDriverThread` performs the current radio exchange using the *previously prepared* `dataOut`, processes/enqueues that exchange's downlink, and only then dequeues `out_queue` to prepare this `pk` for the next exchange. Therefore the exact `packet_sent` marker used by production can still precede on-air transmission of the current PARAM request.

A delayed same-id reply from an earlier PARAM transaction can consequently enter after `packet_sent` and satisfy these new tests' freshness predicate even though the current request has not yet been transmitted. The fake cannot falsify that case because it collapses enqueue and physical exchange into one method.

Repair must move the causal boundary below the enqueue-level `Crazyflie.packet_sent` marker or establish response uniqueness by another proven invariant. Add a deterministic regression that models the pinned RadioDriver ordering explicitly: enqueue current request -> fire `packet_sent` -> deliver an old matching reply while the current request is still only queued -> only later perform the current radio exchange. That old reply must not authorize either write acknowledgement or readback completion. Preserve exactly one PARAM write, no generic cflib retry surface, same-epoch poisoning on uncertainty, and the parameter-free AST boundary.

Draft CI success on this exact head remains development evidence only and does not make this causal precondition known. No real-device qualification, checkpoint or `TEST_REQUIRED` follows.

djibian (commented):
DRAFT_FINDING d403b04a59afc6d23c7c861ede594212669d4478

The callback-entry race from the prior finding is now exercised deterministically, but `packet_sent` is not an actual-radio-emission boundary on the exact pinned radio transport used by the reference Crazyflie path.

Pinned cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873` does call `self.link.send_packet(pk)` and then `self.packet_sent.call(pk)` in `Crazyflie.send_packet()`. However, its exact `RadioDriver.send_packet()` only performs `self.out_queue.put(pk, ...)` and returns. `_RadioDriverThread.run()` transmits a previously prepared `dataOut`, processes that ACK/downlink, only then dequeues the next `outPacket` and promotes it to `dataOut` for a later loop iteration. Therefore `packet_sent` proves enqueue-after-call, not that this exact request has crossed the radio link.

This leaves a causal freshness hole: a delayed same-id PARAM response from an earlier transaction can be delivered after the new request has been queued (so `sent.is_set()` is already true) but before that new request is actually transmitted. `header_callback()` would classify that old response as fresh and it can satisfy the exact echo/readback check. The new fake explicitly places `packet_sent` before incoming delivery, so it models the assumption rather than falsifying this queue-to-RF gap.

Smallest repair boundary: bind acceptance to evidence that is causally downstream of the exact request's actual radio transaction, not merely the `Crazyflie.send_packet()` enqueue return. This likely requires a narrow exact-pinned radio/link transaction hook (or an equivalent protocol-level discriminator) that can associate the received payload with the actual transmitted CRTP packet without enabling generic retries. Add a deterministic regression with three ordered events: new request queued -> stale old matching PARAM response delivered -> new request physically transmitted; the stale response must remain ineligible.

Do not weaken the one-shot/no-resend, same-epoch poisoning, process-wide effect exclusion, or caller-parameter boundaries. No real-device qualification, checkpoint or `TEST_REQUIRED` follows from this Draft finding.

djibian (commented):
GO 5f29d280f9352817629a64eac146f6d0b4e3f133

Independent exact-head review against current `main@b910d8edd9d922b93fd0600c8f933772faad89a9`; this execution has not mutated PR #341 or its candidate branch.

The prior lifecycle and PARAM-freshness refutations are resolved on this candidate. The transport no longer treats cflib `packet_sent` as an RF boundary. Instead it serializes the epoch, reads a never-reused absent PARAM id, and requires the exact protocol-v2 `ENOENT` reply before installing the target write listener. On pinned firmware `54f31e243a0b28b67efef5ba20dbb6d9890a5478`, PARAM requests are serialized by the single PARAM task and replies pass the FIFO CRTP/radiolink transmit path; on pinned cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873`, received packets are dispatched through the single FIFO incoming handler. Therefore any older same-id PARAM reply causally precedes the unique fence response and is dispatched before the target listener exists, including the previously refuted local-enqueue/on-air gap.

The deterministic regressions model stale write/read replies arriving after local enqueue but before the fence exchange, fence/write/readback uncertainty poisons the connection epoch, write-listener teardown occurs before acceptance, and the host ambiguity regression now drives the fake execution domain to `RECOVERY_REQUIRED` so the exact light claim becomes terminal rather than retryable. The exact AST/teacher/session/watchdog/SafeLink/current-program and parameter-free caller boundaries remain preserved. The merge of current main into this exact head only carries the independently integrated #344 X3 documentation boundary and does not alter these Color LED mechanisms.

No unresolved applicable exact-head finding remains from the prior DRAFT_FINDING reviews. This GO is deterministic integration evidence only; it does not qualify real-device Color LED operation and does not justify `TEST_REQUIRED`.

djibian (commented):
NO_GO 5f29d280f9352817629a64eac146f6d0b4e3f133

Independent exact-head review against current `main@b910d8edd9d922b93fd0600c8f933772faad89a9`; this execution has not mutated PR #341 or its candidate branch.

One still-applicable deterministic refutation from `23592ec...` remains unchanged on this Ready head. In `tools/ci/test_physical_host_color_led_sequence.py`, `test_ambiguous_light_effect_makes_host_sequence_terminal()` still executes `self.execution_domain.phase = physical_execution_domain.RECOVERY_REQUIRED`. The exact `PhysicalExecutionDomain` exposes `phase` as a read-only `@property` with no setter, so that assignment raises `AttributeError` while the process-wide phase is still `FLYING`. Production `_execute_light()` catches the exception and `_release_or_poison_sequence()` therefore takes its `phase == FLYING` branch and releases the pending light claim as retryable. The second parameter-free request can reach the same light transport again instead of proving terminal ambiguity. The newly added `attempts["count"] == 1` assertion would expose this once the Runtime suite actually executes; Draft run 1302 skipped Runtime and is not decision evidence.

Repair the regression through the real execution-domain lifecycle (for example, enter an `effect_transaction`, `mark_emitted()`, then raise while leaving the effect unresolved so the domain itself transitions to `RECOVERY_REQUIRED`) or use a faithful fake with an actually mutable internal lifecycle. Keep the exactly-one-attempt assertion and prove the second request is rejected before reaching the Color LED transport.

This verdict does not require a real-device test and establishes no `TEST_REQUIRED`.

djibian (commented):
DRAFT_FINDING ed9be35d254f44345974302b31ed98824306f6c8

Independent exact-head falsification review against current `main@b910d8edd9d922b93fd0600c8f933772faad89a9`; this execution has not mutated PR #341 or its candidate branch.

The new ambiguity regression does not actually reach the real `PhysicalExecutionDomain.effect_transaction()` lifecycle it intends to exercise. `host.run_host_sequence()` calls `host.install_fakes()`, which in turn calls `test_physical_host_activation.install_fakes()`. That installation replaces both `physical_execution_domain.PhysicalExecutionDomain` and `activation.PhysicalExecutionDomain` with `FakeExecutionDomain`. The exact fake instantiated for the active run has only mutable `phase` and `run_reset_establishment()`; it has no `effect_transaction()` method. Therefore the new `with self.execution_domain.effect_transaction(...)` raises `AttributeError` before `mark_emitted()`, while the fake domain is still `FLYING`. Production `_execute_light()` then sees the same retryable phase and releases the sequence claim rather than proving terminal ambiguity. The earlier direct assignment was in fact writing this mutable fake, not the production read-only property.

The added postcondition assertion cannot repair that: it is reached only after the two host step replies, and the intended exactly-one-attempt assertion should already fail if the second parameter-free request retries the light. Smallest repair boundary: either preserve/use a saved real `PhysicalExecutionDomain` instance for this host ambiguity test and compose the surrounding fakes around it, or extend the host test execution-domain fake with a faithful emitted/unresolved transaction lifecycle that transitions to `RECOVERY_REQUIRED` on exceptional exit. Keep the exactly-one-attempt assertion and an explicit recovery-required postcondition.

This deterministic finding requires no real-device qualification and establishes no `TEST_REQUIRED`.

djibian (commented):
GO 48cdc20997b63fa341e274fdf165077394192786

Independent exact-head falsification review against current `main@b910d8edd9d922b93fd0600c8f933772faad89a9`; this execution has not mutated PR #341 or its candidate branch.

The previously established listener-cleanup and PARAM-freshness repairs remain unchanged and applicable. The last unresolved host-ambiguity finding on `ed9be35d...` is now resolved: the production-host composition harness intentionally uses `FakeExecutionDomain`, and the current regression models the production emitted-unresolved close rule through a faithful local transaction on that exact fake. It starts from `FLYING`, records the emitted effect as `EFFECT_UNRESOLVED`, closes exceptional unresolved emission as `RECOVERY_REQUIRED`, then proves the second parameter-free request does not re-attempt the light (`attempts == 1`), the exact observed domain is recovery-required, and no landing follows. This repairs the prior false assumption that the host fake exposed the production `effect_transaction()` API while preserving the black-box host behavior under test.

The exact Ready head has canonical pull-request `CI Gate` run 1311 (`34705976098`) completed successfully, including its final `CI Gate` check on this SHA. No still-applicable prior refutation remains on this candidate. This is deterministic integration evidence only; it does not qualify real-device Color LED operation and does not establish a `TEST_REQUIRED` checkpoint.

PR #339, opened 2026-09-12
Add X3 pre-LPF timing observer

djibian:
Refs #70.

## Scope

This is a logging-only Lab diagnostic for the independent-displacement X3 path. It does not change estimator/controller equations, ToF/Flow/S3 behavior, parameters, Runtime, checkpoint authority or motor authority.

- applies only to exact Crazyflie firmware `2026.08@54f31e243a0b28b67efef5ba20dbb6d9890a5478` and exact pinned sensor source;
- retains gravity-aligned acceleration immediately **before** the existing 30 Hz firmware LPF, plus MCU `usecTimestamp()` begin/end around the accelerometer register read;
- retains one compensated BMP3xx barometer snapshot plus MCU begin/end around `bmp3_get_sensor_data()` and exact runtime chip ID;
- exposes complete 24-byte `x3AccObs` and 25-byte `x3BaroObs` function-backed log groups;
- uses one packet-timestamp latch plus short producer/latch critical sections so one complete observer group requested in one log block cannot mix adjacent producer-loop snapshots;
- adds deterministic contract coverage for exact transformation/idempotence, pre-LPF placement, callback-to-latch wiring, partial-application fail-closed behavior and payload budgets;
- adds a separate `run_x3_prelpf_build_oracle.sh` diagnostic path that operates on its own pinned firmware checkout and compiles the observer without changing the canonical S3 checkpoint oracle or artifact;
- wires that bounded diagnostic into the existing canonical `.github/workflows/ci.yml` `Select suites` job, retaining only a distinctly named diagnostic evidence artifact (`x3-prelpf-observer-build-evidence-2026-08`).

The existing `run_s3_build_oracle.sh`, `ci-webots.yml` S3 production path and trusted `experimental-s3-surface-offset-2026-08` / `s3-props-off` artifact identity remain unchanged. The earlier standalone `x3-prelpf-observer.yml` workflow has been removed rather than weakening the V4 workflow inventory contract.

The retained interrupt timestamp is gyro-DRDY, not an accelerometer producer timestamp. The MCU read windows likewise do **not** establish sensor-internal sample production/filter timing. BMP3xx conversion/oversampling/IIR dynamics and deterministic IMU/filter/physical-motion allowances remain separate obligations. This candidate cannot set `sensor_producer_timing_validated`, select frozen-predictor bounds or establish physical success.

[...]

PR #342, opened 2026-09-12
Fail closed on X3 interrupt timestamp provenance

djibian:
Refs #70.

## Scope

Documentation-only source-governance correction for the integrated X3 pre-LPF observer after reconciling primary firmware, hardware-schematic and Bosch pin-definition evidence.

- exact pinned Crazyflie firmware `54f31e243a0b28b67efef5ba20dbb6d9890a5478` configures BMI088 gyro DRDY on `BMI088_INT_CHANNEL_3`, listens on STM32 `PC14` / `EXTI14`, and stores the sensor callback time in `sensorData.interruptTimestamp`;
- published Crazyflie 2.1 Rev.B hardware at Bitcraze hardware commit `823a6844155633c35d6dfd0919b921c967c0dc79` routes STM32 `PC14` / net `INT_GYR` to BMI088 package pin 1 / `INT2`, while BMI088 pin 12 / `INT3` is shown unconnected;
- Bosch BMI088 datasheet rev 1.9 defines pin 1 / INT2 as an accelerometer interrupt output and pin 12 / INT3 as a gyroscope interrupt output;
- therefore the electrical source of `sensorData.interruptTimestamp` is downgraded to **UNPROVEN** until the exact production-board wiring/revision is reconciled with the pinned firmware configuration;
- `stabilizer.intToOut` remains diagnostic only relative to that firmware interrupt timestamp, not a proven gyro- or accelerometer-producer timestamp;
- the Rev.B schematic also exposes BMI088 pin 16 / INT1 as `INT_ACC` on STM32 `PC13`, but the pinned implementation neither configures accelerometer DRDY nor routes `EXTI13` into the sensor callback, so this is only a future investigation path.

This does not change firmware code, logging payloads, estimator/controller behavior, Runtime, checkpoint profiles, trusted artifacts, parameters or physical-test authority. The #339 CPU-side `x3AccObs.readBeg/readEnd` and `x3BaroObs.readBeg/readEnd` evidence remains valid with its existing non-producer-time semantics.

Exact Draft candidate: `a4b1ff52494dd900e192a2e9a155d81022aa76ea`, based directly on current `main@33a4f1d7dee3243b2444cc5b45e94da69dc5a9a8`. Draft CI is non-decision evidence; a fresh canonical Ready `CI Gate` and independent exact-head review are required before integration.

No `TEST_REQUIRED`, flash, parameter write, estimator reset, commander action or motorized test follows.

PR #344, opened 2026-09-12
Refresh X3 checkpoint artifact boundary

djibian:
Refs #70.

## Scope

Documentation-only checkpoint-support correction after #339/#342 and the integrated X3 roadmap refresh.

- keeps exact #251 as the only currently established physical artifact, without claiming it must remain the future confirmation artifact under all bound models;
- records that #339 is a logging-only source/build diagnostic, not a silent replacement for #251 and not an enabled trusted checkpoint profile;
- makes the two future paths explicit: exact #251 is usable only if deterministic bounds cover its post-LPF acceleration/timing semantics; otherwise a separately identified instrumented artifact/capture/profile must be prepared from the integrated observer;
- records that the existing `run_x3_independent_capture.sh` bundle is #251-specific and does not implicitly capture `x3AccObs` / `x3BaroObs`;
- preserves the hard deterministic-bound, metric-reference/publication, one bounded props-off checkpoint and no-motorized-test boundaries.

No firmware binary, Runtime, estimator/controller, parameter, CI workflow, trusted checkpoint profile or physical authority changes. No `TEST_REQUIRED` follows.

Exact Draft candidate: `2c27ab196070996e39551f0a359d56721d6b970a`, based directly on `main@04fbd3156d58f080e11d44ec47f4133534827d4b`. This execution mutated the candidate and therefore cannot provide independent review.

### DIFF
--- a/experiments/crazyflie-ukf-surface-range/X3_CHECKPOINT_SUPPORT.md
+++ b/experiments/crazyflie-ukf-surface-range/X3_CHECKPOINT_SUPPORT.md
@@ -3,25 +3,27 @@
 This document is the post-#326 preparation boundary for #70. It does **not**
 enable a human-checkpoint profile and it does not request a physical test.
 
-The repository now contains the four offline/acquisition components that were
-missing after the archived-input audit:
+The repository now contains the offline/acquisition components that were missing
+after the archived-input audit:
 
-- continuous raw barometer/IMU acquisition on the unchanged #251 firmware
+- continuous barometer/IMU acquisition on the exact #251 physical firmware
   (`capture_independent_inputs.py`, integrated by #299);
 - the frozen pressure-window calculation (`frozen_pressure_probe.py`, #300);
 - hash-bound conditional external metric-reference envelopes
   (`metric_reference.py`, #311);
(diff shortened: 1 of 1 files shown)


djibian (commented):
GO 2c27ab196070996e39551f0a359d56721d6b970a

Independent exact-head review against current `main@04fbd3156d58f080e11d44ec47f4133534827d4b`; this execution has not mutated PR #344 or its candidate branch.

The one-file checkpoint-support correction resolves the now-stale implication that exact #251 must remain the future X3 confirmation artifact under every deterministic-bound model. It correctly keeps #251 as the only currently established physical artifact, records #339 as logging-only source/build observability rather than physical authority, and makes any future instrumented path require a separate exact artifact identity, compatible capture support and reviewed trusted checkpoint profile. It also preserves #342's UNPROVEN interrupt-source boundary, deterministic bound provenance, metric-reference/publication prerequisites, one bounded props-off checkpoint, and the no-motorized-test boundary.

No firmware, Runtime, estimator/controller, parameter, CI/checkpoint authority or physical-test effect is introduced. Exact Ready canonical run 1299 (`34704747626`) completed successfully with final job `CI Gate` successful; no unresolved review threads were observed.

Issue #345, opened 2026-09-12
Consume fresh Multi-ranger values in the trusted physical run


djibian:
## Goal

Advance #157 from `main@3048dfe532afd1a7bb954827e069302a4bb7a88a` by consuming the already-generic student `range(direction)` intent on the trusted physical Crazyflie path for the **Multi-ranger Deck**, without widening caller authority or conflating deterministic host proof with real-device qualification.

Current main already has backend-neutral `range(front/back/left/right/up)`, intent-dependent exact-AST physical capability/preflight, a live reconnect-sensitive Crazyflie session, exact teacher authorization, powered-session/watchdog/current-program provenance, and trusted physical action composition. The remaining capability gap recorded by `docs/REFERENCE_CAPABILITY_MATRIX.md` is the student physical range-value observation path itself.

Pinned cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873` exposes Multi-ranger values through firmware log variables `range.front`, `range.back`, `range.left`, `range.right`, `range.up`; its official `Multiranger` helper converts values below 8000 mm to metres and treats values at/above 8000 mm as unavailable. Reuse that source contract narrowly; do not expose generic log access to the browser/caller.

## Required boundary

- Support exactly the existing student directions `front/back/left/right/up`; do not add `down`. Flow Deck downward range remains infrastructure-only under current product evidence.
- Derive which direction may be read from the exact teacher-bound canonical program/interpreter demand. Ordinary caller IPC must not be able to substitute an arbitrary log variable, log configuration, conversion rule, connection epoch, freshness proof or authority object.
- Read only the exact corresponding firmware log variable for the requested generic direction. Do not expose a generic physical log-variable API.
- Bind every accepted sample to the exact live Crazyflie and reconnect-sensitive connection epoch, current teacher run, powered session, watchdog and current-program provenance already established by the trusted host.
- Require a **fresh post-request sample**. Cached/pre-existing log values, duplicate/stale samples, samples from another epoch, malformed/non-finite values, unavailable/out-of-range firmware values, log setup failure or timeout must fail closed; they must not become a student sensor value.
- Preserve the official unit contract: firmware millimetres become backend-neutral metres exactly once. Do not silently clamp or manufacture a distance when the firmware reports no usable measurement.
[...]

PR #348, opened 2026-09-12
Consume fresh Multi-ranger values in trusted physical run

djibian:
Refs #345 #157.

This PR is the first bounded increment of the durable #157 physical Multi-ranger decomposition: establish the non-authority fresh same-epoch observer before composing control flow and later effects.

- exact five-direction firmware-log mapping only (`range.front/back/left/right/up`), with no `down` and no generic log-variable surface;
- exact pinned cflib availability contract: `0..7999 mm` converts to metres exactly once; `>=8000 mm` is unavailable and fails closed rather than being clamped or manufactured into a distance;
- exact live `LOG_UINT16` TOC binding for the selected direction;
- fresh post-request observation on the same reconnect-sensitive connection epoch, rejecting cached/pre-existing, duplicate/stale, malformed, timeout, disconnect and epoch-change evidence;
- callback-entry generation fencing so a callback admitted before the read cannot finish late and become fresh evidence;
- serialized cleanup, with uncertain teardown poisoning observer reuse;
- deterministic observer regressions wired into the Runtime v2 core harness;
- no commander, SETPOINT, PARAM-write, branch-selection or execution-authority surface.

This deliberately does **not** claim full #345 or student physical `range(direction)` continuity. The durable #157 boundary requires later increments to compose this observer into exact host-owned expression/control-flow execution (`variables/compare/logic/repeat/if`) and only then drive existing physical effects from the host-selected path while the ordinary caller remains parameter-free. Those later increments are outside this PR.

The earlier #345 `DRAFT_FINDING 931a7e7edc97557de51b0bfd3d4a59709c08a7ad` is addressed in this exact candidate by preserving the pinned `<8000`/`>=8000` semantics and regression coverage. No real-device qualification, motorized flight, checkpoint or `TEST_REQUIRED` is claimed.

Exact candidate HEAD at promotion must receive a fresh canonical Ready `CI Gate` and an independent exact-head review before integration.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -61,6 +61,10 @@ def main():
     if physical_yaw_test.returncode:
         print('FAIL fresh physical yaw observation',file=sys.stderr);print(physical_yaw_test.stdout,file=sys.stderr);print(physical_yaw_test.stderr,file=sys.stderr);return physical_yaw_test.returncode
     print(physical_yaw_test.stdout.strip())
+    physical_range_test=subprocess.run([sys.executable,'tools/ci/test_physical_range_observer.py'],text=True,capture_output=True)
+    if physical_range_test.returncode:
+        print('FAIL fresh physical directional range observation',file=sys.stderr);print(physical_range_test.stdout,file=sys.stderr);print(physical_range_test.stderr,file=sys.stderr);return physical_range_test.returncode
+    print(physical_range_test.stdout.strip())
     physical_watchdog_test=subprocess.run([sys.executable,'tools/ci/test_physical_watchdog_liveness.py'],text=True,capture_output=True)
     if physical_watchdog_test.returncode:
         print('FAIL physical watchdog activation/liveness guard',file=sys.stderr);print(physical_watchdog_test.stdout,file=sys.stderr);print(physical_watchdog_test.stderr,file=sys.stderr);return physical_watchdog_test.returncode
--- a/tools/ci/test_physical_range_observer.py
+++ b/tools/ci/test_physical_range_observer.py
@@ -1,0 +1,363 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+import importlib.util
+import pathlib
+import threading
+
+ROOT = pathlib.Path(__file__).resolve().parents[2]
+MODULE_PATH = ROOT / "tools" / "physical" / "range_observer.py"
+
+spec = importlib.util.spec_from_file_location("webeeblocks_range_observer", MODULE_PATH)
+if spec is None or spec.loader is None:
+    raise RuntimeError("cannot load range observer")
+range_observer = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(range_observer)
+
+
+def require(condition: bool, message: str) -> None:
+    if not condition:
+        raise AssertionError(message)
+
+
+def expect_error(callable_, pattern: str) -> None:
+    try:
+        callable_()
+    except range_observer.RangeReadError as exc:
+        require(pattern in str(exc), f"expected {pattern!r} in {exc!r}")
+        return
+    raise AssertionError(f"expected RangeReadError containing {pattern!r}")
+
+
+class FakeCaller:
+    def __init__(self):
+        self.callbacks = []
+
+    def add_callback(self, callback):
+        self.callbacks.append(callback)
+
+    def remove_callback(self, callback):
+        self.callbacks.remove(callback)
+
+    def call(self, *args):
+        for callback in list(self.callbacks):
+            callback(*args)
+
+
+class FakeTocElement:
+    def __init__(self, direction, *, ctype="uint16_t", pytype="<H"):
+        self.group = "range"
+        self.name = direction
+        self.ctype = ctype
+        self.pytype = pytype
+
+
+class FakeToc:
+    def __init__(self, direction, *, present=True, ctype="uint16_t", pytype="<H"):
+        self.variable = f"range.{direction}"
+        self.element = (
+            FakeTocElement(direction, ctype=ctype, pytype=pytype)
+            if present
+            else None
+        )
+
+    def get_element_by_complete_name(self, name):
+        return self.element if name == self.variable else None
+
+
+class FakeConfig:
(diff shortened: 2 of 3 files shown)


djibian (commented):
DRAFT_REVIEW 33eb163764aea25702cdd500b6bead8c2466c18c

RESOLVED: the applicable `DRAFT_FINDING 931a7e7edc97557de51b0bfd3d4a59709c08a7ad` on #345 is repaired on this exact head. `range_mm_to_m()` now preserves the pinned cflib contract exactly: exact uint16 values `0..7999` convert from millimetres to metres once, while `>=8000` fails closed instead of being clamped or mapped to a synthetic 2 m distance. The deterministic regression covers `7999 -> 7.999`, `8000`, `32767` and `65535`.

The exact-head observer also closes the delayed-callback freshness race at the standalone observation boundary: callback entry snapshots a host-local request generation under `_arrival_lock`, while `read()` increments that generation before taking its baseline; a callback that entered before the read cannot later satisfy the fresh read after blocking behind `_condition`. Same-epoch, strictly later firmware timestamp, exact five-direction TOC typing, disconnect/log errors and uncertain teardown remain fail-closed, and the module exposes no command/PARAM effect surface.

No GO/NO_GO is issued because the PR is correctly still Draft and its own body records the remaining #345 work: trusted-host/interpreter composition, exact teacher-bound direction derivation, reset/effect exclusion and downstream action-authority continuity are not yet implemented on this head.

djibian (commented):
GO 33eb163764aea25702cdd500b6bead8c2466c18c

Independent review against current `main@4c835b3407e66fbbcf266ff07c64e3e653b24ff8` supports this Ready candidate's deliberately bounded first #345 increment.

The applicable earlier #345 refutation `DRAFT_FINDING 931a7e7edc97557de51b0bfd3d4a59709c08a7ad` is resolved on this exact head: the pinned cflib Multi-ranger contract is preserved (`0..7999 mm` -> metres exactly once; `>=8000 mm` unavailable/fail-closed), with deterministic boundary regressions. The observer is restricted to exact `range.front/back/left/right/up` live uint16 TOC entries, binds to one reconnect-sensitive epoch, rejects cached/duplicate/stale/malformed/disconnected/epoch-crossing evidence, and uses callback-entry request-generation fencing so a pre-request callback cannot complete late as current evidence. Uncertain teardown poisons reuse. No generic log API, branch selection, commander/SETPOINT/PARAM effect or execution authority is exposed.

This GO is only for the observer increment claimed by the Ready PR body. It does not establish full student physical `range(direction)` continuity, host-owned expression/control-flow interpretation, later effect selection, real-device qualification, motorized flight, or a human checkpoint. Those remain later #345/#157 work.

djibian (commented):
GO 33eb163764aea25702cdd500b6bead8c2466c18c

Independent exact-head review against current `main@4c835b3407e66fbbcf266ff07c64e3e653b24ff8`; this execution has not mutated PR #348 or its candidate branch.

The bounded claim is valid as an observer-core increment, not as completion of #345. The exact five student directions map only to `range.front/back/left/right/up`; the live TOC is required to match the pinned `uint16_t` representation; values `0..7999` are converted from millimetres to metres exactly once while `>=8000` remains unavailable and cannot become a numeric student value, matching pinned cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873`.

The applicable earlier `DRAFT_FINDING 931a7e7edc97557de51b0bfd3d4a59709c08a7ad` is resolved on this exact head: the synthetic 2 m clamp/sentinel mapping is gone and deterministic coverage includes the exact boundary. Freshness is also fail-closed at the standalone observer boundary: request generation is captured at callback entry, each read advances that generation before taking its baseline, accepted callbacks must carry a strictly later 24-bit firmware timestamp on the same connection epoch, and cached/duplicate/pre-request-late, malformed, disconnect/log-error and epoch-change cases cannot produce a value. Uncertain teardown poisons observer reuse.

`tools/ci/run_runtime_v2_core_harness.py` only adds this deterministic regression as a required fail-closed subtest; it does not weaken or bypass any existing CI oracle. The observer introduces no commander, SETPOINT, PARAM-write, generic log-variable or execution-authority surface.

The PR body correctly keeps trusted-host/interpreter composition, teacher-bound direction derivation, reset/effect exclusion and subsequent-action authority continuity outside this candidate, so this GO does not claim full #345 closure or real-device qualification. No applicable unresolved refutation remains for this exact observer-core candidate.

djibian (commented):
NO_GO 33eb163764aea25702cdd500b6bead8c2466c18c

The code-level observer finding remains resolved, but the current shared PR state no longer matches the candidate scope on which the earlier exact-head GO was issued. PR #348 is currently Ready while its durable body has reverted to the earlier text that explicitly says `Still Draft` and states that trusted-host/interpreter composition, exact teacher-bound direction derivation, reset/effect exclusion integration and downstream action-authority continuity must be completed and proven **before Ready**.

That metadata contradiction is material under the V4 Draft/Ready contract: a Ready candidate must be the stable offered scope actually being validated. The successful Ready `CI Gate` on this head cannot turn a candidate whose current durable description says it is incomplete-before-Ready into an integrable result.

Resolve by choosing one durable scope and making PR state/body agree without changing the code claim implicitly: either return this candidate to Draft for the larger #345 composition, or restore the bounded observer-only Ready scope that explicitly defers host/control-flow continuity to later increments. Any later GO must review the resulting shared candidate state; do not merge while this contradiction remains.

djibian (commented):
GO 33eb163764aea25702cdd500b6bead8c2466c18c

RESOLVED: the applicable `NO_GO 33eb163764aea25702cdd500b6bead8c2466c18c` caused by the transient Ready/body contradiction is no longer applicable. The current durable PR body has restored one coherent bounded Ready scope: this candidate establishes only the fresh non-authority Multi-ranger observer core and explicitly defers trusted-host expression/control-flow composition and later effect selection to subsequent #345 increments.

The source HEAD is unchanged from the previously reviewed observer candidate, the prior `<8000`/`>=8000` cflib-semantics finding remains resolved, and the canonical Ready `CI Gate` run #1318 succeeded on this exact head with final job `CI Gate` success. No unresolved review thread or code-level refutation is known for this bounded candidate.

This GO does not claim full #345 closure, real-device qualification, motorized flight or `TEST_REQUIRED`.

PR #349, opened 2026-09-12
Add read-only X3 STM32 OTP provenance probe

djibian:
Refs #70.

This bounded X3 support slice turns the source-only STM32 OTP provenance finding into deterministic, fail-closed acquisition support without changing firmware or enabling a physical checkpoint.

- reads the exact platform-string OTP layout used by pinned Crazyflie firmware `54f31e243a0b28b67efef5ba20dbb6d9890a5478` (`0x1fff7800`, 16 × 32-byte blocks);
- mirrors the firmware's first-nonzero-block / `0xff` fallback rule without manufacturing fallback identity;
- uses pinned-cflib-compatible `LogConfig.add_memory()` raw-memory observations in groups of at most five descriptors, with three identical observations required per group;
- retains the exact selected 32-byte block and accepts unit-reported platform identity only for an explicitly observed `CF21`; an `R=` field is retained only when actually present;
- keeps `electrical_interrupt_source: UNPROVEN` and `physical_verdict: null`: STM32 OTP identity does not prove BMI088 INT2/INT3 net continuity, sensor interrupt provenance, producer timing or any frozen-predictor bound;
- adds deterministic no-radio tests for selection/fallback, bounded packet grouping, repeated-observation coherence and fail-closed identity;
- runs those tests from the existing isolated X3 pre-LPF diagnostic oracle, without changing the canonical S3 firmware source/artifact identity.

The standalone physical probe is read-only with respect to firmware, parameters and memory; it sends only CRTP log-control traffic needed for raw-memory observation. Normal cflib link close retains its documented safety-zero setpoint boundary. Its standalone runtime fingerprints are explicitly not a complete cflib source/runtime attestation; any later trusted physical use must package and verify the pinned closure separately.

Exact candidate HEAD: `62defe6701ff5d8ccbed9080ebdae8bd1798147e`, based directly on current `main@4c835b3407e66fbbcf266ff07c64e3e653b24ff8`.

No bound is narrowed, no interrupt source is asserted, no checkpoint profile is enabled, and no `TEST_REQUIRED`, flash, parameter write, reset, commander action or motorized test follows. Fresh canonical Ready `CI Gate` and independent exact-head review are required before integration.

### DIFF
--- a/experiments/crazyflie-ukf-surface-range/X3_PLATFORM_PROVENANCE.md
+++ b/experiments/crazyflie-ukf-surface-range/X3_PLATFORM_PROVENANCE.md
@@ -1,0 +1,55 @@
+# X3 STM32 OTP platform provenance probe
+
+This support slice narrows one provenance question discovered while reviewing the
+Crazyflie 2.1 BMI088 interrupt path for #70. It does **not** resolve that interrupt
+wiring contradiction and does not enable a checkpoint.
+
+The exact pinned Crazyflie firmware
+`54f31e243a0b28b67efef5ba20dbb6d9890a5478` reads its platform string from STM32
+OTP at `0x1fff7800`: 16 blocks of 32 bytes, selecting the first block whose first
+byte is non-zero. If there is no such block, or if that first non-zero byte is
+`0xff`, the firmware falls back to `0;CF20`. The X3 probe mirrors only that
+selection rule and retains the observed bytes.
+
+`probe_x3_platform_otp.py` uses pinned-cflib-compatible `LogConfig.add_memory()`
+raw-memory log variables. It deliberately limits each log-control configuration
+to at most five memory descriptors, takes three identical observations of every
+byte group, and fails closed if a group changes or is malformed. It then reads the
+selected 32-byte OTP block and accepts platform identity only when the observed
+STM32 OTP string explicitly reports device type `CF21`. An `R=` field is retained
+only when actually present in those observed bytes.
+
+The probe never flashes firmware, writes parameters, writes memory, resets the
+estimator or invokes a commander. It does send the ordinary CRTP log-control
+packets needed to read memory; normal cflib link close may still emit its documented
+safety-zero commander setpoint. The expected cflib source is
+`45fdb784c9d13074c42835f3b5ac1d12133bf873`, but the standalone script does not
+itself attest the complete installed Python runtime. A later trusted bundle must
+pin and verify that closure before relying on a physical observation.
+
+Most importantly, an observed `0;CF21` or even `0;CF21;R=B1` proves only the
+unit-reported STM32 OTP platform string. It does **not** prove BMI088 INT2/INT3 net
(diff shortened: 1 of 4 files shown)


djibian (commented):
GO 62defe6701ff5d8ccbed9080ebdae8bd1798147e

Independent exact-head falsification review against current `main@4c835b3407e66fbbcf266ff07c64e3e653b24ff8`; this execution has not mutated PR #349 or its candidate branch.

The bounded provenance claim matches the pinned primary sources. Firmware `54f31e243a0b28b67efef5ba20dbb6d9890a5478` reads 16 × 32-byte STM32 OTP blocks beginning at `0x1fff7800`, selects the first block whose first byte is non-zero, and falls back to `0;CF20` when no block is selected or that first non-zero byte is `0xff`. The candidate mirrors that selection but deliberately does not manufacture the fallback identity: only a selected observed block whose device field is exactly `CF21` becomes `EXPLICIT_CF21`; revision data is retained only from an observed `R=` field.

The acquisition surface is also consistent with pinned cflib `45fdb784c9d13074c42835f3b5ac1d12133bf873`: `LogConfig.add_memory()` is the raw-memory log descriptor path, each descriptor costs five bytes in the log-control block, and the candidate limits groups to five while requiring three coherent observations. The complete selection read plus selected-block read uses eleven stopped log blocks, below cflib's 16-block / 128-variable limits for this standalone fresh Crazyflie instance. Deterministic tests cover firmware selection/fallback behavior, explicit identity parsing, bounded grouping, repeated-observation incoherence and scope preservation; the isolated X3 diagnostic oracle invokes those tests without changing the canonical S3 artifact identity.

No electrical interrupt continuity, producer timing, frozen-predictor bound or physical #70 verdict is inferred. The script writes no firmware, PARAM or target memory and exposes no commander action; its normal cflib close-path safety-zero boundary and incomplete runtime-attestation limitation are stated explicitly. No checkpoint or `TEST_REQUIRED` follows from this support slice.

The canonical Ready CI remains a separate merge precondition; it was still in progress at the time of this review.

djibian:
Abandoning this candidate after a post-CI falsification of its physical-read premise. On exact pinned firmware the CRTP protocol is v12, so pinned cflib selects log V2; `LogConfig.add_memory()` still serializes the legacy `type + 32-bit address` shape, while firmware V2 recognizes a memory operation only through the V2 `id == 0xffff` form. At OTP `0x1fff7800`, the proposed packet would be parsed with `id=0x7800`, not as raw memory. The deterministic tests therefore proved only the Python model, not compatibility with the actual pinned transport.

The durable #70 conclusion is recorded in issue comment 5647699358. Current main also already has a reviewed read-only Platform `Get device type name` probe for exact Crazyflie 2.1 model evidence; the unique OTP revision field sought here remains unproven and would not resolve BMI088 net continuity anyway.

This PR stays Draft and is closed rather than repairing around undocumented/raw protocol internals. Its prior GO and successful CI apply only to the superseded exact head and do not rehabilitate the falsified acquisition claim. No checkpoint or `TEST_REQUIRED` follows.

PR #353, opened 2026-09-12
Prove dynamic physical paths safe before takeoff

djibian:
Refs #345 #157.

This bounded infrastructure slice establishes the conservative pre-takeoff safety prerequisite for later trusted-host shared-interpreter composition. It does not execute the shared interpreter or expose physical sensor values.

- consumes only the exact canonical teacher/current-program `astBinding` and emits no CRTP/cflib/commander effect;
- reuses the already integrated move/vertical/turn, wait, horizontal-speed, light-palette and 0.2–1.5 m nominal-altitude bounds rather than inventing a second physical semantics;
- treats both branches of every `if` as reachable and expands each exact bounded `repeat` count, so a sensor/variable decision can never hide an unsafe physical path from pre-takeoff validation;
- tracks a conservative reachable altitude interval through nested branch/repeat composition and rejects before effect if any reachable intermediate path can leave the established physical envelope;
- keeps expression evaluation and branch selection explicitly outside this module: the existing shared Runtime `interpreter.js` remains the sole language evaluator;
- rejects nested takeoff/land, unknown statement kinds, unsupported control metadata and out-of-bounds physical actions fail-closed;
- adds deterministic regressions for representative range/variable/if/repeat structure, unsafe branches, repeated altitude drift, action bounds and exact canonical binding, wired into the Runtime v2 core harness;
- after Ready CI exposed a test-fixture serialization mismatch, the regression now uses the same exact JS-compatible canonical JSON serializer already enforced by the physical AST binding, without changing the production safety proof.

This deliberately does **not** complete #345. Later work must still run the existing shared interpreter inside the trusted host, derive each `range(direction)` only from that bound interpreter, execute the fresh observer under the integrated observation exclusion, and preserve independent downstream effect authority. No caller-facing direction/result/branch/action parameter is introduced here.

No real-device qualification, motorized flight, checkpoint or `TEST_REQUIRED` is claimed. Current candidate HEAD `f934e1614fac83d05b4730e893ded1203ff99080` is based directly on `main@9dd344acd83385b97311855a22ddb3c099928d58`. Promotion requires a fresh canonical Ready `CI Gate` and an independent exact-head review.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -121,6 +121,10 @@ def main():
     if physical_program_sequence_test.returncode:
         print('FAIL exact physical program sequencing',file=sys.stderr);print(physical_program_sequence_test.stdout,file=sys.stderr);print(physical_program_sequence_test.stderr,file=sys.stderr);return physical_program_sequence_test.returncode
     print(physical_program_sequence_test.stdout.strip())
+    physical_dynamic_preflight_test=subprocess.run([sys.executable,'tools/ci/test_physical_dynamic_preflight.py'],text=True,capture_output=True)
+    if physical_dynamic_preflight_test.returncode:
+        print('FAIL conservative dynamic physical pre-takeoff envelope',file=sys.stderr);print(physical_dynamic_preflight_test.stdout,file=sys.stderr);print(physical_dynamic_preflight_test.stderr,file=sys.stderr);return physical_dynamic_preflight_test.returncode
+    print(physical_dynamic_preflight_test.stdout.strip())
     physical_color_led_transport_test=subprocess.run([sys.executable,'tools/ci/test_physical_color_led_transport.py'],text=True,capture_output=True)
     if physical_color_led_transport_test.returncode:
         print('FAIL trusted one-shot bottom Color LED transport',file=sys.stderr);print(physical_color_led_transport_test.stdout,file=sys.stderr);print(physical_color_led_transport_test.stderr,file=sys.stderr);return physical_color_led_transport_test.returncode
--- a/tools/ci/test_physical_dynamic_preflight.py
+++ b/tools/ci/test_physical_dynamic_preflight.py
@@ -1,0 +1,284 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+import json
+from pathlib import Path
+import sys
+
+ROOT = Path(__file__).resolve().parents[2]
+PHYSICAL = ROOT / "tools/physical"
+if str(PHYSICAL) not in sys.path:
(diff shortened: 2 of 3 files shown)


djibian (commented):
GO d1eb893d6948497cc3e11b1ac20b82c802c92e2c

Independent bounded review of the exact Ready candidate. I tried to falsify the claim at the dynamic-control-flow safety boundary: exact canonical binding, reuse of the already-integrated physical action/wait/speed/light bounds, both-branch altitude interval propagation, exact repeat accumulation, nested flight-boundary rejection, and the Runtime-core oracle wiring. The candidate does not evaluate expressions or choose branches, emits no effect, and conservatively rejects any syntactically reachable altitude path leaving 0.2–1.5 m. No applicable unresolved finding found.

This GO is only for the pre-takeoff all-path proof claimed by #353; it does not establish shared-interpreter execution, physical range continuity, real-device qualification or #345 completion.

djibian (commented):
NO_GO d1eb893d6948497cc3e11b1ac20b82c802c92e2c

Ready CI refutes this exact candidate in its own deterministic Runtime-core oracle. The failure is in the fixture builder, not the preflight implementation: `test_physical_dynamic_preflight.py::canonical()` uses `json.dumps(..., sort_keys=True)`, which preserves spellings such as `-90.0`, while the trusted exact AST binding contract uses `takeoff_command._canonical_json()` / JavaScript-number canonicalization and therefore emits `-90`. The candidate test consequently feeds a non-canonical binding and fails before exercising the claimed envelope.

Smallest repair boundary: build positive-test bindings with the existing exact canonicalizer (`takeoff_command._canonical_json`) rather than a second JSON serializer; keep the explicit noncanonical negative fixture separate. Then offer a new HEAD for fresh CI and independent review. This supersedes my earlier GO on this SHA.

djibian (commented):
GO f934e1614fac83d05b4730e893ded1203ff99080

Independent exact-head review of the repaired Ready candidate. The prior NO_GO on d1eb893d6948497cc3e11b1ac20b82c802c92e2c is resolved by the only subsequent code change: positive regression fixtures now use the existing `takeoff_command._canonical_json()` exact JS-compatible AST serializer, while the explicit noncanonical negative fixture remains deliberately separate. No production preflight semantics changed.

I rechecked the bounded claim: exact canonical teacher-bound AST only, no effect surface, both `if` branches conservatively reachable, exact repeat propagation, reuse of integrated movement/wait/speed/light bounds, and rejection of any reachable intermediate nominal altitude outside 0.2–1.5 m. No applicable unresolved finding remains.

This GO is only for #353's pre-takeoff all-path safety proof. It does not establish shared-interpreter execution, physical `range(direction)` continuity, real-device qualification or #345 completion.

PR #355, opened 2026-09-12
Run bound physical control flow through shared interpreter

djibian:
Refs #345 #157.

This bounded infrastructure slice follows the integrated all-path physical preflight from #353 and establishes the host-owned bridge into the **existing** Runtime v2 JavaScript interpreter. It deliberately does not compose a live Crazyflie run yet.

- accepts only the exact canonical teacher/current-program AST binding and runs the integrated conservative dynamic physical preflight before starting the interpreter worker;
- loads `plugins/robot_windows/blockly/webeeblocks/interpreter.js` directly in a private Node child, so the existing shared Runtime interpreter remains the sole language evaluator for variables, arithmetic/comparisons, short-circuit logic, `if`, `repeat` and `range(direction)`;
- exposes no caller-selected direction, sensor value, branch, iteration or action parameter: `run()` is parameter-free after construction from the exact bound AST;
- translates only the interpreter's fixed backend vocabulary over a private UTF-8 stdio protocol with exact monotonic call IDs, exact message shapes, bounded messages and fail-closed protocol-idle timeout;
- validates the fixed movement/vertical/range/light vocabulary and finite numeric arguments again at the host boundary before dispatch; `range(down)` and unknown/malformed calls are rejected before reaching the supplied backend;
- lets only a finite `readRange` number cross back into interpreter data; action/no-effect callback return objects are discarded inside the Python host so they cannot become interpreter authority or branch/cursor state;
- fails closed on malformed/closed/silent child protocol, non-finite range data, backend failure, wrong call order and unsuccessful worker termination;
- adds deterministic Runtime-core regression coverage using the real shared interpreter for variable state, repeated sensor-dependent branching, AND/OR short-circuiting with no phantom range read, exact action trace, unsafe reachable altitude rejection before backend use, language validation before backend use, non-finite range rejection, forbidden `down` substitution and silent-worker timeout.

This deliberately does **not** complete #345. `readRange` is not yet bound to the integrated fresh same-epoch Multi-ranger observer, action callbacks are not yet bound to the existing trusted physical effect consumers, and this adapter does not establish current-program/session/watchdog/exclusion/effect provenance by itself. The next composition step must perform those bindings inside the trusted physical host while preserving each existing downstream authority boundary.

No real-device qualification, motorized flight, human checkpoint or `TEST_REQUIRED` is claimed. Current candidate HEAD `dce66ca0f594c999f49286b5227865dc3fc948cc` is based directly on `main@1e7be7803f3cce1bc2b9c7362abdc747e812cda3`.

### DIFF
--- a/tools/ci/run_runtime_v2_core_harness.py
+++ b/tools/ci/run_runtime_v2_core_harness.py
@@ -125,6 +125,10 @@ def main():
     if physical_dynamic_preflight_test.returncode:
         print('FAIL conservative dynamic physical pre-takeoff envelope',file=sys.stderr);print(physical_dynamic_preflight_test.stdout,file=sys.stderr);print(physical_dynamic_preflight_test.stderr,file=sys.stderr);return physical_dynamic_preflight_test.returncode
     print(physical_dynamic_preflight_test.stdout.strip())
+    physical_shared_interpreter_test=subprocess.run([sys.executable,'tools/ci/test_physical_shared_interpreter.py'],text=True,capture_output=True)
+    if physical_shared_interpreter_test.returncode:
+        print('FAIL trusted shared Runtime interpreter physical core',file=sys.stderr);print(physical_shared_interpreter_test.stdout,file=sys.stderr);print(physical_shared_interpreter_test.stderr,file=sys.stderr);return physical_shared_interpreter_test.returncode
+    print(physical_shared_interpreter_test.stdout.strip())
     physical_color_led_transport_test=subprocess.run([sys.executable,'tools/ci/test_physical_color_led_transport.py'],text=True,capture_output=True)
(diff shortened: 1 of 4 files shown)


djibian (commented):
GO dce66ca0f594c999f49286b5227865dc3fc948cc

Independent falsification review of the bounded #345 adapter slice found no applicable refutation. The candidate keeps `interpreter.js` as the sole language evaluator, runs the integrated all-path dynamic safety validation before backend use, preserves exact interpreter-owned control flow and range demand, exposes no caller-selected semantics, discards action-return authority at the Python boundary, and fails closed on protocol/backend/range anomalies. The follow-up composition boundary is correctly left outside this PR: fresh same-epoch Multi-ranger observation and existing effect-authority consumers still must be bound by the trusted physical host.

PR #356, opened 2026-09-12
Compose shared interpreter with trusted physical backend

djibian:
Refs #345 #157.

Bounded deterministic infrastructure slice for trusted physical `range(direction)` composition. Current exact Ready head: `664f548a19aea14f31103cfd7f520f8f803c6c19`, based directly on `main@f8699533c679cdccb6b298e7c9a935d2d17a2c73`.

This candidate deliberately stops one boundary before production activation wiring. It establishes the process-local components needed for that next step without claiming that `activate_validated_run()` already uses them.

Complete scope of this candidate:
- adds `TrustedDynamicPhysicalBackend`, reached only through the exact bound shared Runtime interpreter, so variables/expressions/short-circuiting/branches/repeats and sensor direction remain interpreter-owned rather than caller-selected;
- keeps the already-completed causal takeoff verification-only: no second takeoff effect is emitted, and the expected height is derived from the exact canonical AST through the integrated conservative dynamic preflight rather than accepted as an auxiliary semantic input;
- adds backend-free `validate_bound_shared_program()` using the same existing JavaScript `Interpreter.validateProgram()` contract, so deterministic shared-language invalidity (including read-before-assignment/unsupported expression semantics) can be rejected before a future causal physical effect without adding a second evaluator;
- binds exact interpreter-selected `front/back/left/right/up` demand to one fresh same-epoch `FreshRangeObserver` inside the shared physical observation exclusion, with immediate current-program, teacher, watchdog and connection-epoch checks before/after the sample and fail-closed teardown;
- preserves range values as finite non-authority data while selected move/turn/vertical/light/wait/speed/landing operations delegate to the existing trusted physical consumers;
- adds `TrustedDynamicControlledLandingTransport`, whose initial nominal altitude is derived from the exact teacher-bound AST/preflight, advances only after definitively accepted vertical effects, and builds terminal command-10 descent from the completed interpreter-selected runtime path while retaining the existing acknowledgement and fresh landing-completion chain;
- cross-checks the dynamic landing transport's AST-derived altitude root against the backend's independently re-derived exact preflight height before interpreter progression, rejecting any split root before a range/action/landing can run;
- adds a one-shot `BoundDynamicPhysicalRun` owner around the exact shared-interpreter/backend pair, with fail-closed worker failure and explicit observer teardown;
- uses the repository's Runtime-compatible canonical AST serializer in positive backend fixtures; stale auxiliary altitude constructor inputs have been removed from both production and test surfaces;
[...]

PR #357, opened 2026-09-12
Derive path-invariant dynamic terminal landing

djibian:
Refs #345 #157.

This narrow path-invariant landing fallback is superseded by the more complete runtime-altitude design now durably present on Draft #356 at exact head `e1ac7c1001bb3e9d601597870f0f76ebd3e10999`.

The experiment established one useful boundary: an immutable AST can safely pre-derive terminal landing altitude only when all reachable dynamic paths end at the same nominal altitude; branch-dependent terminal altitude must not be guessed from static structure. However, #356 now owns the stronger required solution: `TrustedDynamicControlledLandingTransport` keeps host-local nominal altitude, advances it only after definitively accepted interpreter-selected vertical effects, and derives terminal descent from that completed runtime state while preserving the existing effect authority/completion chain.

Keeping this separate fallback would add a second dynamic-landing path through the legacy flat landing command without being required by the product boundary. The branch and commits remain durable historical evidence, but this candidate is intentionally not integrated.

No real-device qualification or `TEST_REQUIRED` follows.

### DIFF
--- a/tools/ci/test_physical_dynamic_preflight.py
+++ b/tools/ci/test_physical_dynamic_preflight.py
@@ -10,6 +10,7 @@ PHYSICAL = ROOT / "tools/physical"
 if str(PHYSICAL) not in sys.path:
     sys.path.insert(0, str(PHYSICAL))
 
+import landing_command  # noqa: E402
 import takeoff_command  # noqa: E402
 from physical_dynamic_preflight import (  # noqa: E402
     DynamicPhysicalPreflightError,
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO b4acda3bf4c4c58c5f646016fb3d3871aef30357

Independent review of the exact Ready head against current main/contract found the bounded claim coherent: flat landing behavior stays unchanged; dynamic fallback is accepted only when conservative pre-takeoff proof establishes one path-invariant terminal nominal altitude; branch-dependent terminal altitude remains fail closed; no runtime sensor/branch/caller value can select command 10. The canonical dynamic-preflight harness covers altitude-neutral dynamic control, equal vertical terminal branches, and explicit branch-dependent rejection. No applicable unresolved refutation found.

PR #361, opened 2026-09-12
Add resumable exact-AST shared interpreter session

### LINKED ISSUE #345: Consume fresh Multi-ranger values in the trusted physical run

djibian:
Refs #345 #157.

Draft infrastructure slice for the next trusted-host composition boundary after #356. The existing durable branch is intentionally preserved as-is; this PR targets current `main@f198e5fafbefe49e5ce2bedcf842beda2cb5b9eb`, so the eventual merge result composes the resumable session with the just-integrated dynamic backend infrastructure rather than rewriting concurrent history.

Intended bounded scope:
- keep the existing CommonJS Runtime `interpreter.js` authoritative for variables, expressions, short-circuiting, branches and repeats;
- run one exact canonical teacher-bound AST in a private host-owned Node session;
- resolve exact interpreter-selected `range(direction)` requests through a trusted host callback as non-authority data;
- pause only at the next exact action selected by the interpreter and expose an opaque process-local `SelectedPhysicalAction` whose node/path/payload are re-derived from the exact bound AST;
- require exact claim identity before interpreter progression, and poison the session on rejection/ambiguity/worker loss;
- preserve the ordinary caller boundary as parameter-free in the later production composition; this Draft itself exposes no browser/HTTP/CRTP effect surface.

This is not production activation wiring, does not replace `execute-next-inflight`, does not close #345, and establishes no real-device qualification or `TEST_REQUIRED`.

Known Draft review debt before Ready:
1. `test_local_surface_exposes_no_browser_or_physical_effect_api()` currently reads the compatibility wrapper and the already-integrated `tools/physical/shared_interpreter_worker.js`, not the new implementation/worker under `tools/session/`. The oracle must inspect the actual new files so it cannot miss a duplicated evaluator or authority leak.
2. `SharedInterpreterSession` launches the child with `stderr=subprocess.PIPE` but never drains stderr; the private protocol must not acquire an avoidable child-process deadlock path (use a non-blocking/drained sink or equivalent fail-closed handling).

Further review must also verify the resumable session consumes #356's pre-effect shared-language validation contract before any future causal takeoff, and that canonical Runtime CI exercises the corrected session oracle before promotion.

### DIFF
--- a/tools/ci/test_physical_dynamic_preflight.py
+++ b/tools/ci/test_physical_dynamic_preflight.py
@@ -15,6 +15,7 @@ from physical_dynamic_preflight import (  # noqa: E402
     DynamicPhysicalPreflightError,
     validate_bound_dynamic_program,
 )
+import test_physical_shared_interpreter_session as shared_interpreter_test  # noqa: E402
 
 
 def require(condition: bool, message: str) -> None:
@@ -271,6 +272,7 @@ def main() -> int:
     test_all_existing_action_bounds_apply_inside_dynamic_control_flow()
     test_control_flow_metadata_cannot_hide_unsupported_fields()
     test_noncanonical_or_malformed_binding_fails_closed()
+    shared_interpreter_test.main()
     print(
         "PASS dynamic physical preflight: every reachable control-flow path "
         "stays inside integrated action and altitude bounds before effect"
--- a/tools/ci/test_physical_shared_interpreter_session.py
+++ b/tools/ci/test_physical_shared_interpreter_session.py
@@ -1,0 +1,242 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+from pathlib import Path
+import sys
+import tempfile
+
+ROOT = Path(__file__).resolve().parents[2]
+PHYSICAL = ROOT / "tools" / "physical"
+if str(PHYSICAL) not in sys.path:
+    sys.path.insert(0, str(PHYSICAL))
+
+import takeoff_command  # noqa: E402
+from shared_interpreter_session import (  # noqa: E402
+    SelectedPhysicalAction,
+    SharedInterpreterSession,
+    SharedInterpreterSessionError,
+)
+
+
+def require(condition: bool, message: str) -> None:
+    if not condition:
+        raise AssertionError(message)
+
+
+def expect_error(callable_, pattern: str) -> None:
+    try:
+        callable_()
+    except SharedInterpreterSessionError as exc:
+        require(pattern in str(exc), f"expected {pattern!r} in {str(exc)!r}")
+        return
+    raise AssertionError(f"expected SharedInterpreterSessionError containing {pattern!r}")
+
+
+def canonical(program: list[dict[str, object]]) -> str:
+    return takeoff_command._canonical_json(
+        {"version": 1, "semantics": "webeeblocks-ast-v1", "program": program}
+    )
+
+
+def ref() -> dict[str, str]:
+    return {"id": "v-range", "name": "distance"}
+
+
+def number(value: float) -> dict[str, object]:
+    return {"kind": "number", "value": value}
+
+
+def program() -> list[dict[str, object]]:
+    variable = ref()
+    return [
+        {"kind": "takeoff", "height_m": 0.8},
+        {
+            "kind": "set_variable",
+            "variable": variable,
+            "value": {"kind": "range", "direction": "front", "unit": "m"},
+        },
+        {
+            "kind": "if",
+            "condition": {
+                "kind": "logic",
+                "op": "AND",
+                "left": {
+                    "kind": "compare",
+                    "op": "LT",
+                    "left": {"kind": "variable_get", "variable": variable},
+                    "right": number(1.0),
+                },
+                "right": {
+                    "kind": "compare",
+                    "op": "LT",
+                    "left": {"kind": "range", "direction": "right", "unit": "m"},
+                    "right": number(1.0),
+                },
+            },
+            "then": [{"kind": "set_light", "color": "green"}],
+            "else": [{"kind": "set_light", "color": "red"}],
+        },
+        {
+            "kind": "repeat",
+            "count": 2,
+            "body": [{"kind": "move", "direction": "forward", "distance_m": 0.3}],
+        },
+        {"kind": "land"},
+    ]
+
+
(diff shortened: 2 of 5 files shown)


djibian (commented):
DRAFT_FINDING 65fe8a8c049ebae256407df2bed995cd349b0850 — the current slice introduces a second resumable interpreter/cursor after #356 instead of consuming the substrate that #356 integrated for the final #345 boundary.

Current `main@f198e5fafbefe49e5ce2bedcf842beda2cb5b9eb` already contains `TrustedDynamicPhysicalBackend` + `BoundDynamicPhysicalRun`, whose exact bound `interpreter.js` execution owns range/control-flow/action selection and delegates every selected effect through the established trusted consumers. The owner’s post-#356 #345 evidence names the remaining substantive work as **production activation/recovery wiring**: pre-effect `validate_bound_shared_program()`, then compose that integrated dynamic run after the sole causal takeoff, and prove post-takeoff interpreter/range/protocol/observer failure enters terminal recovery.

This Draft instead adds another private Node worker plus `SharedInterpreterSession`, pauses at every interpreter-selected action, and exports `SelectedPhysicalAction` as a new process-local cursor that a later host layer must dispatch. That leaves the integrated #356 backend/run unused for action composition and creates another semantic/progression boundary between `interpreter.js` and the trusted effects—the exact competing-cursor surface the final #345 note says to remove. The currently parameter-free `execute-next-inflight` IPC does not by itself establish a product requirement that one caller request must correspond to one interpreter action; it only establishes that ordinary IPC cannot supply semantics/authority.

Smallest repair boundary before investing in the known local test/stderr debts: either (a) identify durable product/technical authority that actually requires resumable per-action IPC and then show how this adapter reuses, rather than bypasses/duplicates, #356’s dynamic backend/validation/recovery contracts; or (b) treat this branch as superseded and wire the already-integrated `BoundDynamicPhysicalRun` directly into `activate_validated_run()` with deterministic recovery proof. Until one of those is established, this 800-line infrastructure slice is not yet a justified prerequisite for #345 and risks widening the trusted semantic surface unnecessarily.

Independent code-level findings already listed in the Draft body (wrong files inspected by the local-surface oracle; undrained child stderr) remain applicable if this architecture is retained. No real-device checkpoint or `TEST_REQUIRED` follows.

PR #363, opened 2026-09-12
Correct X3 timing provenance for pinned firmware clock

djibian:
Refs #70.

The X3 logging/provenance documentation currently describes `usecTimestamp()`-derived values as microseconds on the exact firmware pinned by the #70/#251 work. Upstream Bitcraze PR #1684 is based exactly on that same firmware SHA, `54f31e243a0b28b67efef5ba20dbb6d9890a5478`, and establishes a TIM7 prescaler off-by-one: the programmed counter advances at 84/85 of the intended nominal 1 MHz rate. Its current exact upstream head is `bb58cadfa348f903e4e8e6995a5557c42cd5754d` and it has two upstream approvals.

This documentation-only correction:
- records that `x3AccObs.readBeg/readEnd`, `x3BaroObs.readBeg/readEnd` and `stabilizer.intToOut` inherit that exact `usecTimestamp()` clock-scale defect on the pinned firmware;
- records the deterministic divider-only correction `raw_delta * 85 / 84` after modular subtraction;
- keeps oscillator accuracy, BMI088 interrupt-source provenance, producer/filter/bus/scheduling uncertainty and `sensor_time_error_s` explicitly separate and unresolved;
- distinguishes the wireless FreeRTOS log-packet timestamp from the TIM7 clock so the two are not silently treated as one proven clock domain;
- preserves the existing X3 `UNPROVEN`, no-motorized and no-checkpoint boundaries.

No firmware binary, runtime behavior, estimator/controller equation, parameter, physical action, checkpoint mechanism or human boundary changes. No `TEST_REQUIRED` follows from this candidate.

### DIFF
--- a/experiments/crazyflie-ukf-surface-range/INDEPENDENT_INPUT_CAPTURE.md
+++ b/experiments/crazyflie-ukf-surface-range/INDEPENDENT_INPUT_CAPTURE.md
@@ -30,17 +30,26 @@ Every field is fetched as a float; each block fits cflib's 26-byte log payload.
 Units remain the pinned firmware units: barometer altitude in m, pressure in
 mbar, temperature in Celsius; accelerometer in g; filtered gyro in degrees/s;
 downward range in mm; roll/pitch in degrees; UKF Z in m and VZ in m/s.
-`stabilizer.intToOut` is the pinned firmware's microsecond latency from the IMU
-interrupt timestamp carried by the stabilizer's current `sensorData` to the end
-of that stabilizer iteration. Diagnostic UKF outputs and this latency diagnostic
-must not become independent displacement inputs.
+`stabilizer.intToOut` is a difference of two values from the pinned firmware's
+`usecTimestamp()` clock, from the interrupt timestamp carried by the stabilizer's
+current `sensorData` to the end of that stabilizer iteration. The pinned
+`2026.08@54f31e...` TIM7 prescaler has a known 84/85 nominal-rate defect documented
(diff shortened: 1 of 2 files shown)


djibian (commented):
GO b6e0aa1b2f2d7df35c52be1e3228c5c3952d7a79

Independent exact-head documentation review against current `main@f18ae12dfee054b0f142202c023fecd3c9da2d2c` finds the bounded provenance correction supported and conservative.

The cited upstream Bitcraze PR #1684 is currently open at exact head `bb58cadfa348f903e4e8e6995a5557c42cd5754d` and is based on the exact firmware SHA pinned by this repository, `54f31e243a0b28b67efef5ba20dbb6d9890a5478`. Its source/evidence states that TIM7 is programmed with PSC=84 while the STM32 timer divides by PSC+1, so the affected `usecTimestamp()` counter advances at 84/85 of intended nominal rate; the divider-only interval correction `raw_delta * 85 / 84` is therefore the correct deterministic scale repair after modular subtraction. The upstream evidence also explicitly distinguishes wireless logging as using the FreeRTOS tick timestamp path and being unaffected by that TIM7 defect.

This candidate keeps that correction narrowly scoped: it does not convert the divider ratio into an absolute clock-accuracy guarantee, does not collapse interrupt/sensor-producer/filter/bus/scheduling uncertainty into the timer correction, does not treat wireless log timestamps as the same clock domain, and leaves the X3 independent-displacement result `UNPROVEN`. No estimator/controller/runtime/firmware effect or human acceptance boundary changes.

Exact Ready run #1372 (`34720057146`) publishes successful canonical `CI Gate` on this head, including the bounded X3 observer diagnostic. No `TEST_REQUIRED` follows.

PR #350, opened 2026-09-12
Prepare hermetic Boost support for historical Webots CI

djibian:
Refs #303.

Hermetic Boost-1.74 support for the historical Webots CI jobs. This remains bounded CI-support work; it does not change Runtime/product behavior, human checkpoint authority or physical-flight authorization.

Current exact candidate: `54f66b231a643935c2b7c29911ba436e0e1f4ceb`, synchronized with `main@af2d43b0849e08f32d9b60aaa6360be2282ab0ae`.

Established by the branch design and exact-head Draft evidence:
- exact Jammy `libboost1.74-dev_1.74.0-14ubuntu3_amd64.deb` filename, size, SHA-256 and `BOOST_VERSION=107400` are pinned;
- `prepare_historical_boost.py` has no network/package-manager fallback and reconstructs the exposed Boost tree from the verified archive on every use;
- the canonical historical consumer is restore-only, exact-keyed and fail-closed on cache miss;
- public acquisition is confined to non-decision provisioning: scheduled/manual maintenance runs, plus an internal same-repository Draft PR bootstrap; Ready PR acceptance does not acquire from the public mirror;
- all five affected historical Webots jobs use the single verified offline builder inside `cyberbotics/webots:R2025a-ubuntu22.04`, preserving their product oracles while removing the live `apt-get update` / `libboost-dev` critical path;
- the Linux header-only regression is scoped away from the Windows-only `-lboost_*` Makefile branch.

[...]

Issue #303, opened 2026-09-11
Make historical Webots CI independent of live apt-index latency


djibian:
## Problem

Required Ready `CI Gate` evidence is repeatedly being lost before product oracles run because several historical Webots jobs execute live `apt-get update` / package installation inside the pinned `cyberbotics/webots:R2025a-ubuntu22.04` container under short job-level time bounds.

Durable observations:

- PR #299, exact HEAD `f681639ed463a8a0ea3903e89919436d7f0c8ba6`, canonical run `34591532059`: `encoders-historical` job `103237727710` and `historical-blockly-ui` job `103237727796` were cancelled while Ubuntu archive/security dependency acquisition was still running, before their product test commands. This is already recorded on #299.
- PR #276, exact HEAD `1da5355829312ba7525dff03a84652d1071a3c9d`, canonical run `34599884420`: `encoders-historical` job `103264376370` again reached the pinned Webots image successfully, then spent the remaining job budget in `apt-get update`; it was cancelled while fetching `jammy-updates/restricted`, before building or executing the encoder product oracle.

This is missing CI evidence caused by mutable external dependency latency, not a product failure and not a reason to treat an older green run as current evidence.

## Goal

Make the affected required historical Webots jobs reproducible enough that their bounded job time is spent on the actual product oracle rather than fetching mutable Ubuntu package indexes.

Prefer a smallest deterministic dependency closure: reuse/prepackage the exact required build/runtime dependencies, a repository-controlled/pinned support image or another verifiable offline/cacheable mechanism. Preserve the official pinned Webots R2025a execution environment and the existing product tests.

## Acceptance criteria

- The affected historical jobs no longer require a fresh mutable `apt-get update` from public Ubuntu mirrors inside their critical timed path.
- Required added dependencies are pinned/verifiable or supplied by a deterministic closure whose provenance is reviewable.
- Existing historical test semantics, Webots R2025a pin, result oracles and fail-closed behavior are unchanged.
- Do not solve this primarily by weakening assertions, skipping required suites, converting cancellation to success, or inflating timeouts until network latency happens to fit.
- A deterministic regression/contract proves the new dependency path cannot silently fall back to ambient mutable packages.
- `CI Gate` remains the required branch-protection context; this work does not create a second acceptance path.

## Boundary

[...]

PR #368, opened 2026-09-12
Pre-provision exact Boost support outside canonical CI

djibian:
Refs #303.

Superseded by the current #350 implementation; closed without merge.

The useful architectural conclusion from this Draft is preserved in #350: the canonical PR path must only restore an exact pre-provisioned Boost cache and fail closed on cache miss, while public acquisition is permitted only outside pull-request acceptance. Current #350 keeps that maintenance provisioning inside the already-authorized V4 `ci.yml` scheduled/manual path, so this additional workflow no longer provides unique useful work.

This candidate also cannot be integrated under the current V4 workflow contract because it adds a new workflow outside the exact expected workflow set (`test_workflow_contract.py`). Its earlier `push` trigger was removed on head `88d60ec5b901cf7ba1dcf7f516ea6fa36e572568`, but the extra-workflow violation remains.

No branch deletion is implied by closing this PR; its source ref remains available until a later reconstruction establishes that deletion is safe and useful.

### DIFF
--- a/.github/workflows/provision-historical-boost.yml
+++ b/.github/workflows/provision-historical-boost.yml
@@ -1,0 +1,69 @@
+name: Provision historical Boost cache
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "17 3 * * 0"
+  push:
+    branches:
+      - main
+    paths:
+      - .github/workflows/provision-historical-boost.yml
+
+permissions:
+  contents: read
+
+jobs:
+  provision:
+    name: Provision exact Boost 1.74 archive
+    runs-on: ubuntu-22.04
+    timeout-minutes: 5
+
+    steps:
+      - name: Restore exact repository cache
+        id: boost-cache
+        uses: actions/cache/restore@v4
+        with:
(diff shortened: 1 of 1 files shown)


djibian:
SUPERSEDED 88d60ec5b901cf7ba1dcf7f516ea6fa36e572568

Current `main@a43b368df20ba23c0934182a52050280eebfba4d` and current #350 Draft `a2aaaa778b99122180f63167e4348a5b2f8f064e` make this extra-workflow slice unnecessary. #350 now uses the already-authorized canonical `ci.yml` schedule / `workflow_dispatch` maintenance path to restore, acquire only outside pull-request acceptance, verify and save the exact Boost archive cache; its canonical consumer action remains restore-only and fail-closed. That preserves this PR's useful pre-provisioning purpose without widening the V4 workflow inventory, which is the only remaining failing contract on this head.

No unique #303 product or support work depends on keeping this additional workflow. The #303 goal remains engaged in #350, where the five historical Webots jobs still need to be wired to the verified offline builder. Closing this PR as superseded does not abandon #303.

PR #371, opened 2026-09-13
Attach Color LED Deck to the Crazyflie body

### LINKED ISSUE #370: [Bug] Color LED Deck is visibly detached from the Crazyflie in Webots
Refs #157 #235.

## Human observation

On the real Webots R2025a Windows classroom path, the simulated bottom Color LED Deck is visibly separated from the Crazyflie instead of appearing physically mounted to it.

The deck follows the Crazyflie because it is inside the `extensionSlot`, so this is not a parent/attachment-lifecycle failure. It is a visual geometry/offset defect.

The current Runtime world places the Color LED visual section at `translation 0 0 -0.006` and represents four mounting posts with `size 0.0022 0.0022 0.0065`. In the real 3D view this geometry does not visually meet the Crazyflie body/deck mounting surface.

## Why existing evidence missed it

PR #235 explicitly claimed a bottom deck “visibly attached to the Crazyflie through four mounting posts”. The current structural oracle `tools/ci/test_color_led_deck_visual.py` checks for the hard-coded translation, four post tokens, deck/diffuser dimensions and light elements, but does not establish actual visible contact with the rendered Crazyflie mesh. A structurally present set of posts can therefore pass while a visible gap remains.

## Required repair boundary

Keep the existing generic `set_light` semantics, `color_led` device identity, diffuser/halo behavior and bottom-mounted design. Correct only the visual mounting geometry and strengthen evidence so the same defect cannot silently regress.

Acceptance should establish:

1. in normal side and three-quarter Webots views, the Color LED Deck is visibly mounted under the Crazyflie with no unintended air gap;
2. the deck remains bottom-mounted and follows the robot rigidly through movement;
3. OFF/colored LED rendering remains visible and controlled by the existing `color_led` device;
4. both Runtime worlds remain geometrically consistent;
5. automated/render evidence is strengthened beyond token existence so a visibly detached deck is rejected.

djibian:
Closes #370.

## Repair

Move the existing bottom Color LED visual assembly upward as one rigid unit in both Runtime worlds, from `z=-0.006` to `z=0.0066` in the Crazyflie `BODY` / `extensionSlot` coordinate system. The internal PCB/post/diffuser geometry is unchanged, so the existing `color_led` device, `set_light` semantics, diffuser/halo rendering and four-post bottom-mounted design are preserved.

The corrected offset is tied to the pinned Webots R2025a Crazyflie mesh: `cf_body.001` is translated to `z=0.015` and its PCB vertices extend to local `z=-0.001246929`, giving a body PCB underside at `z=0.013753071`. With the existing 6.5 mm posts centered 4 mm above the deck PCB, the new assembly position puts the post tops at `z=0.01385`, a ~0.097 mm overlap that removes the render seam without materially penetrating the body.

## Regression evidence

Strengthen `test_color_led_deck_visual.py` beyond token existence. It now derives the deck PCB and post endpoints from the actual world geometry and rejects either a gap to the pinned R2025a body mounting plane or excessive interpenetration. It also proves the posts still meet the deck PCB and the circular diffuser still meets its underside, while retaining the existing two-world identity and LED/render contract checks.

Local syntax + geometric-oracle execution passes on the repaired section. CI still owns the official R2025a fixed-view render evidence (top / three-quarter / side, OFF + blue).


djibian (commented):
GO e9c2fbb739e60793b20d768eb77f8fbfc357ed78

Independent exact-head review against current `main@a43b368df20ba23c0934182a52050280eebfba4d`: the three-file repair stays inside #370's visual-geometry boundary. Both Runtime worlds move the existing Color LED assembly as one rigid unit to `z=0.0066`; device identity, `set_light` semantics, diffuser/halo structure and the shared Runtime authority path are unchanged.

The strengthened oracle closes the defect that the prior token-only check missed. With the candidate geometry, the 1.5 mm deck PCB top is `0.00735 m`; each 6.5 mm post starts exactly there and ends at `0.01385 m`. Against the pinned R2025a body PCB underside `0.015 - 0.001246929 = 0.013753071 m`, that is a bounded `0.000096929 m` overlap: no air gap and well below the candidate's 0.25 mm excessive-penetration guard. The diffuser top remains exactly coincident with the deck PCB underside (`0.00585 m`). The oracle also requires the two Runtime visual sections to stay identical and preserves the existing LED/render contract tokens.

No product-authority, physical-flight, CI-trust or human-checkpoint boundary is widened, and no applicable unresolved refutation was found. This GO applies only to exact Ready head `e9c2fbb739e60793b20d768eb77f8fbfc357ed78`; merge still requires a successful exact-head canonical `CI Gate`, including the existing official Webots R2025a fixed-view OFF/blue render proof.

Issue #369, opened 2026-09-13
[Bug] Make Windows Robot Window startup deterministic and package all referenced UI assets


djibian:
Refs #81 #87 #338.

## Human evidence

On the Windows 11 low-end reference path using Webots R2025a and the exact #338 artifact (`WebeeBlocks-Windows-R2025a`, artifact id `10299471299`, target `0593ed990f495b5c6bd5f08945010748c0d34b7e`), both Firefox and Chrome showed nondeterministic/broken Robot Window startup before cache intervention.

Observed failure modes included:

- Firefox loading multiple `.js` resources from `http://localhost:1234/...` with cached/inconsistent `text/css` content/MIME, producing syntax errors and leaving the UI at `INITIALISATION`;
- Chrome exposing native File System Access (`showOpenFilePicker` / `showSaveFilePicker` both functions) but `project_ui.js` failing with `ReferenceError: workspace is not defined`, while `main.js` was present as a script tag and a later explicit no-cache fetch returned the correct `200 application/javascript` body;
- after disabling browser cache in DevTools and reloading the Robot Window, WebeeBlocks initialized successfully in both Chrome and Firefox; Chrome selected `browser-native`, and Firefox direct `.wbb` file management also worked. This narrows the Firefox defect away from the native project-file broker itself and toward startup/resource-cache consistency.

The original #338 FAIL remains valid because cache disabling/reload was extra procedure not allowed by that checkpoint.

## Independent packaging defect

The Windows classroom archive also omits two files that `plugins/robot_windows/blockly_v2/blockly_v2.html` references directly:

- `led_observability.js`
- `physical_preflight_runtime.js`

Both exist in the repository, but `tools/build_windows_classroom_release.ps1` copies a fixed `blockly_v2` file list that does not include them. Real Chrome/Webots consequently reports 404 for both resources.

## Required repair boundary

Preserve the validated Chrome File System Access path and the integrated Firefox Qt/WWI broker semantics. Repair only the Windows Robot Window distribution/startup boundary and release completeness.

Acceptance should establish at minimum:

1. every local `src`/`href` resource referenced by `blockly_v2.html` is present in the Windows release archive (with an automated archive-contract check, not a hand-maintained assumption);
2. a normal one-action Windows launch reaches a usable Runtime without DevTools, cache disabling or manual reload;
3. initial resource responses are the correct bytes/types, with no stale CSS/JS substitution or equivalent cache poisoning;
[...]

PR #373, opened 2026-09-13
Make Windows Robot Window startup cache-safe and package-complete

djibian:
Refs #369.

Bounded Windows classroom distribution/startup repair based on the exact #338 human failure evidence.

The release builder now:
- keeps an explicit Runtime-v2 root asset closure and admits the two direct dependencies that were missing from the failed archive: `led_observability.js` and `physical_preflight_runtime.js`;
- deliberately excludes unrelated build-time root helpers such as `prepare_blockly_vendor.js` instead of sweeping every sibling `.js`/`.css`/`.html` file into the classroom ZIP;
- resolves every local `src`/`href` from the packaged `blockly_v2.html`, rejects non-local, escaping, or missing dependencies, and fingerprints each packaged reference from the exact asset SHA-256 bytes;
- derives the packaged top-level Robot Window directory/HTML name from the complete packaged `plugins/robot_windows` byte closure, so nested imports and the top-level HTML move to a new localhost URL when Runtime/plugin bytes change;
- rewrites the packaged world to select that exact content-addressed Robot Window identity, avoiding reuse of Webots R2025a's hour-cached top-level `blockly_v2.html?name=...` URL across artifact changes;
- keeps no-cache document directives as defense in depth while remaining fully offline.

The release oracle independently reconstructs the canonical pre-rename plugin tree, recomputes the digest, requires exactly one `blockly_v2_<16-hex>` Robot Window, and proves the packaged world selects it. The static contract preserves #374's direct-HTML-resource omission oracle and the bounded root-package boundary.

This preserves the existing Chrome File System Access implementation, Firefox Qt/WWI broker semantics, Runtime execution, physical authority and checkpoint contract; it changes only Windows release completeness/resource-cache identity.

Current exact candidate: `4e0b2da908986e0152f210a4db43305e64bdb1ee`, synchronized with `main@77e4783cdfc91c23726bba1866b0799e2b262f97`. The earlier `NO_GO acef710e...` broad-package finding and `DRAFT_FINDING bcd723c...` top-level-cache finding are addressed by the bounded explicit closure and content-addressed top-level identity respectively.

After integration, a fresh exact-artifact Windows Firefox checkpoint remains required for #369/#87; #338 remains an authoritative FAIL for its older artifact only. Keep #369 open until that checkpoint resolves PASS, FAIL, or valid NOT_NEEDED.


djibian (commented):
NO_GO acef710e62be9b79ac19091f771a5f9f3e800017

The content-addressed `src`/`href` rewrite is a useful repair for the #338 cache boundary, and the candidate correctly detects a referenced asset that is missing from the assembled package. However, the new root-asset copy rule is broader than the Runtime dependency closure it claims to package:

`Get-ChildItem ... | Where-Object { $_.Extension -in @('.html', '.css', '.js') }`

copies every root JavaScript helper into the classroom ZIP, including `plugins/robot_windows/blockly_v2/prepare_blockly_vendor.js`. That file is explicitly a build-time Node preparation tool: it imports `node:fs`/`node:path`, consumes `node_modules`, rewrites the generated `vendor` tree and checks the Blockly package. It is not referenced by `blockly_v2.html` and was deliberately outside the previous runtime archive. The current release oracle does not reject this extra development helper, so this Ready candidate weakens the package boundary while fixing completeness.

Smallest repair boundary: package the actual Robot Window runtime dependency closure rather than all sibling files. For example, keep an explicit shell/runtime-root set with the two missing modules and let the new fail-closed HTML dependency resolver prove completeness, or derive/copy the root assets from the local `src`/`href` references themselves. The resulting archive check should establish that every referenced local asset exists without admitting unrelated build helpers merely because they share `.js`/`.css`/`.html` extensions.

This finding is independent of the later human acceptance requirement: after the packaging/startup repair is corrected and passes exact-head CI/review, #369/#87 still need a fresh exact-artifact Windows Firefox checkpoint.

djibian:
DRAFT_FINDING bcd723c31e33c29bb6857f9840717e788fe8618d

The earlier archive-boundary finding is resolved on this Draft head: the root runtime set is explicit again, admits the two missing HTML dependencies, and no longer sweeps build-only helpers into the classroom ZIP.

One startup/cache boundary is still unresolved. The pinned official Webots R2025a implementation opens the Robot Window at the stable top-level URL `http://localhost:<port><windowFile>?name=<robot>` (`src/webots/gui/WbRobotWindow.cpp`, `WbRobotWindow::setupPage`). Its HTTP server sends `Cache-Control: public, max-age=3600` for both `200` and `304` file replies (`src/webots/core/WbHttpReply.cpp`, `WbHttpReply::forgeFileReply`). Therefore an already-cached `blockly_v2.html?name=Crazyflie%20Runtime` can be reused for up to one hour before the new packaged HTML is fetched at all.

The new content-derived `?wb=<digest>` keys are a strong repair for subresources once the new HTML has loaded, but the three `<meta http-equiv>` directives live inside that new document and cannot invalidate the previously cached top-level response that prevents reaching those new URLs. This leaves the exact #338/#369 failure mode materially possible on a normal one-action launch after switching artifacts, which is the acceptance boundary that forbids DevTools/cache disabling/manual reload.

Smallest repair boundary: make the top-level Robot Window cache identity change deterministically with the packaged Runtime release (for example a content-addressed packaged Robot Window path/name, or another repository-controlled mechanism that causes official R2025a to request a fresh top-level URL/origin) while preserving offline use and the existing Chrome/Firefox file backends. Keep the subresource fingerprints. Add a source/packaging oracle that proves the top-level identity derives from the packaged content rather than relying on HTML meta cache directives.

After that source repair has exact-head CI/review, the fresh exact-artifact Windows checkpoint required by #369/#87 remains the real acceptance boundary; this finding does not replace it.

djibian (commented):
DRAFT_FINDING abbacf0167c39cfaa41f604810038a87ed0b63f1

The startup/cache direction is materially stronger than the prior candidate: the earlier broad root-file sweep is gone, the Runtime root closure is explicit, child `src`/`href` URLs are content-addressed from packaged bytes, and the packaged Robot Window itself is renamed to a digest-derived directory/HTML identity with the world rewritten to select that exact name. That directly addresses the top-level localhost URL reuse left open by child-only query fingerprinting.

One integration regression remains on this exact head. Current `main@77e4783cdfc91c23726bba1866b0799e2b262f97` already integrated #374's lightweight static `test_packager_tracks_direct_html_resource_graph`, which derives direct resource names from `blockly_v2.html` and requires the packager to track them. Because this branch is still based on pre-#374 `13836b2...`, its version of `tools/ci/test_windows_release_contract.py` replaces that integrated test with an older hard-coded runtime-root assertion. The Windows archive/runtime oracle here is stronger, but dropping an already-integrated static omission guard while synchronizing would silently discard useful main work.

Smallest repair boundary: synchronize this Draft with current main while preserving the new content-addressed top-level Robot Window identity and retain #374's HTML-derived static resource-graph check (or an equivalent dynamic assertion) alongside the cache-isolation assertions. Then obtain fresh exact-head evidence/review. This is not a GO and does not itself request human testing.

djibian (commented):
GO 4e0b2da908986e0152f210a4db43305e64bdb1ee

Independent exact-head review finds the prior applicable findings resolved on current `main@77e4783cdfc91c23726bba1866b0799e2b262f97`.

The release root remains bounded to the explicit Runtime files, so the earlier broad `Get-ChildItem` sweep cannot package build-only helpers. The #374 HTML-derived omission guard is retained equivalently inside the current cache-isolation contract by deriving direct `src`/`href` names from source HTML and requiring each direct Runtime root reference in the packager.

For the remaining #369 startup boundary, direct local HTML dependencies are verified and fingerprinted from their packaged bytes; the top-level Robot Window identity is then derived from the complete packaged `plugins/robot_windows` byte closure, the plugin directory and HTML basename are renamed together to Webots R2025a's native `<window>/<window>.html` convention, and only the packaged world is rewritten to select that identity. The archive oracle independently normalizes the renamed tree, recomputes the digest, rejects a surviving stable `blockly_v2` directory, and proves the packaged world selects the exact content-addressed identity. This addresses both stale direct resources and nested relative imports without changing the source world, controller, Chrome file backend, Firefox broker, or offline authority boundary.

The older `NO_GO acef710e...` and `DRAFT_FINDING abbacf01...` therefore do not apply to this HEAD. Human Windows acceptance remains separate evidence required after integration; this GO is machine/source review only.

djibian (commented):
GO 4e0b2da908986e0152f210a4db43305e64bdb1ee

Independent exact-head source review against `main@77e4783cdfc91c23726bba1866b0799e2b262f97`. The bounded Runtime closure, byte-fingerprinted direct resources, content-addressed top-level Robot Window directory/HTML identity, packaged-world rewrite, independent release digest oracle, and preserved #374 HTML-derived omission guard resolve the applicable earlier source findings. Pinned Webots R2025a resolves `window()` as `<window>/<window>.html`, so this changes the top-level localhost path before stale HTML can be reused.

No source-level blocker found. This GO does not replace the fresh exact-artifact Windows Firefox checkpoint required by #369/#87 after canonical Ready CI.

djibian:
MERGE_OUTCOME_UNCERTAIN 4e0b2da908986e0152f210a4db43305e64bdb1ee

A conditional squash merge request was sent with `expected_head_sha=4e0b2da908986e0152f210a4db43305e64bdb1ee` after exact Ready CI #1425 succeeded and applicable exact-head GO evidence was present, but the GitHub transport returned HTTP 502. Immediate reconstruction still observed PR #373 open/unmerged at this exact HEAD and `main@77e4783cdfc91c23726bba1866b0799e2b262f97` unchanged.

That observation does not prove the merge request failed. Do not blindly retry from this uncertainty marker. A later execution must reconstruct current PR/main state first; if the PR is still open and main still excludes this candidate after the outcome is no longer ambiguous, it may decide eligibility afresh under the current contract.

PR #374, opened 2026-09-13
Package every Robot Window HTML resource on Windows

djibian:
Refs #369.

This is the smallest complete repair for the independently proven Windows release-completeness defect; it does not claim to resolve the separate startup/cache-consistency portion of #369.

- include the two direct `blockly_v2.html` dependencies that the current Windows archive omits: `led_observability.js` and `physical_preflight_runtime.js`;
- make the release builder treat the packaged `blockly_v2.html` as authority for its initial local `src`/`href` graph;
- strip query/fragment cache revisions, resolve each relative target against the packaged Robot Window directory, reject path escape, and fail closed if any referenced file is absent before the manifest/ZIP is produced;
- add a lightweight repository contract that derives the direct Robot Window resource set from the HTML itself and proves the packager tracks it, so this exact omission class fails before the Windows release build.

The runtime check validates the actual directory that becomes the archive instead of relying only on a duplicated hand-maintained required-file assumption. Existing Chrome/Firefox file-management semantics, WWI broker behavior, controller/runtime code, and physical-flight authority are unchanged.

Exact implementation head: `e85d5b90ab6092822b908dc817132ba81dc00f18`.


djibian (commented):
GO 9f94717ef88d2525c054e8745f9239cfe68bd4f2

Independent exact-head review against current `main@13836b2df76135dc8fafaaced81de7b513aa8161`: this candidate is correctly bounded to the deterministic release-completeness defect and does not overclaim the separate #338/#369 startup-cache repair.

The two HTML-required root modules that were absent from the Windows classroom ZIP (`led_observability.js` and `physical_preflight_runtime.js`) are added to the existing explicit runtime list without broadening the archive to build-time helpers. The new check runs against the assembled package directory before manifest/ZIP creation, strips only query/fragment cache revisions, resolves relative references from the packaged Robot Window root, rejects release-root escape and fails closed when a referenced file is absent. Existing `../blockly/...` dependencies remain inside the package root and are already copied before this closure check.

The change does not alter Chrome File System Access semantics, Firefox Qt/WWI broker semantics, Runtime execution, physical authority or the human-checkpoint mechanism. This GO establishes only the packaging-closure slice. #369 remains open for deterministic normal startup/cache isolation, and #87 still requires a fresh exact-artifact Windows Firefox checkpoint after the full repair.

Integration still requires successful exact-head canonical `CI Gate` for this Ready HEAD.

djibian (commented):
GO e85d5b90ab6092822b908dc817132ba81dc00f18

Independent exact-head review against current `main@13836b2df76135dc8fafaaced81de7b513aa8161`: the predecessor packaging-closure claim remains valid, and the new commit only strengthens its machine contract.

The candidate keeps the bounded explicit runtime-root file set, including the two HTML-required modules that were absent from the failed Windows archive, so it does not admit unrelated build-time siblings. The assembled-release check still derives local `src`/`href` targets from the packaged HTML, strips query/fragment revisions, rejects package-root escape and fails closed on missing files before manifest/ZIP creation. The added repository test now independently derives the direct root resource names from `blockly_v2.html` and requires the packager to track them, specifically guarding the omission class that produced the two 404s without changing runtime semantics.

No Chrome File System Access, Firefox Qt/WWI broker, Runtime execution, physical authority or checkpoint semantics are changed. This GO is limited to the release-completeness slice; #369 still requires the separate deterministic startup/cache-consistency repair and the eventual exact-artifact Windows Firefox checkpoint.

Integration still requires successful exact-head canonical `CI Gate` for this Ready HEAD.

PR #377, opened 2026-09-13
Prepare trusted physical qualification launcher

djibian:
Refs #157

## Scope

Prepare the smallest machine-only bridge from the already integrated trusted physical host to a representative real-device qualification path, without opening a second human checkpoint and without adding a student/browser authority surface.

- start `serve_physical_host.py` with distinct caller, browser-bootstrap and teacher channels;
- pass only the host-generated non-authority capability/preflight bootstrap to an actual Runtime v2 Robot Window;
- keep the post-reset teacher decision on a separate one-shot trusted host-side socket;
- submit only the exact current profile/AST/epoch from browser preflight and keep all later ordinary in-flight execution parameter-free;
- refresh the existing browser preflight fail-closed after the host deliberately rotates the connection epoch, accepting only the exact unchanged profile/canonical AST;
- use a temporary Runtime v2 Robot Window/world overlay so normal simulation UI semantics remain unchanged outside qualification mode;
- add deterministic no-hardware contract coverage through the existing Runtime CI harness.

## Safety boundary

This PR performs no motorized CI test and creates no `TEST_REQUIRED`. The existing #376 human checkpoint remains the sole open human slot. Importing the launcher performs no I/O or physical effect. Real effects still require the existing reset, fresh post-reset preflight, exact host-first teacher authorization, watchdog, SafeLink, acknowledgement and completion chain.

## Current evidence

The branch is based directly on current `main@4ec27ee2bac0092f27ba50a554e072c51552facc`. Draft CI is exploratory; integration evidence must be re-established for the exact Ready head under the current controller contract.


djibian:
This Draft is now dominated by #378 for the same #157 qualification-launcher slice. Its unique useful idea — refreshing exact current-program evidence after the trusted post-reset epoch rotation — is preserved as the explicit repair boundary in the authoritative `NO_GO c6b6d082dd2f362952696d161750cb4c10e8baf3` on #378. The implementation here is not a viable fallback: current exact head `f95e4004679a9b3048e8b93706c65f57bfc88d91` has a failing newest canonical CI run (#1430), where the refresh path clears the configured adapter on the existing AST-mismatch regression. No integration evidence is claimed. Closing this duplicate reduces engaged work; the branch is left intact, so no unique Git history is destroyed.

PR #380, opened 2026-09-13
Prime historical Boost support before human Webots evidence

djibian:
Refs #369.

The exact Windows repair on `main@e20d943f2ed9a83da6a71e794cb5133641817aac` is machine-ready for a fresh human checkpoint, but direct `human-checkpoint.yml` full Webots evidence currently fails deterministically when the historical Boost 1.74 cache is absent: the historical jobs invoke the fail-closed local support action before any cache-preprovisioning step exists in that workflow.

This bounded infrastructure repair adds one checkpoint-only preprovisioning job before the reusable Webots evidence. It restores the exact pinned cache key, downloads the exact archive only on a miss through the existing verified helper, verifies its pinned digest/provenance, saves it, and makes Full Webots evidence depend on that preparation. The normal canonical CI path and historical builder remain unchanged. A contract regression test proves the checkpoint path cannot regress to fail-on-cache-miss before preprovisioning.

This does not publish or claim a human PASS and does not change the checkpoint authority, single-slot rule, notification boundary, product Runtime, or Windows repair semantics. After integration, the existing exact Windows checkpoint request can be retriggered through the trusted mechanism.


djibian (commented):
GO 8b2007f4ef8f2af996038cd8b19e56143aace25c

Independent falsification review of the Ready candidate against the current V4 human-boundary and CI-evidence obligations.

The change is narrowly checkpoint-infrastructure-only. It does not alter request authority, supported profiles, single-slot/deduplication, notification semantics, target-SHA binding, Runtime/product behavior, or the canonical CI decision workflow. The new job depends on validated request output, restores one exact cache key, acquires only on an exact miss through the existing fixed-URL helper, verifies the already-pinned archive size/SHA-256 through `prepare_historical_boost.verify_archive`, and saves only that exact verified archive. Full Webots evidence is then causally gated on successful preparation while still executing against `needs.validate.outputs.target_sha`.

The added regression tests cover the new workflow topology, exact cache path/key, verified acquisition path, absence of fail-on-cache-miss in the priming job, and YAML job/heredoc contract. Network availability remains support preparation rather than an acceptance oracle: inability to obtain/verify the archive fails closed and cannot publish TEST_REQUIRED.

The newest Ready `CI Gate` run is currently red only because `Runtime suite / student-ui` timed out waiting 5 s for the real repeat tooltip to become visible; all patch-relevant workflow/repository contracts and the Windows release path in that run passed. This GO is code-review authority for the exact HEAD only; integration still requires a newer completed successful canonical Ready `CI Gate`.

PR #382, opened 2026-09-13
Allow exact Boost acquisition only inside human checkpoints

djibian:
Refs #369.

The first post-#380 human-checkpoint run (`34778801418`) exposed a narrower failure than the original cache-miss diagnosis: the new priming job acquires and verifies the pinned Boost archive successfully, but the issue-comment workflow receives a read-only cache token. `actions/cache/save@v4` therefore warns `cache write denied: token has no writable scopes`, the prep job still reports success, and the historical Webots jobs then fail their strict target-local cache restore.

This repair does **not** widen workflow permissions and does not weaken normal CI. Instead, the existing historical Boost composite action remains strict cache-only for every non-checkpoint context, while an `issue_comment` human checkpoint may, on an exact cache miss only, invoke the existing fixed-URL acquisition helper from the exact target checkout. That helper verifies the already-pinned size/SHA-256 before atomically exposing the archive, and the composite action verifies it again before any historical sidecar build.

The contract test is updated to prove:
- ordinary PR/maintenance contexts retain fail-on-cache-miss behavior;
- network acquisition is gated specifically to `github.event_name == 'issue_comment'` and an exact cache miss;
- no cache save, direct URL, curl/wget/apt fallback, or unverified archive path is introduced in the action;
- all affected historical Webots jobs still consume the same pinned support action and offline builder.

No product/runtime code changes. No human PASS is claimed. After integration, any new checkpoint must bind a fresh exact main SHA so its target checkout contains this support behavior.


djibian (commented):
GO 8574a3c5a7533b8600009f37af58ae162801ab73

Independent falsification review on this exact Ready head. The bounded checkpoint-only fallback preserves the current trust and CI boundaries.

- Ordinary pull-request, schedule and workflow-dispatch historical consumers remain strict restore-only/fail-closed because cache misses still fail whenever `github.event_name != 'issue_comment'`.
- The only current repository `issue_comment` workflow that can reach the reusable Webots evidence is the trusted human-checkpoint path, whose upstream validation is owner-only and requires an exact supported `CHECKPOINT_REQUEST`; an untrusted comment cannot produce a target output or reach the called evidence jobs.
- On the checkpoint path, acquisition occurs only after an exact cache miss and uses the existing target-checkout helper with fixed archive identity, pinned size/SHA-256 verification and atomic exposure; the composite action verifies the archive again before the historical builder consumes it. No cache write privilege, apt/curl/wget fallback or unverified support path is added.
- Normal canonical Ready CI remains hermetic and the exact candidate passed canonical `CI Gate` run `34779466405`; all historical jobs restored the already-provisioned exact cache successfully there.
- Because the fallback implementation lives in the target checkout, the PR correctly does not claim to repair the already-requested older target in place: the next checkpoint must bind a fresh exact main containing this change. That is an explicit provenance boundary rather than an implicit promotion of old evidence.

No unresolved source blocker found on this exact candidate. Integration still requires all other V4 merge preconditions to remain true immediately before the conditional squash merge.

PR #384, opened 2026-09-13
Pin physical qualification runtime closure

djibian:
Refs #157.

This candidate isolates the execution-qualification Python runtime from the existing read-only probe bundle instead of widening that artifact's claim. It pins an exact seven-wheel Ubuntu 22.04 / CPython 3.10 x86-64 closure for the already-pinned cflib source, including SciPy, packaging and PyYAML in addition to the proven probe dependencies, and adds an effect-free environment runner that rejects ambient packages, installs only the locked wheelhouse, imports the cflib paths used by the trusted host, and exercises both qualification launcher and host entry points under `python3 -S` / `PYTHONNOUSERSITE=1` using `--help` only.

No human checkpoint profile is enabled, no artifact is published, and no hardware or motorized action is authorized. The next step is to wire this isolated closure proof into deterministic machine CI before considering qualification-artifact packaging.


djibian (commented):
NO_GO e2aea983437c5c1c7f14dc86c859a46b8b968120

The runtime lock/isolated runner direction is useful, but the exact candidate still makes package-network availability part of canonical pull-request acceptance. `.github/workflows/ci.yml` calls `python3 -m pip download ... "$spec"` directly inside the decision workflow and only afterwards verifies the expected wheel hashes. Under the current V4 contract, network availability is explicitly not an acceptance oracle: a transient PyPI/network failure would make an otherwise identical candidate red, and successful network resolution is supplying acceptance inputs during the decision run rather than consuming previously prepared exact support.

Smallest repair boundary: separate acquisition from acceptance. Populate the exact seven verified wheels only in a non-decision support path/cache, then make canonical Ready CI restore that exact keyed support fail-closed, verify filenames/hashes, and run the isolated `python3 -S` proof without any network fallback. Preserve the current exact lock and no-hardware runner semantics. The earlier accidental historical-Boost cache-key edit is no longer present on this exact head.

djibian:
Superseded by #385. The useful #384 boundary (exact seven-wheel/cflib runtime lock plus isolated effect-free launcher/host proof) is preserved there, while #385 also addresses this PR's authoritative network-as-acceptance NO_GO by separating maintenance acquisition from Ready decision evidence and structurally scoping the exact support path. Closing this dominated duplicate; its branch is retained, so no unique work is destroyed.

PR #389, opened 2026-09-14
Isolate Windows Robot Window startup sessions

djibian:
Advances #369 after the owner-authoritative #383 Firefox relaunch and school-PC launcher failures.

The failing relaunch path had two coupled properties: the top-level content-addressed Robot Window URL remained stable across launches while its child assets were rewritten to a launch-local ephemeral loopback origin, and that loopback server ran inside `Start-Job`. A stale Firefox top-level document could therefore retain a dead prior child-asset origin; separately, #383 narrowed the school-PC AMSI failure to the `Start-Job` child-script boundary.

This repair keeps the HTTP bridge loopback-only, exact-byte/MIME checked, no-cache and `Connection: close`, but removes `Start-Job`/`Receive-Job` entirely. Each launch now creates a unique temporary Robot Window identity and same-directory temporary world, rewrites only that session copy to its own ephemeral loopback port, and services requests from the launcher process while Webots is alive. A lingering earlier launcher can therefore keep its own session and port without colliding with or poisoning a later launch.

`-ValidateOnly` now constructs two simultaneous complete launch/server lifetimes and proves distinct Robot Window identities and ports, correct per-session world selection, correct per-session rewritten origins, and exact HTTP bytes/MIME for both. The static Windows release contract rejects background-job reintroduction and preserves the Chrome-safe module import guard merged in #386.

This is machine-side repair evidence, not a human PASS. A fresh exact-artifact Windows checkpoint remains necessary only after the candidate is independently reviewable and canonical CI has passed.

PR #386, opened 2026-09-14
Resolve Chrome RobotWindow import from loaded script URL

djibian:
Refs #369 #383.

The owner-authoritative #383 FAIL on exact current `main@9e706ae605d5d95f9744352f3d2289b19c5f8954` demonstrated a persistent Chrome startup failure on the diagnostic Windows path: `main.js` is loaded as a cross-origin classic script and later executes `await import('./webots/RobotWindow.js')`; Chrome therefore gives that dynamic import an `about:blank` base and rejects the relative module specifier.

This bounded repair captures `document.currentScript.src` synchronously while `main.js` itself is executing, then resolves the already-pinned local `./webots/RobotWindow.js` module with `new URL(..., WEBEEBLOCKS_MAIN_SCRIPT_URL).href` before the deferred `window.onload` import. On the Windows classroom path introduced by #379, `main.js` is served by the loopback asset server with CORS enabled, so the module and its relative `request_methods.js` dependency remain local/offline while no longer relying on Chrome to invent a base URL for a cross-origin classic script. Normal same-origin Robot Window loading keeps resolving against the exact loaded `main.js` URL as well.

The Windows release contract now requires this explicit loaded-script base and rejects reintroduction of the relative dynamic-import literal. The pinned Webots bridge provenance, Firefox native file broker, Chrome File System Access path, runtime authority and physical-effect boundaries are otherwise unchanged.

This PR intentionally does **not** claim to repair the two other #383 defects: the reference-machine AMSI rejection of the PowerShell launcher or the nondeterministic Firefox/Robot Window startup/readiness behavior. #369 remains open for those trunk-health repairs, and no human PASS/checkpoint is claimed here.

Exact Draft candidate: `42b25a6ba536feb87ef1eee8c98ff72c739b5dee`, based directly on `main@9e706ae605d5d95f9744352f3d2289b19c5f8954`. This execution changed the candidate and therefore does not provide independent review.

PR #390, opened 2026-09-14
Replace PowerShell classroom launch path with native bootstrap

djibian:
Advances #369 from the owner-authoritative #383 FAIL while preserving the already-working project-file semantics and the Chrome module-resolution fix now on `main`.

The reference Windows 11 acceptance machine rejects the current PowerShell `Start-Job` launch path through AMSI before the browser can start. This candidate moves the normal classroom entry path to a small native Win32 launcher built with the pinned Webots R2025a MinGW toolchain. `Launch-WebeeBlocks.cmd` starts that executable directly; the release archive no longer ships the PowerShell launcher as an executable classroom dependency, and the build rejects dynamic MinGW runtime imports.

The native launcher owns the loopback exact-byte/MIME asset server with `Connection: close`, prepares a unique Robot Window + world identity for every process while keeping canonical packaged files immutable, and routes both static startup dependencies and runtime-generated Blockly media through the packaged local proxy.

Earlier Ready Windows release runs on `1bf4de8a082fd33f60f40c6778043f2ef7c19156` and `d384957a507fb0a3c90fb5c876ff5df799c443da` exposed a deterministic native `g++.exe` failure after the controller build; the branch now compiles the launcher inside the same pinned Webots MinGW/MSYS environment. The authoritative NO_GO on `719c16431b57f615f50ae1cc737f456b96425f6b` then found that an accepted idle client could block the single-threaded server indefinitely. `85f3bf5418e66de17737bbf8e9f9f48ed8a7c675` bounds accepted-client receive/send I/O to 500 ms, contains timeout/aborted-client failures locally, and adds a `--validate-only` idle-client oracle before proving the same listener still serves exact packaged CSS and Blockly sprite bytes/MIME. Ready CI on `719c16431b57f615f50ae1cc737f456b96425f6b` also exposed two C++ raw-string literals whose default delimiter collided with regex content; exact head `5f6346c7052fbd0f8342346170c1836ce39a21c9` uses explicit raw-string delimiters so the pinned MinGW compiler can parse those regexes without changing their semantics.

This branch is reconciled with `main@8ccec834149c3d666c02935d7dbd03ee10a1fc43`, including #386. #389 remains a separate bounded PowerShell startup-session repair path; this PR does not erase its distinct evidence. No human PASS is claimed. A fresh exact-artifact offline Windows checkpoint remains necessary only after a machine candidate is stable and eligible under the repository checkpoint contract.


djibian:
DRAFT_FINDING 3c3b9296db75b52d41bf13ab1307e2d6eaa853a9

The per-process Robot Window directory currently contains only the rewritten session HTML, but `main.js` still creates Blockly with `media: 'vendor/media/'` and `trashcan: true`. Blockly resolves that configured media path from the page and its trashcan implementation builds the sprite URL from `workspace.options.pathToMedia + Blockly.SPRITE.url`.

For this candidate the page lives under `plugins/robot_windows/webeeblocks_session_<pid>_<tick>/`, while that session directory has no `vendor/media` subtree. The HTML `src`/`href` rewrite does not cover these runtime-generated Blockly media URLs. A normal session can therefore request the trashcan sprite from the new session path through the Webots top-level HTTP server, where the file is absent, instead of through the native exact-byte proxy. This both regresses the classroom UI asset closure and leaves a dynamically generated child request on the R2025a server path that the repair is intended to avoid.

Smallest repair boundary: make the Blockly media base resolve to the native proxy/canonical packaged `vendor/media/` location (or otherwise materialize the exact required media under the session identity), and add a machine oracle proving the runtime-generated sprite URL resolves to the exact packaged bytes without traversing the Webots child-asset path. Preserve the unique session top-level identity, canonical packaged-file immutability, loopback-only `Connection: close` serving, and the #386 absolute RobotWindow module resolution.

djibian (commented):
NO_GO 3c3b9296db75b52d41bf13ab1307e2d6eaa853a9

Independent falsification review of the exact Ready candidate against the #369 startup boundary.

The unique session directory contains only `<session>.html`, while the unchanged Runtime still injects Blockly with `media: 'vendor/media/'` and `trashcan: true`. Blockly treats that as the page-relative media path; its trashcan uses `workspace.options.pathToMedia + Blockly.SPRITE.url`. Because the page now lives under `plugins/robot_windows/webeeblocks_session_<pid>_<tick>/`, the runtime-generated sprite request resolves under that session directory, which has no `vendor/media` subtree. That URL is not one of the static `src`/`href` attributes rewritten by `rewrite_html`, so it bypasses the native exact-byte proxy and goes back through the Webots Robot Window origin to a missing session asset.

This is inside the claimed classroom startup/asset-closure result: the candidate can regress the Blockly UI media and still leaves a dynamically generated child request on the R2025a server path being removed from the critical closure.

Smallest repair boundary: make the configured Blockly media base resolve to the native loopback/canonical packaged `vendor/media/` location (or materialize the exact media closure beneath the session identity), and add a machine oracle that exercises the runtime-generated sprite URL and verifies exact packaged bytes/MIME through the intended origin. Preserve the native non-PowerShell entry point, unique per-process top-level identity, canonical packaged-file immutability, loopback-only `Connection: close` serving, and #386 module-resolution behavior.

djibian (commented):
NO_GO 719c16431b57f615f50ae1cc737f456b96425f6b

Independent falsification review of the exact Ready candidate against current V4 and the #369 deterministic-startup boundary.

The earlier `3c3b9296db75b52d41bf13ab1307e2d6eaa853a9` Blockly-media finding is resolved in this HEAD: Runtime derives the Blockly media URL from the loaded proxied `main.js` origin, and the native validation path checks the exact packaged sprite through that proxy. The prior silent native compiler failure also now has a concrete causal repair: the launcher compile runs with the pinned Webots MinGW/MSYS2 directories restored on `PATH`, matching the controller toolchain environment. Exact Ready CI is still running, so no machine PASS is inferred from that source change alone.

A separate deterministic-startup failure remains in the single-threaded native HTTP server. `Server::serve_one()` waits only for the listening socket, then accepts one client and calls `serve_client()`. That function performs blocking `recv()` calls until a complete HTTP header arrives, but the accepted client socket has no receive timeout or nonblocking/readability deadline. A loopback client that connects and then sends no complete request can therefore block the launcher indefinitely inside `recv()`. Because the normal Webots loop calls `server.serve_one(50)` synchronously, one idle/speculative browser connection can prevent all later asset requests from being served and also prevent the launcher from observing Webots exit. The current aborted-client oracle covers a peer that closes immediately; it does not cover an accepted peer that stays connected without completing headers.

This is directly inside the repeated normal-startup reliability claim. Smallest repair boundary: bound the lifetime of each accepted request (for example with a receive timeout/nonblocking deadline that is contained as a per-client failure), then add a Windows validation oracle that connects an idle client, proves the server returns to accept another request within a bounded interval, and then successfully serves an exact asset. Preserve loopback-only serving, exact bytes/MIME, `Connection: close`, unique session identity, canonical-file immutability, the native non-PowerShell entry point, and the already-fixed dynamic Blockly media path.

djibian:
CI_DIAGNOSIS 719c16431b57f615f50ae1cc737f456b96425f6b

Ready CI Gate run `34832875824` reached the native launcher compiler successfully, so the prior silent `g++.exe` failure is causally narrowed further. `runtime-v2-windows-release` now emits concrete C++ diagnostics before failing.

The immediate compile failure is in `packaging/windows/WebeeBlocksLauncher.cpp`, not the MinGW environment. Two default-delimiter raw string literals contain their own terminating `)"` sequence:

- line 311: `R"((src|href)=\"([^\"]+)\")"`
- line 350: the Robot Window regex includes `...{16})\"...` inside another `R"(...)"` literal.

GCC therefore terminates each raw string early and reports `missing terminating " character`, followed by the cascading stray-backslash/parser errors shown in the exact Windows job log. This is deterministic source syntax, independent of the separate current-head NO_GO about an idle accepted client blocking `recv()`.

Smallest source repair: use a custom raw-string delimiter that cannot occur in each regex (or an ordinary escaped string), without changing the regex semantics. A later Ready candidate still needs the existing idle-client lifetime repair plus fresh exact-head Windows release CI; rerunning `719c164...` unchanged would be blind.

djibian (commented):
GO 5f6346c7052fbd0f8342346170c1836ce39a21c9

Independent falsification review of the exact Ready candidate against current V4 and the #369 Windows startup/distribution boundary.

The applicable earlier refutations are resolved in this HEAD. The `3c3b9296db75b52d41bf13ab1307e2d6eaa853a9` runtime-generated Blockly media escape is closed by deriving the media root from the loaded proxied `main.js` URL and validating the packaged sprite through the native exact-byte server. The `719c16431b57f615f50ae1cc737f456b96425f6b` idle-client deadlock is closed by bounded accepted-client receive/send I/O plus an idle-client containment oracle followed by exact asset serving on the same listener. The Ready CI compiler failure on that older HEAD is also causally addressed: the two default-delimiter C++ raw strings that collided with regex content now use explicit `WB` delimiters without changing the intended regex semantics.

Review of the exact current launcher path found no further bounded refutation: the normal classroom CMD path invokes the native GUI bootstrap rather than PowerShell; the canonical packaged world/plugin bytes remain immutable while per-process session identities isolate Webots top-level caching; static startup dependencies, dynamic module loading and Blockly media resolve through the loopback-only packaged server with explicit MIME/CORS/no-cache/Connection-close behavior; path traversal is contained under the package root; accepted clients have finite I/O lifetime; and the existing Chrome project-file / Firefox native broker semantics are not replaced by this bootstrap.

This GO is source/review evidence only. Exact-head canonical CI remains an independent integration precondition, and no real Windows acceptance result is inferred here.

djibian:
SUPERSEDED_BY #389 / main@2229c5b9b4c5f162301357cb883639dd914e7ca5

The smaller PowerShell-path repair has now been integrated through #389. That integrated change addresses the same owner-authoritative #383 startup/relaunch boundary without introducing the additional native-launcher/toolchain surface: it removes the `Start-Job`/`Receive-Job`/`Stop-Job` AMSI boundary, gives every launch a unique Robot Window/world identity and ephemeral loopback listener, preserves the complete Robot Window tree for page-relative media, and contains per-client request failures. Exact integrated target `2229c5b9b4c5f162301357cb883639dd914e7ca5` already has a trusted `windows-low-end` checkpoint request on #369 for the indispensable real Windows/antivirus/relaunch evidence.

This PR therefore no longer has a demonstrated product need while that exact integrated repair awaits its real-world verdict. Its native bootstrap work remains preserved in this PR/branch as fallback evidence; it should be reconsidered only if the integrated #389 path is falsified by the checkpoint or a new concrete need appears. Continuing or rebasing this alternative now would add change surface rather than reduce the current uncertainty.

PR #385, opened 2026-09-13
Prove exact offline physical qualification runtime closure

djibian:
Advances #157 machine-only qualification preparation.

Pins the exact Ubuntu 22.04 / CPython 3.10 x86_64 seven-wheel runtime closure plus exact cflib commit/tree/subtree used by the production physical qualification launcher. Canonical CI separates package-network acquisition from decision evidence: same-repository Draft/maintenance runs may prime an exact content-keyed cache, while Ready candidates only restore and verify that exact support and fail closed if it is absent.

The verifier rejects missing, extra or hash-mismatched wheels, installs with `--no-index --no-deps` under `python3 -S` + `PYTHONNOUSERSITE=1`, poisons ambient package roots, pins dependency and cflib import origins, and exercises the real qualification launcher and trusted host entrypoint imports through effect-free `--help` before any radio/session authority can be created.

The workflow selector is structural and fail-closed: a bounded qualification-support block in `.github/workflows/ci.yml` is compared across base/head, while the qualification lock/oracle and `tools/physical/*.py` paths independently select the proof. This keeps unrelated Ready candidates independent of a PR-local cache while ensuring provenance, acquisition, cache and verification edits cannot bypass qualification evidence.

Exact branch head `339bee98479b3469de149f0bd7fd42c273bb1a51` is reconciled with current `main@2229c5b9b4c5f162301357cb883639dd914e7ca5` by a non-conflicting merge: the intervening #386/#389 changes touch only Windows launcher/Robot Window files, disjoint from this PR's five qualification-runtime files. No hardware/radio/motorized effect, no new student capability, and no checkpoint profile are introduced. The current #369 Windows checkpoint is independent and this PR neither opens nor consumes a human-test slot. Representative real-device qualification remains future human evidence.

Issue #393, opened 2026-09-14
Package deterministic offline physical qualification artifact


djibian:
## Goal

Advance #157 after integrated #378 and #385 by packaging the already-proven trusted physical qualification launcher, prepared Runtime v2 Robot Window, exact cflib source and exact seven-wheel Ubuntu 22.04 / CPython 3.10 x86_64 runtime closure into one deterministic **machine-prepared offline artifact** suitable for a later bounded real-device qualification request.

This issue is preparation only. It must not perform a radio connection, physical effect, motorized run, or enable a new human checkpoint profile.

## Current proven prerequisites

- #378 integrates the trusted real-Robot-Window qualification launcher with distinct caller/browser/teacher channels, browser `executionAuthority:false`, exact current-program binding across trusted reset, one-shot teacher decision and parameter-free post-approval execution.
- #385 integrates the exact qualification runtime closure: exact cflib commit/tree/subtree plus seven locked wheel filenames/SHA-256s, isolated `python3 -S` / `PYTHONNOUSERSITE=1` proof, hostile ambient roots and effect-free real launcher/host `--help` imports. Ready acceptance restores and verifies this support without package-network fallback.
- #157 still requires representative real-device qualification; deterministic packaging is the remaining machine-side prerequisite before any future applicable human execution profile can be designed.

## Required boundary

- Build from an exact repository SHA and preserve repository-relative layout required by `tools/physical/launch_physical_qualification.py`.
- Package the prepared Runtime v2 Robot Window closure, `worlds/crazyflie_runtime_v2.wbt`, trusted physical launcher/host source needed by that launcher, exact pinned cflib source, exact seven locked wheels and the qualification lock/provenance needed to verify them.
- Prepare Blockly/Runtime assets deterministically before packaging; missing generated/vendor assets must fail closed rather than produce a partial artifact.
- Emit a deterministic manifest covering every packaged regular file with relative path, size and SHA-256 plus exact source SHA and qualification-runtime provenance. Reject missing, extra or hash-mismatched files during verification.
- Provide an offline environment/entry helper that uses only the bundled exact wheelhouse/cflib closure (`--no-index --no-deps`, isolated from user/site packages) and launches the bundled trusted qualification entrypoint. Machine acceptance may exercise only effect-free help/validation paths; it must never connect to `radio://` hardware.
- Keep browser preparation non-authority, teacher authorization separate, and ordinary execution parameter-free. Packaging must not introduce an alternate physical-effect API.
[...]

PR #394, opened 2026-09-14
Package exact physical qualification bundle

djibian:
Advances #157 after the integrated exact qualification-runtime closure in #385.

This Draft starts the deterministic packaging boundary for a future representative physical qualification artifact. Exact branch head `00f3e6c10cee2a493ea51596c75f8f0e847efde9` adds a machine-only packager that binds the requested WebeeBlocks source SHA, verifies the pinned cflib + seven-wheel runtime support, writes provenance, and emits a normalized deterministic tar.gz plus an offline runner. The packager itself creates no Crazyradio session, teacher authority, Webots process, or physical effect.

This is not yet a complete candidate. The current source archive alone does not contain generated Runtime v2 `vendor/` browser assets, and the canonical `crazyflie_runtime_v2.wbt` still contains pinned remote EXTERNPROTO references; both must be made explicitly offline/self-contained (without treating transient package-network availability as a Ready acceptance oracle) before this PR can be offered for validation. A deterministic regression oracle for the final bundle is also still required.

No human checkpoint profile or notification mechanism is changed. The existing Windows TEST_REQUIRED #392 remains the sole open human checkpoint, and no real-device acceptance is claimed here.


djibian:
DRAFT_FINDING 6d6b57c48444911e710bdad192a90a4e146cceff

Two additional closure points should stay explicit while this Draft is completed:

1. `_copy_verified_support()` currently copies the entire pinned cflib checkout with `shutil.copytree(...)`. The final tar normalizes metadata, but the copied `.git` directory contains checkout-local/index/config state that is not constrained by the verified commit/tree/subtree and can vary between equivalent checkouts. Because the stated result is a deterministic exact bundle, either the bundled cflib representation must itself be content/provenance bounded or the archive oracle must demonstrate reproducibility across independently prepared equivalent support trees; normalizing tar mtimes alone does not establish this.
2. `git archive` contains the C/C++ Runtime v2 controller sources but no built `crazyflie_runtime_v2` executable. `launch_physical_qualification.py` starts Webots directly and does not build that controller. A runnable qualification artifact therefore also needs an exact machine-prepared controller binary (or an explicitly deterministic offline build step before launch) in addition to the already-recorded generated Blockly `vendor/` and local Webots PROTO/world closure.

These are Draft findings, not an exact-candidate verdict. Preserve the current non-authority preparation boundary and do not turn package/network reachability into Ready acceptance evidence.

djibian:
DRAFT_FINDING 6d6b57c48444911e710bdad192a90a4e146cceff

The current normalized tar path still embeds checkout-local Git administration bytes for cflib. `_copy_verified_support()` first proves the exact cflib commit/tree/subtree, but then `shutil.copytree(cflib_root, bundled_cflib, symlinks=True)` copies the whole cached checkout, including `.git/`. `_write_archive()` normalizes tar ownership/mode/mtime, but it cannot normalize `.git/index`, reflogs/config/refs and other checkout metadata whose bytes may differ between two valid checkouts of the same proven commit. The final tar digest can therefore depend on how the support checkout was materialized, rather than only on the exact cflib tree + locked wheels + WebeeBlocks SHA.

That also makes the extracted runner's `verify_qualification_runtime.py` depend on shipping this mutable Git metadata merely to re-prove provenance already established at bundle construction.

Smallest repair boundary: after build-time `verify_cflib_provenance`, package a canonical content representation of the pinned cflib source (for example an exact `git archive`/subtree or an explicitly selected source tree without `.git`), bind those copied bytes plus the already-verified commit/tree/subtree in the bundle manifest, and make bundle-local verification consume that content/provenance without requiring checkout administration state. Add a regression oracle that would reject `.git/` inside the final artifact and preserves byte-identical output for identical canonical inputs. This is separate from the already-recorded generated Runtime-v2 vendor/EXTERNPROTO incompleteness.

djibian:
DRAFT_FINDING b3932420e8463f3e6f63f79599ff402f9e0ee4ee

The top-level bundle manifest currently records `"executionAuthority": false`, and canonical CI explicitly treats any other value as “bundle crossed authority boundary”. That claim is too broad for the artifact being built. The bundle contains `serve_physical_host.py` / `launch_physical_qualification.py`; after the trusted host establishes a live session and the teacher explicitly authorizes the exact run, that host is precisely the component that can create and consume physical execution authority. What remains non-authority is the browser/preparation bearer before that trusted decision, not the whole runnable bundle.

Smallest repair boundary: make the manifest describe the actual trust boundary rather than asserting that the complete artifact has no execution authority — e.g. bind that preparation itself performs no effect and that execution requires the trusted host + explicit teacher authorization, while preserving the existing browser `executionAuthority:false` invariant inside the runtime protocol. Update the CI oracle accordingly. Do not weaken the teacher authorization, live-session, preflight or effect gates.

djibian:
CI_DIAGNOSIS b3932420e8463f3e6f63f79599ff402f9e0ee4ee

Draft run `34842202338` failed deterministically in `Select suites` step `Verify selector and repository contracts`, not in qualification-runtime restoration or bundle construction. The bundle was packaged, verified and uploaded successfully; the contract suite then failed `tools/ci/test_workflow_contract.py::test_ready_gate_is_exact_head_and_draft_has_distinct_check_name` because the workflow now contains the exact-head expression `${{ github.event.pull_request.head.sha || github.sha }}` 7 times while the existing oracle requires exactly 4.

This is a trust-boundary regression in the workflow contract, not an external/network failure and not a reason to rerun the unchanged SHA. Any repair must preserve the existing exact-head Ready/Draft decision semantics rather than weakening/removing the oracle merely to accommodate extra artifact plumbing. The smallest direction is to avoid duplicating the authoritative candidate expression throughout new support steps (derive/reuse an already-authorized exact target once, or extend the contract only if the same exact-head invariant remains structurally established), then run fresh CI on the new HEAD.

djibian:
DRAFT_FINDING b3932420e8463f3e6f63f79599ff402f9e0ee4ee

The current bundle runner does not bind the Webots runtime that defines the qualification world/controller environment. `QUALIFICATION_MANIFEST.json` names Ubuntu/Python but no Webots version; `Run-Physical-Qualification.sh` validates only Python/Linux and then lets `launch_physical_qualification.py` use its default `webots` executable. This matters independently of the already-recorded missing controller/vendor/PROTO closure: once the world is made offline by resolving its R2025a dependencies through the installed Webots tree, running the same artifact under another Webots release can silently change those resources and runtime behavior while the current bundle checks still pass.

Smallest repair boundary: bind Webots `R2025a` as part of the artifact provenance, resolve one explicit executable (preferably from an explicit `WEBOTS_HOME`/override), fail closed before any session if its reported version is not R2025a, and pass that exact executable to `launch_physical_qualification.py --webots`. Add a machine oracle for the runner/manifest contract. This remains machine-preparation evidence only and does not request a hardware checkpoint.

djibian:
DRAFT_SUPERSEDED 77a5d5f0681980e1440b0903725ac7e71d984074

This Draft is now dominated by #395 for the same #393/#157 packaging boundary. #395 preserves and extends the useful intent here with the prepared Runtime v2 browser closure, localized `webots://` world, built Runtime v2 controller, exact cflib/seven-wheel support, complete file manifest/verifier, offline runner and direct artifact-directory upload required by #393. The unique normalized-tar direction in this branch is not an acceptance requirement and does not repair this Draft's recorded `.git` provenance, missing generated/runtime closure, over-broad authority claim or unbound Webots-runtime findings.

#395 is still Draft and currently has explicit exact-head findings; closing this duplicate does not validate it. This PR and its comments remain the durable reference for the earlier diagnostics, including the requirement that package proof ultimately live inside the existing V4-authorized CI boundary rather than weaken workflow governance.

PR #378, opened 2026-09-13
Compose trusted physical qualification launcher

djibian:
Refs #157.

This bounded product slice composes the already-integrated trusted physical host into a human-executable real-device qualification launcher without changing any existing physical authority or effect primitive.

- starts `tools/physical/serve_physical_host.py` with three distinct inherited channels: ordinary caller IPC, one-way browser bootstrap, and the trusted teacher decision socket;
- creates an ephemeral copy of the real `blockly_v2` Robot Window plus world binding, injects only the host-generated non-authority capability/preflight bootstrap, and invokes the existing `WebeeBlocksPhysicalPreflight.configure(...)` / current-program preflight path;
- never exposes the teacher channel or an execution endpoint to the Robot Window/student context;
- stages the exact browser-produced profile/AST/pre-reset epoch through the existing bounded `validate-run-context` request, then presents the existing host-first post-reset exact binding to the teacher for one explicit approve/deny decision;
- after approval, sends only parameter-free `execute-next-inflight` requests. Static sequences derive only their request count from the exact already-bound AST; dynamic shared-interpreter programs remain one parameter-free host execution request. No caller supplies direction, distance, color, branch, sensor value or other effect semantics;
- cleans the ephemeral Robot Window/world and terminates the composed processes fail-closed on completion/error;
- adds a deterministic fake-host/fake-Webots composition oracle proving exact binding preservation, distinct teacher authority, non-authority browser bootstrap, parameter-free execution and cleanup, wired into the existing Runtime core physical lifecycle test.

No motorized run is performed by this change. No `TEST_REQUIRED` is opened: #376 remains the only unresolved human checkpoint. This slice does not alter `AGENTS.md`, checkpoint/notification mechanisms, workflows, the trusted host's authority semantics, or any CRTP/effect implementation.

Exact Draft candidate: `c6b6d082dd2f362952696d161750cb4c10e8baf3`, based directly on `main@4ec27ee2bac0092f27ba50a554e072c51552facc`.

Ready promotion requires exact-candidate automated evidence; independent review must be supplied by another execution because this execution created the candidate.

PR #395, opened 2026-09-14
Package deterministic offline physical qualification artifact

djibian:
Advances #393 / #157 after integrated #378 and #385.

This Draft builds one deterministic machine-prepared offline qualification artifact around the already-integrated trusted physical launcher. It packages the prepared Blockly 13.2.1 Runtime v2 Robot Window, a localized `webots://` R2025a world, the built Linux Runtime v2 controller, the trusted physical host/launcher source closure, a canonical tracked cflib tree at commit `45fdb784c9d13074c42835f3b5ac1d12133bf873` (without checkout `.git` administration bytes), and exactly the seven locked CPython 3.10/Linux x86-64 wheels.

The package is bound to the exact clean WebeeBlocks source SHA, exact cflib tree/subtree, wheel hashes, Blockly version, Webots R2025a target and the exact controller build-image digest `sha256:f0023e30daf38b172e4e6ad24ed345909bcd9551df34d63d824e121a7cebf099`. Manifest verification rejects missing, extra or modified files and independently reconstructs the canonical cflib Git tree hash. Effect-free import verification and the packaged runner disable Python bytecode writes so verification does not mutate manifest-covered source bytes. The runner also fails closed unless it resolves an executable reporting Webots R2025a and passes that exact executable to the trusted launcher.

Packaging evidence is integrated into the existing canonical `CI Gate`; no fifth workflow is introduced. Exact runtime/prepared-input caches may be primed only on same-repository Draft or maintenance runs. Ready runs restore exact-key support only and fail closed if it is absent. The prepared artifact directory is uploaded directly as a GitHub Actions artifact, without a repository-created nested archive.

The trust boundary remains unchanged: packaging/preparation mints no execution authority and emits no physical effect; browser preparation remains non-authority; a real run still requires the existing live trusted host, post-reset/preflight gates and separate explicit teacher authorization before parameter-free execution. This PR performs or claims no radio session, motorized run, `CHECKPOINT_REQUEST`, `TEST_REQUIRED` or real-device PASS. Existing Windows checkpoint #392 is independent and unchanged.


djibian:
DRAFT_FINDING 5dae3f6f27d99d2a50456d7372fc9ddac72989e2

Current exact-head reconstruction exposes two deterministic CI refutations plus two artifact-closure gaps that must be resolved before this Draft can become a candidate:

1. Canonical `CI Gate` run `34847852600` still fails the existing V4 workflow oracle because this PR adds `.github/workflows/physical-qualification-package.yml`. Current `main` deliberately permits only the four established workflows, and `AGENTS.md` makes canonical `.github/workflows/ci.yml` the required integration evidence. An ordinary #393 product PR cannot self-authorize a fifth workflow. Smallest repair: place the package preparation/proof in the existing canonical workflow boundary (or another already-authorized workflow path) without weakening `test_only_v4_workflows_exist` or changing governance obligations.
2. Package run `34847852572` now gets through exact cache restoration, runtime verification and bundle construction, but then fails its own exactness oracle. The first full `verify_physical_qualification_package.py` call invokes the bundled Python closure and writes `__pycache__/*.pyc` into `tools/physical/` and `support/cflib-source/`; the subsequent manifest-only verification correctly rejects those newly created unmanifested files. Verification of an exact artifact must be observational. Run effect-free import verification with bytecode generation disabled (`PYTHONDONTWRITEBYTECODE=1` / `-B`) and keep the packaged runner from writing bytecode into manifest-covered source paths as well.
3. `_copy_tree(cflib_root, bundle / "support" / "cflib-source")` still copies checkout administration state including `.git/`. Those bytes are not determined by the verified cflib commit/tree/subtree and can differ between equivalent checkouts, so the claimed deterministic artifact is not yet content-canonical. Package a canonical source representation (or otherwise exclude checkout-local Git metadata) and verify provenance from bound content rather than mutable checkout administration bytes.
4. Provenance records `webots_version: R2025a` and the world is rewritten to `webots://`, but `run_packaged_physical_qualification.sh` neither resolves nor verifies an R2025a Webots executable and does not pass an explicit `--webots` path. Because `webots://` is interpreted by the installed Webots tree, a different release can silently change runtime resources while package checks pass. Fail closed on exact R2025a before opening a session and pass that exact executable to the trusted launcher.

These are machine-preparation findings only. They do not justify a real-device checkpoint, package-network acceptance, or any weakening of teacher/preflight/effect authority boundaries.

djibian:
DRAFT_FINDING 5dae3f6f27d99d2a50456d7372fc9ddac72989e2

The latest Draft run narrows several deterministic repair boundaries; none requires human evidence or a rerun of this unchanged SHA.

1. Canonical `CI Gate` is red because this PR adds `.github/workflows/physical-qualification-package.yml`. The current V4 workflow contract deliberately admits only `human-checkpoint.yml`, `ci.yml`, `ci-runtime.yml`, and `ci-webots.yml`; automated integration evidence belongs in canonical `CI Gate`, not a fifth candidate-evidence workflow. Smallest repair: fold bounded package preparation/build/verify/upload into `.github/workflows/ci.yml` (reusing the integrated #385 qualification-runtime support) and delete the extra workflow. Preserve Draft-vs-Ready exact-head semantics and do not weaken `test_only_v4_workflows_exist`.

2. The dedicated package job now restores both caches and gets through package construction plus the first full verifier, but the regression oracle then fails because effect-free import verification writes `__pycache__/*.pyc` into the bundle. That mutates the supposedly exact artifact after its manifest was written. Smallest repair: make package verification/import probes read-only with respect to packaged bytes (for example `PYTHONDONTWRITEBYTECODE=1` / `-B`, or execute against a disposable copy), and add an oracle that the manifest-valid bundle remains byte/file-set identical after verification.

3. `package_physical_qualification.py` still copies the entire verified cflib checkout into `support/cflib-source`, including `.git/`. The build-time verifier proves exact commit/tree/subtree, but checkout-administration bytes (`index`, config, refs/logs, etc.) are not part of that proven source tree and can differ between equivalent materializations. For a deterministic artifact, package canonical tracked cflib content without `.git`, preserve the exact commit/tree/subtree in provenance, and verify the bundled canonical tree/content independently. Do not make runtime execution depend on mutable Git administration metadata.

4. `PROVENANCE.json` currently records `execution_authority_packaged: false`. That overstates the trust boundary: the bundle contains the trusted physical host/launcher which, only after the live-session/preflight chain and explicit teacher decision, can create/consume execution authority. The accurate invariant is that packaging/preparation mints no authority and emits no effect; browser preparation remains `executionAuthority:false`; execution still requires the existing trusted host + explicit teacher authorization. Adjust the provenance field/oracle to state that boundary rather than claiming the complete runnable artifact can never hold authority.

The prior root-owned cached controller chmod failure is resolved at this HEAD: the restored cached binary passes `test -x`. Preserve that improvement while applying the boundaries above.

djibian:
DRAFT_FINDING 5dae3f6f27d99d2a50456d7372fc9ddac72989e2

Two additional determinism boundaries are still unproven on this exact Draft head:

1. `package_physical_qualification.py::_require_source_sha()` verifies only that `git rev-parse HEAD` equals the requested SHA. The bundle is then assembled from worktree paths, but tracked worktree cleanliness is never checked. A locally modified tracked `tools/physical`, Robot Window, world or controller source can therefore be packaged while `SOURCE_SHA` / provenance still claim the unmodified commit. #393 requires an artifact bound to the exact repository SHA. Preserve generated/ignored prepared inputs, but fail closed on tracked modifications (for example `git status --porcelain --untracked-files=no`) before copying repository-controlled content, and cover that with a regression.

2. The prepared Runtime/controller cache key says `r2025a` but its cache miss path builds with the mutable Docker tag `cyberbotics/webots:R2025a-ubuntu22.04`; the image digest is neither pinned in workflow source nor bound into package provenance. The observed Draft pull happened to resolve to `sha256:f0023e30daf38b172e4e6ad24ed345909bcd9551df34d63d824e121a7cebf099`, but a future cache re-prime under the same key could resolve the tag to different bytes and produce a different controller while all current key inputs remain identical. Deterministic preparation needs the build environment content-addressed (or an equivalently exact controller-output proof) and that provenance must participate in the cache/artifact contract.

Both findings are machine-only. They require no hardware or new human checkpoint.

djibian:
DRAFT_STATUS 752c8d6c757aeae27cd7cc6c6bc1a268295c950d

Exact-head re-evaluation narrows the earlier findings:

- RESOLVED on this head: package verification now disables Python bytecode writes, and the packaged runner also exports `PYTHONDONTWRITEBYTECODE=1`; package run `34848635799` succeeds end-to-end. The runner now resolves an explicit Webots executable, requires an `R2025a` version string and passes that exact path to the trusted launcher. Those two earlier closure findings are no longer applicable.
- STILL APPLICABLE: canonical `CI Gate` run `34848636862` remains red because the fifth `.github/workflows/physical-qualification-package.yml` violates the existing V4 workflow allowlist. The package proof still needs to move into an already-authorized canonical workflow boundary rather than weakening the oracle.
- STILL APPLICABLE by unchanged source: the bundle copies cflib checkout administration state (`.git/`); `execution_authority_packaged: false` still overstates the runnable artifact trust boundary; `_require_source_sha()` still does not reject tracked worktree modifications; and the prepared Runtime/controller cache is still keyed only by the mutable `cyberbotics/webots:R2025a-ubuntu22.04` tag inputs rather than content-addressed Webots build-environment provenance.

No human evidence is needed for these remaining machine-preparation boundaries.

djibian:
DRAFT_FINDING 752c8d6c757aeae27cd7cc6c6bc1a268295c950d

The existing worktree-cleanliness finding needs one fail-closed refinement before implementation: checking only tracked modifications (for example `git status --porcelain --untracked-files=no`) is not sufficient for the current packager.

`package_physical_qualification.py` recursively copies repository-controlled directories such as `tools/physical/` and the shared Blockly source tree. A non-ignored **untracked** file placed under one of those copied roots would therefore be included in the bundle and manifest while `SOURCE_SHA` / `PROVENANCE.json` still name the unchanged Git commit. That is the same provenance failure as a modified tracked file, just through a different worktree state.

Smallest robust boundary: before copying repository-controlled content, require the worktree to contain neither tracked changes nor non-ignored untracked files (while ignored/generated prepared inputs such as the deliberately materialized Runtime assets/controller may remain), or construct the repository-controlled portion directly from the exact Git tree rather than broad `copytree` calls. Add a regression that an untracked file under a copied source root cannot enter an artifact claiming the clean source SHA.

This is machine-only provenance hardening; it changes no execution authority or human checkpoint boundary.

djibian:
DRAFT_FINDING fda1024595f7a652be75b8997093b4e9c7313e49

The new provenance field `webots_build_image_digest = sha256:f0023e30...` does not yet prove that the controller was built by that image. The workflow still does `docker pull cyberbotics/webots:R2025a-ubuntu22.04` and `docker run ... cyberbotics/webots:R2025a-ubuntu22.04`; it never verifies the resolved image digest against the constant now written into `PROVENANCE.json`.

If that mutable tag moves, a cache miss can build a different controller while the package still asserts the old digest. This turns the new provenance field into an unsupported claim rather than a deterministic bound.

Smallest repair: make the build actually content-addressed (`cyberbotics/webots@sha256:f0023e30...`) for both pull/run, or fail closed by verifying the locally resolved RepoDigest/image identity before the build and bind that verified value into the cache key/provenance. Add an oracle that changing/resolving a different image cannot produce an artifact that still claims the pinned digest.

This is machine-only build provenance; no physical checkpoint follows.

djibian:
DRAFT_FINDING a29ff716e8e624b38c5fd3f01d2a2de5ac1d0ea3

Folding the package path into canonical `ci.yml` fixes the fifth-workflow governance defect, but the new CI selector/cache boundary is still too narrow for the artifact it claims to prove.

`package_physical_qualification.py` packages repository-controlled Runtime inputs beyond `tools/physical/**`: notably `plugins/robot_windows/blockly_v2/**`, shared `plugins/robot_windows/blockly/webeeblocks/**`, `crazyflie_v2.js`, `worlds/crazyflie_runtime_v2.wbt`, and controller/PID sources. The removed dedicated workflow used path filters covering these surfaces. The canonical `select_qualification_runtime_support.py`, however, currently selects only the bounded CI block plus `tools/physical/*.py`/a few support files. A future Ready PR changing (for example) `blockly_v2/webots/RobotWindow.js`, the world, shared semantic/runtime JS or controller source can therefore skip the package proof entirely.

There is a second consequence in the same boundary: the prepared-input cache restores the tracked `plugins/robot_windows/blockly_v2/webots` directory, but its key does not include those tracked Webots bridge files. If one changes while package proof is selected for some other reason, an older exact-key cache can overwrite the checked-out candidate bytes before packaging; the bundle may then be internally self-consistent yet no longer represent the named source SHA.

Smallest repair: make the qualification-package selector cover every repository-controlled path that contributes bytes/semantics to the bundle, and make cache keys cover every tracked path they restore (or stop caching tracked repository bytes and cache only genuinely generated inputs). Add regressions that a change to each representative packaged surface selects package proof and that cache restoration cannot replace candidate-controlled bytes while provenance still claims the candidate SHA.

This is a canonical machine-evidence/provenance finding only; it creates no human checkpoint.

djibian:
DRAFT_FINDING a29ff716e8e624b38c5fd3f01d2a2de5ac1d0ea3

Exact-head canonical run `34849571755` now reaches the integrated package build/proof, but the machine oracle still has two deterministic contract defects:

1. The package regression checks for the literal `env["PYTHONDWRITEBYTECODE"] = "1"`, while the verifier correctly sets `env["PYTHONDONTWRITEBYTECODE"] = "1"`. That typo is the immediate `Select suites` failure after bundle construction and full effect-free verification already pass. Repair the oracle to assert the actual `PYTHONDONTWRITEBYTECODE` contract; do not change the working verifier to satisfy the misspelling.
2. `Verify selector and repository contracts` invokes `test_physical_qualification_package.py <bundle>` unconditionally, but the bundle is created only when `qualification-runtime-scope.outputs.run == 'true'`. A future unrelated PR for which that selector is false will therefore fail merely because the artifact directory does not exist. Keep the static package/workflow contract runnable independently (for example a `--contract-only` mode) and reserve mutation/package-byte tests for the bounded package-build step, or condition that dynamic invocation on the same selector.

The existing untracked-worktree provenance finding also remains applicable at this head: `_require_source_sha()` still uses `--untracked-files=no` while repository-controlled roots are copied recursively.

These are machine-only Draft findings; no human checkpoint follows.

djibian (commented):
NO_GO 9f130a7086343678abfc242e340b552f13090ecc

The exact Ready candidate still has two applicable provenance/evidence defects despite the green Draft run; the newest Ready CI is currently a separate pending decision and cannot cure either source-level issue.

1. `select_qualification_runtime_support.py::relevant_path()` still selects only the qualification tooling / `tools/physical/*.py` surface. The artifact also packages repository-controlled Robot Window/runtime assets, shared `plugins/robot_windows/blockly/webeeblocks/**`, `crazyflie_v2.js`, `worlds/crazyflie_runtime_v2.wbt`, and controller/PID sources. A later candidate changing one of those packaged inputs can therefore skip the package build/proof entirely. Smallest repair: extend the selector (and regression oracle) to every repository-controlled input whose bytes/semantics enter the bundle.

2. `_require_source_sha()` still checks `git status --porcelain --untracked-files=no` while the packager recursively copies repository-controlled roots. A non-ignored untracked file under one of those roots can enter the bundle and manifest while `SOURCE_SHA`/provenance still claim the unchanged commit. Smallest repair: construct repository-controlled bytes from the exact Git tree, or fail closed on every non-ignored untracked file that can enter copied roots; add a regression that such a file cannot be packaged under the clean SHA.

The prior fifth-workflow, bytecode-mutation, cflib `.git`, Webots-version, build-image and authority-provenance findings are resolved by this head. Keep those fixes. No human checkpoint is involved.

djibian (commented):
NO_GO 9f130a7086343678abfc242e340b552f13090ecc

The earlier untracked-worktree provenance finding remains applicable on this exact Ready head. `package_physical_qualification.py::_require_source_sha()` still runs `git status --porcelain --untracked-files=no`, while the packager later recursively copies repository-controlled roots such as `tools/physical/` and `plugins/robot_windows/blockly_v2/`. A non-ignored untracked file under one of those roots can therefore enter the manifest/artifact while `SOURCE_SHA` and provenance still claim the unchanged Git commit.

Smallest repair boundary: fail closed on all non-ignored worktree changes (`--untracked-files=all`, preserving ignored generated Runtime/controller inputs), and add an executable regression proving a non-ignored untracked file under a copied source root cannot pass exact-SHA packaging. The current static oracle explicitly requires `--untracked-files=no`, so it must be corrected rather than preserving the gap.

No human checkpoint is implicated.

djibian:
NO_GO 9f130a7086343678abfc242e340b552f13090ecc

The current Ready head still leaves one provenance boundary from the prior Draft finding unresolved.

`tools/ci/select_qualification_runtime_support.py` selects only `tools/physical/**` plus a few support files, while the packaged artifact also embeds repository-controlled `plugins/robot_windows/blockly_v2/**`, shared `plugins/robot_windows/blockly/webeeblocks/**`, `crazyflie_v2.js`, `worlds/crazyflie_runtime_v2.wbt`, and Runtime v2/PID controller sources. A Ready PR changing one of those packaged surfaces can therefore skip the qualification-package proof entirely.

The same boundary restores `plugins/robot_windows/blockly_v2/webots` from a cache whose key does not bind those tracked bridge files. Cache restore can overwrite candidate-controlled tracked bytes before packaging, allowing an internally valid artifact to diverge from the claimed source SHA.

Smallest repair: make the qualification selector cover every repository-controlled packaged/preparation input, and stop caching tracked repository bytes (in particular the tracked `blockly_v2/webots` tree), leaving only genuinely generated vendor/controller outputs in the prepared-input cache. Add machine oracles for representative packaged paths and for the no-tracked-overwrite cache contract. No human checkpoint is implicated.

djibian:
DRAFT_FINDING a0077bde814a18374b8223ddae2bd9cf988a05e2

The new untracked-package fence still has a filename-encoding hole. `_require_source_sha()` asks Git for `status --porcelain=v1 --untracked-files=all`, then `_unexpected_untracked_package_paths()` only strips the literal `?? ` prefix and treats the remainder as a raw repository-relative path. Porcelain v1 C-quotes pathnames containing spaces and other special bytes; for example an untracked `tools/physical/rogue file.py` is emitted as `?? "tools/physical/rogue file.py"`. `_would_copy_untracked()` then does not see the `tools/physical/` prefix, while `_copy_tree()` will copy that file into the artifact. The exact-SHA provenance gap therefore remains for quoted untracked names.

Smallest repair: consume an unambiguous NUL-delimited Git form (`--porcelain=v1 -z` or `git ls-files --others --exclude-standard -z`) and classify decoded repository-relative paths, rather than parsing human/C-quoted status lines. Extend the executable oracle with at least an untracked packaged path containing a space (and preferably another quoted-path case) and require rejection.

This is machine-only provenance hardening; no human checkpoint is involved.

djibian:
DRAFT_FINDING 9b1f4ff58b147a7b6623221e6ef3e08035e24608

The untracked-path repair closes the ordinary-name case, but its parser is still not exact for every path Git can report. `_require_source_sha()` asks `git status --porcelain=v1 --untracked-files=all` without `-z`, then `_unexpected_untracked_package_paths()` line-splits the human/quoted form and assumes every record begins `?? ` followed by the literal repository path. Git C-quotes unusual pathnames in non-`-z` porcelain output; for example an untracked copied file containing spaces, quotes, non-ASCII escaping or a newline can be represented with quoting/escapes, so `_would_copy_untracked()` no longer sees a string starting with `tools/physical/` or another packaged root and the file can still enter the artifact under an unchanged `SOURCE_SHA`.

Smallest robust repair: consume a machine-stable NUL-delimited form (`status --porcelain=v1 -z` or `git ls-files --others --exclude-standard -z` scoped to packaged inputs) and parse NUL records, then extend the regression with at least a packaged untracked pathname containing a space plus an escaped/non-ASCII or newline-sensitive case. Preserve the explicit generated-controller allowance.

This is machine-only exact-source provenance hardening; no human checkpoint follows.

djibian:
DRAFT_FINDING f9114d8000e817b24d33d189fc1b15dc0e2a29cd

The exact-source provenance gap around unusual untracked filenames remains present on this head. `package_physical_qualification.py::_require_source_sha()` still calls `git status --porcelain=v1 --untracked-files=all` without `-z`, and `_unexpected_untracked_package_paths()` still parses line-oriented human/C-quoted porcelain by stripping a literal `?? ` prefix. Git quotes/escapes pathnames containing spaces, quotes, non-ASCII escapes or newlines; a non-ignored untracked file under a copied root can therefore evade `_would_copy_untracked()` while `_copy_tree()` still includes the real file in the artifact under the unchanged `SOURCE_SHA`.

Smallest repair: consume a machine-stable NUL-delimited source of untracked names (`git ls-files --others --exclude-standard -z` or `status --porcelain=v1 -z`) and classify decoded repository-relative records. Extend the executable oracle with at least a packaged untracked pathname containing a space and one escape/newline-sensitive case, while preserving the explicit generated-controller allowance.

The selector/cache-scope changes on this head address the separate prior path-selection/cache-overwrite finding; keep them. This is machine-only exact-source provenance hardening and does not require a human checkpoint.

djibian:
DRAFT_FINDING f9114d8000e817b24d33d189fc1b15dc0e2a29cd

The current untracked-source guard still has a second exact-provenance hole beyond the already-recorded porcelain quoting issue: ignored files are invisible to `git status --porcelain=v1 --untracked-files=all`, but the packager's broad `_copy_tree()` calls still copy many ignored names.

Repository `.gitignore` ignores `build/`, `*.so`, `*.dll`, `*.exe`, `.DS_Store`, `*.swp`, etc. `_copy_tree(REPO_ROOT / "tools" / "physical", ...)` and the shared Blockly copy exclude only `__pycache__`/`*.pyc` (and `node_modules` for `blockly_v2`). Therefore an ignored `tools/physical/rogue.so`, `plugins/robot_windows/blockly/webeeblocks/.probe.swp`, or `.DS_Store` under a copied root can enter `SHA256SUMS.json` while `SOURCE_SHA` still names the unchanged commit. The current allowance needs to be for the *specific deterministic generated inputs* (prepared `blockly_v2/vendor/**` and the built controller), not for arbitrary ignored bytes under repository-controlled copied roots.

Smallest repair: construct repository-controlled portions from the exact Git tree plus an explicit allowlist of deterministic generated inputs, or independently enumerate/reject ignored files that would be copied except those exact generated paths. Add an executable regression proving an ignored file under a copied source root cannot enter an exact-SHA artifact while the intended generated vendor/controller remain allowed.

This is machine-only package provenance and does not require a human checkpoint.

djibian (commented):
NO_GO bdf7a14b1e0eb0ce953e4543920a0cb3a448dbe2

The lossless `-z` parsing fixes the unusual non-ignored pathname finding, but the separate exact-source provenance refutation remains applicable: Git status without `--ignored` cannot see ignored files, while broad `_copy_tree()` still copies ignored bytes under repository-controlled roots. With the current `.gitignore`, for example `tools/physical/rogue.so`, `plugins/robot_windows/blockly/webeeblocks/.probe.swp` or `.DS_Store` can enter the bundle/manifest while `SOURCE_SHA` still names the unchanged commit.

Smallest repair boundary: derive repository-controlled package content from the exact Git tree and overlay only the explicitly intended deterministic generated inputs (`blockly_v2/vendor/**` and the exact built controller), or equivalently enumerate/reject every ignored path that would actually be copied except those explicit generated inputs. Add an executable regression proving an ignored file under a copied source root cannot enter an exact-SHA artifact while the intended generated vendor/controller remain allowed.

The selector/cache scope and NUL-delimited untracked parsing improvements on this head should be preserved. No human checkpoint is implicated.

djibian:
DRAFT_FINDING b96525655970f159a999e036dae3f5d761d16310

The source-level ignored-file provenance defect is now addressed in the packager: ignored records are included via `--ignored`, `!! ` records are classified, and only the intended generated vendor prefix / exact controller are exempted. Preserve that repair.

One part of the exact-head repair boundary is still missing: `test_physical_qualification_package.py` has no regression for ignored contamination. Its current fixtures cover `?? ` untracked records (including unusual NUL-delimited names), but not an ignored `!! tools/physical/rogue.so` / `.swp`-style path and the intended `!! plugins/robot_windows/blockly_v2/vendor/...` exemption. Add that executable oracle and assert the static contract includes the `--ignored` guard before this becomes Ready again.

No human checkpoint is involved.

djibian:
DRAFT_FINDING 7f1d7daeea0a8b54775f9e07e5681fed1bba8aae

The source-tree contamination refutation is now materially repaired, but the generated-input cache still leaves the artifact's claimed build provenance unproven on Ready runs.

The cache key binds the generator/source inputs and the Webots image digest, but the restored cache payload itself (`blockly_v2/vendor/**` plus the built `crazyflie_runtime_v2` ELF) has no expected content digest/lock. Ready acceptance only checks Blockly `VERSION`, a few file-presence predicates and that the controller is executable. A same-branch Draft can therefore populate that exact key with different generated bytes (for example through a temporarily changed CI build command or other preparation defect), and a later Ready head with the same keyed inputs will restore and package those bytes. The package manifest then self-hashes the restored payload and provenance still claims the exact source/build-image inputs, without independently establishing that the cached outputs are the deterministic outputs of those inputs.

This matters because #393 explicitly requires a deterministic artifact bound to exact source/runtime provenance, and `PROVENANCE.json` now makes an exact Webots build-image claim. Cache identity is useful for lookup, but is not an oracle for cached content.

Smallest repair boundary: bind the generated vendor closure and controller ELF to reviewed expected SHA-256 values (or an equivalently independent content lock) derived for the exact keyed inputs, and verify those digests after every cache restore before packaging/Ready acceptance. Include the lock itself in the cache key/selector contract. Do not weaken the current Draft-only cache priming or offline Ready boundary.

This is machine-only deterministic-provenance evidence; no human checkpoint is implicated.

djibian (commented):
NO_GO 7f1d7daeea0a8b54775f9e07e5681fed1bba8aae

The prior selector/cache-overwrite, untracked-path quoting and ignored-source contamination refutations are resolved by this exact head. One independent deterministic-provenance defect remains in the generated-input cache boundary.

Ready runs restore `plugins/robot_windows/blockly_v2/vendor/**` and the built `controllers/crazyflie_runtime_v2/crazyflie_runtime_v2` from an Actions cache whose key binds source/generator inputs and the Webots image digest, but not the expected bytes of those generated outputs. The Ready fail-closed step verifies only Blockly `VERSION`, file presence and controller executability. A same-branch Draft can therefore populate that exact key with different generated bytes (for example through a temporarily changed preparation/build command), after which a later Ready head with the same keyed inputs accepts and packages them. The package manifest merely hashes the restored bytes themselves, so it cannot establish the claimed relationship between those bytes and the exact source/build-image provenance recorded by #393.

Smallest repair boundary: add reviewed expected SHA-256 locks (or an equivalently independent content oracle) for the generated Blockly closure and controller ELF for the exact keyed inputs, verify those values immediately after every cache restore and before packaging, and include the lock in the cache key/selector contract. Preserve Draft-only cache priming and offline Ready behavior.

This is machine-only deterministic provenance; no human checkpoint is implicated.

djibian (commented):
GO 09bb07c36d5ddbb62be9bc904d2dbe21d13c7811

Independent falsification review finds all previously applicable exact-source/provenance refutations resolved on this Ready HEAD.

- Package selection now covers the repository-controlled closure whose bytes/semantics enter the artifact.
- Tracked dirtiness fails closed; NUL-delimited status parsing covers non-ignored and ignored paths that broad copying could otherwise admit, while only the explicitly intended generated vendor closure and controller binary are exempted.
- Generated cache outputs are no longer accepted from provenance/key presence alone: `qualification_generated_lock.json` binds the complete Blockly vendor closure and exact controller ELF by content/size, the verifier rejects mismatch/symlink, the lock participates in the cache key and package selector, and verification runs after restore and before packaging. The preceding clean Draft cache-miss run regenerated those outputs from the pinned preparation/build path and matched the committed lock.
- Earlier fifth-workflow, bytecode mutation, cflib checkout metadata, Webots-version/build-image, authority-provenance and cache-overwrite findings remain resolved.

The trust boundary stays machine-preparation-only: no radio session, checkpoint request, teacher-authority minting or physical effect is introduced. The newest Ready `CI Gate` run is still in progress; this GO does not substitute for its required exact-head successful completion.

djibian (commented):
GO 89176eecfe591b6aadcec22cae4b2182894dff1a

Independent falsification review on the exact current Ready HEAD finds the previously applicable refutations resolved and the base-sync merge introduces no new #393 delta beyond incorporating current `main@7a5ef247f5d4af5339f5cc92c43ff9f8c6de7f8c`.

- The PR diff from current main remains exactly the eight qualification-package files; the #396 X3 files are inherited unchanged from main.
- Selector coverage, tracked/untracked/ignored source contamination rejection, lossless `-z` handling and exclusion of tracked Webots bytes from cache remain intact.
- `qualification_generated_lock.json` now independently binds the complete generated Blockly vendor closure and exact controller ELF; the verifier rejects closure/count/hash/size/symlink mismatch, the lock is part of the exact cache key and selector, and canonical CI verifies generated outputs after generation/restore and before packaging. The prior clean cache-miss generation matched the committed lock exactly.
- Earlier workflow-governance, bytecode-mutation, canonical cflib-tree, Webots R2025a/build-image, authority-provenance and artifact exactness findings remain resolved.

The machine-only boundary remains unchanged: no radio session, physical effect, teacher-authority minting, checkpoint request or `TEST_REQUIRED` is introduced. This GO does not substitute for the required newest exact-head Ready `CI Gate` success.

